PR #10, opened 2020-12-25
Added missing typecast to float.

dlemstra:
Not sure if this is just a Visual Studio issue but `std::signbit` and `std::isinf` have three template methods (`float`, `double` and `long double`. Adding a typecast resolves this issue.


jan-wassenberg:
Thanks for reporting this. Apparently msvc doesn't realize this is dead code when instantiated for integers. Changing to if constexpr would fix it but imposing C++17 also has costs.

I was concerned about casting infinities from float<->double but it is well-defined, and the alternative of SFINAE to elide the code for integer FromT is rather verbose, so let's go with your approach :)

Will make a minor change to cast to double instead of float (we use the double value later anyway, note that IsFloat<double>()==true).

PR #12, opened 2020-12-28
Use different implementation for PopCount on 32 bit Windows.

dlemstra:
I am trying to compile the jpeg-xl library on 32 bit Windows in Visual Studio and I ran into a compilation issue where `_mm_popcnt_u64` was undefined on 32 bit Windows. This PR adds a custom implementation of PopCount and changes the check to add support for 32 bit compilation on Windows. The code that was used comes from this implementation: https://github.com/shines77/strlen_fast/blob/master/src/strlen_fast/popcnt.h. An other option would be to use `std::popcount` but that requires C++20 and I don't think that should be used already.

### DIFF
--- a/hwy/ops/shared-inl.h
+++ b/hwy/ops/shared-inl.h
@@ -200,13 +200,20 @@ HWY_API void CopyBytes(const From* from, To* to) {
 #endif
 }
 
-HWY_API size_t PopCount(const uint64_t x) {
+HWY_API size_t PopCount(uint64_t x) {
 #if HWY_COMPILER_CLANG || HWY_COMPILER_GCC
   return static_cast<size_t>(__builtin_popcountll(x));
-#elif HWY_COMPILER_MSVC
+#elif HWY_COMPILER_MSVC && _WIN64
   return _mm_popcnt_u64(x);
 #else
-#error "not supported"
+  x -=  ((x >> 1) & 0x55555555U);
+  x  = (((x >> 2) & 0x33333333U) + (x & 0x33333333U));
+  x  = (((x >> 4) + x) & 0x0F0F0F0FU);
+  x +=   (x >> 8);
+  x +=   (x >> 16);
+  x +=   (x >> 32);
+  x  = x & 0x0000007FU;
+  return (unsigned int)x;
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


dlemstra:
On a side note:

There is also a `PopCount` implementation in `bits.h` of `jpeg-xl`. There is an implementation available for `uint32_t` and `uint64_t`  but only the `uint64_t` version is used. Maybe you could also replace that implementation with the one from this library. 

The same issue occurs there for `_BitScanReverse64` and `_BitScanForward64` that is not available on 32 bit Windows. The `strlen_fast` library  also includes an implementation for `ctzll` and `clzll`. And that could be used there instead. If you prefer me to open an issue in the jpeg-xl library I can do that.

Bulat-Ziganshin:
more efficient x86 implementation would be `_mm_popcnt_u32(x) + _mm_popcnt_u32(x>>32)` so you can replace
```
#elif HWY_COMPILER_MSVC && _WIN64
  return _mm_popcnt_u64(x);
#else
```

with 
```
#elif HWY_COMPILER_MSVC && _WIN64
  return _mm_popcnt_u64(x);
#elif HWY_COMPILER_MSVC // && x86_32
  return _mm_popcnt_u32(uint32_t(x)) + _mm_popcnt_u32(uint32_t(x>>32));
#else
```

dlemstra:
Thanks @Bulat-Ziganshin.

hwy/ops/shared-inl.h
```diff
@@ -200,13 +200,22 @@ HWY_API void CopyBytes(const From* from, To* to) {
 #endif
 }
 
-HWY_API size_t PopCount(const uint64_t x) {
+HWY_API size_t PopCount(uint64_t x) {
 #if HWY_COMPILER_CLANG || HWY_COMPILER_GCC
   return static_cast<size_t>(__builtin_popcountll(x));
-#elif HWY_COMPILER_MSVC
+#elif HWY_COMPILER_MSVC && _WIN64
```
jan-wassenberg, line 206: Let's use HWY_COMPILER_MSVC && HWY_ARCH_X86_64?

jan-wassenberg (changes requested):
Nice, one small suggestion, after that LGTM.

jan-wassenberg:
Thanks @dlemstra and @Bulat-Ziganshin, nice improvement!

I've made a note to use hwy popcount in jxl after we uprev to the latest Highway version.

This will be the first time we test the github->internal sync, hopefully we'll get it to work :)

hwy/ops/shared-inl.h
dlemstra: Did not know that existed. Made the change as you suggested.

jan-wassenberg (commented):
LGTM

PR #122, opened 2021-04-08
Include immintrin.h headers when using clang-cl.

deymo:
clang-cl defines _MSC_VER which when including <immintrin.h>
causes all the headers for instruction sets not already defined in
the compile line to be excluded. This means that when compiling for
dynamic dispatch targets the required headers are not actually included
when using clang-cl. This patch worksaround the issue by directly
including those targets in that case.

### DIFF
--- a/hwy/ops/x86_256-inl.h
+++ b/hwy/ops/x86_256-inl.h
@@ -20,6 +20,17 @@
 // particular, "Broadcast", pack and zip behavior may be surprising.
 
 #include <immintrin.h>  // AVX2+
+#if defined(_MSC_VER) && defined(__clang__)
+// Including <immintrin.h> should be enough, but Clang's headers helpfully skip
+// including these headers when _MSC_VER is defined, like when using clang-cl.
+// Inclde these directly here.
+#include <smmintrin.h>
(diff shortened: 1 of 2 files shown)


hwy/ops/x86_256-inl.h
```diff
@@ -20,6 +20,19 @@
...
+#if defined(_MSC_VER) && defined(__clang__)
+// Including <immintrin.h> should be enough, but Clang's headers helpfully skip
+// including these headers when _MSC_VER is defined, like when using clang-cl.
+// Inclde these directly here.
+#include <smmintrin.h>
+#include <avxintrin.h>
+#include <avx2intrin.h>
+#include <f16cintrin.h>
+#include <fmaintrin.h>
+#include <avx512fintrin.h>
```
veluca93, line 32: This seems wrong to me :)

veluca93:
LGTM

hwy/ops/x86_256-inl.h
deymo: It was not compiling without these, but I think it was because I had the order wrong. I just checked that I can remove these two.

PR #148, opened 2021-04-17
scalar: static cast to fix warning in std::abs

boxerab:


jan-wassenberg:
Thanks! Would you mind changing the float to T in case we add support for double later?

boxerab:
> Thanks! Would you mind changing the float to T in case we add support for double later?

Done!

jan-wassenberg (approved):
Thank you!

PR #233, opened 2021-06-18
Consider non-V extensions RISC-V

makotokato:
__fp16 is defined in riscv_vector.h and this is in vector extension.
So if not using RVV, __fp16 is undefined. So highway shouldn't set
HWY_NATIVE_FLOAT16 without vector extension.

And, actually, since VLEN isn't exported to pre-defined macro for compiler/OS,
kMaxVectorSize is too large size. But it is unnecessary without vector
extension. So if compiler doesn't support vector extension, use default (16).


jan-wassenberg (approved):
Thanks! I will try to merge this via our copybara process, not sure how well it will work.

jan-wassenberg:
(There's a permissions issue with CI/CD, working on it.)

jan-wassenberg:
Sync worked :)

PR #413, opened 2021-10-20
Build and test on Windows MSYS2

deymo:
New GitHub workflow to build and test on a Windows MSYS2 build in 32 and
64 bits.

This patch sets the `LANGUAGES` to `CXX` in the cmake project to avoid
setting up the `C` language which is not used by the project.

Due to a bug in cmake ninja generator with MSYS we can't set a prefix on
the target because it creates a duplicated ninja rule.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -55,3 +55,46 @@ jobs:
           path: ~/.cache/bazel
           key: bazel-${{ runner.os }}
       - run: bazel build //...
+
+  windows_msys2:
+    name: Windows MSYS2 / ${{ matrix.arch }}
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: x86_64
+            msystem: mingw64
+          - arch: i686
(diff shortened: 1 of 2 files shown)


jan-wassenberg:
I'm concerned about the 11m runtime of the x64 builder, that might cause timeouts. Anything we can do to speed it up?
For example, the other builders do export CMAKE_BUILD_PARALLEL_LEVEL=2 or "uses: actions/cache@v2".

jan-wassenberg:
> scalar (1m) vs these builders that compile all (5m). We could compile less (like scalar+avx2). 

That sounds useful, it should hopefully bring us under 10min?

> I think CMAKE_BUILD_PARALLEL_LEVEL=2 is not needed with ninja since it builds in parallel by default.

Ah, makes sense.

jan-wassenberg:
@deymo FYI math_test is now disabled in cmake.

deymo:
@jan-wassenberg I rebased it and SortTestGroup/SortTest.TestAllReverse/AVX2 is now failing, and also clang-5

PR #450, opened 2021-11-12
Only run hwy_list_targets if its possible to do so

donny-dont:
When `CMAKE_CROSSCOMPILING` there may not be any value for `CMAKE_CROSSCOMPILING_EMULATOR`. Only run the command if its not a cross compile or when an emulator is available.


donny-dont:
This change worked for me on Windows when running a MSVC 2022 build and on the same system doing a cross compile on a platform without an emulator.

jan-wassenberg:
Hi @donny-dont , thank you for adding this! I'll start the import process now, looks like I have some connectivity issues at the moment, though.

PR #453, opened 2021-11-12
Provide individual toggles for build options

donny-dont:
Split `HWY_EXAMPLES_TESTS_INSTALL` into `HWY_ENABLE_EXAMPLES`, `BUILD_TESTING` and `HWY_ENABLE_INSTALL`. The `HWY_ENABLE_*` values are set to `ON` by default to mirror previous behavior. `BUILD_TESTING` which is created when CTest is included, is set to `ON` by default so use that rather than creating a `HWY_ENABLE_TESTING` value.


jan-wassenberg:
Nice, thanks!

PR #487, opened 2022-01-04
Implement compilation of highway as shared libraries

### LINKED ISSUE #380: Question: why static libraries only?
Hello,
Some of my friends are asking why libhwy.a and libhwy_contrib.a are build as static libraries only:

`add_library(hwy STATIC ${HWY_SOURCES})`

`add_library(hwy_contrib STATIC ${HWY_CONTRIB_SOURCES})`

Some distributions (for example Gentoo) [prefer to avoid static libraries](https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies), shared/dynamic are preferred.

What are the reasons why only static libs are built now?

Would it it good to enable optional building of dynamic libraries according the `BUILD_SHARED_LIBS` setting?

malaterre:
This should fix #380

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -288,7 +288,6 @@ set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
 # Programming exercise with integrated benchmark
 add_executable(hwy_benchmark hwy/examples/benchmark.cc)
 target_sources(hwy_benchmark PRIVATE
-    hwy/nanobenchmark.cc
     hwy/nanobenchmark.h)
 # Try adding either -DHWY_COMPILE_ONLY_SCALAR or -DHWY_COMPILE_ONLY_STATIC to
 # observe the difference in targets printed.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,7 @@ set(HWY_SOURCES
     hwy/detect_targets.h  # private
     hwy/foreach_target.h
     hwy/highway.h
+    hwy/highway_export.h
     hwy/nanobenchmark.cc
     hwy/nanobenchmark.h
     hwy/ops/arm_neon-inl.h
@@ -200,8 +201,19 @@ endif()  # !MSVC
 option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
 # only expose shared option to advanced users:
 mark_as_advanced(BUILD_SHARED_LIBS)
+# Define visibility settings globally:
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
+
+# This preprocessor define will drive the build, also used in the *.pc files:
+if(BUILD_SHARED_LIBS)
+  set(DLLEXPORT_TO_DEFINE "HWY_SHARED_DEFINE")
+else()
+  set(DLLEXPORT_TO_DEFINE "HWY_STATIC_DEFINE")
+endif()
 
 add_library(hwy ${HWY_SOURCES})
+target_compile_definitions(hwy PUBLIC "${DLLEXPORT_TO_DEFINE}")
 target_compile_options(hwy PRIVATE ${HWY_FLAGS})
 set_property(TARGET hwy PROPERTY POSITION_INDEPENDENT_CODE ON)
 set_target_properties(hwy PROPERTIES VERSION ${LIBRARY_VERSION} SOVERSION ${LIBRARY_SOVERSION})
@@ -209,6 +221,7 @@ target_include_directories(hwy PUBLIC ${CMAKE_CURRENT_LIST_DIR})
 target_compile_features(hwy PUBLIC cxx_std_11)
 
 add_library(hwy_contrib ${HWY_CONTRIB_SOURCES})
+target_link_libraries(hwy_contrib hwy)
 target_compile_options(hwy_contrib PRIVATE ${HWY_FLAGS})
 set_property(TARGET hwy_contrib PROPERTY POSITION_INDEPENDENT_CODE ON)
 set_target_properties(hwy_contrib PROPERTIES VERSION ${LIBRARY_VERSION} SOVERSION ${LIBRARY_SOVERSION})
@@ -216,6 +229,7 @@ target_include_directories(hwy_contrib PUBLIC ${CMAKE_CURRENT_LIST_DIR})
 target_compile_features(hwy_contrib PUBLIC cxx_std_11)
 
 add_library(hwy_test ${HWY_TEST_SOURCES})
+target_link_libraries(hwy_test hwy)
 target_compile_options(hwy_test PRIVATE ${HWY_FLAGS})
 set_property(TARGET hwy_test PROPERTY POSITION_INDEPENDENT_CODE ON)
 set_target_properties(hwy_test PROPERTIES VERSION ${LIBRARY_VERSION} SOVERSION ${LIBRARY_SOVERSION})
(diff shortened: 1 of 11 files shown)


malaterre:
@jan-wassenberg There is no point in merging this PR until following issue is adressed:

* https://github.com/google/highway/issues/488

malaterre:
@jan-wassenberg Two things are left to do:

1. Copy/paste https://github.com/libjxl/libjxl/blob/main/lib/jxl/jxl.version
2. Run clang-format (or other?). I made no attempt at correct indentation in my patch.

jan-wassenberg (approved):

jan-wassenberg:
@malaterre the internal CI noticed that highway_export.h is not present. To also build with Bazel, which is required for us, can we check in the result that CMake would generate? It would also need to be added to the BUILD file under the line "hwy/aligned_allocator.h",    - would you mind doing so?

malaterre:
@jan-wassenberg I think I did not understood your instructions. "bazel" is still failing with:

`

> ERROR: /home/runner/work/highway/highway/BUILD:101:11: Middleman _middlemen/_S_S_Chwy-BazelCppSemantics_build_arch_k8-fastbuild failed: missing input file '//:hwy/highway_export.h'

`

jan-wassenberg:
@malaterre Thank you for adding the line to BUILD, that looks good.

The reason Bazel complains is that we do not have the file highway_export.h in git. This would be required because Bazel does not generate the header like CMake does -- instead it expects the file to be checked in as a normal header. If it's checked in, would that still be ok for CMake, or does it insist on writing the file itself?

We might have to have two headers, one generated by CMake, the other one a normal header that includes CMake's iff building with CMake.

Alternatively, we could consider writing our own _export.h. It seems to require only checking HWY_COMPILER/MSVC/GCC, and `<target_name>_EXPORTS`, to choose between declspec(dllimport/export) and visibility. Here's one such header, though with a lot more macro machinery than seems necessary: https://github.com/Eyenseo/ABI/blob/master/include/ABI.h

> could you comment in the choice HIGHWAY_EXPORT (HWY_EXPORT was already taken).

It is a little surprising because all other macros are HWY_*, but seems reasonable. Do you think HWY_DLLEXPORT or HWY_SHARED would be better?

> Run clang-format (or other?).

Yes please, that would be helpful with --style=Google.

jan-wassenberg:
@malaterre Nice, thanks for adding the header and renaming :) I'm feeding this into our internal tests.

malaterre:
> Let's take a step back here and clarify what problem we're trying to solve.

Provide highway are shared lib for Debian.

---

I fail to understand how bazel works exaclty, but it shouldnt be too hard to fix the header file I uploaded into something that default to static build. What is the error message? I could not figure out which link to click to in the above.

jan-wassenberg:
> Provide highway are shared lib for Debian.

OK, it seems like that goal can be met without having to support Windows.

> What is the error message? I could not figure out which link to click to in the above.

Yes, our internal tests are not externally accessible, but here's one of the blockers:

`third_party/highway/hwy/targets.cc(188,5): error: 'hwy::Abort' redeclared without 'dllimport' attribute: 'dllexport' attribute added [-Werror,-Winconsistent-dllimport]`
This is from an MSVC or clang-cl build for Windows. Apparently it wants HWY_DLLEXPORT also where the function is defined. If we do that, we still have the problem that Bazel doesn't provide macros for switching HWY_DLLEXPORT. Thus I believe the best way forward is to make highway_export.h a no-op (HWY_DLLEXPORT defines to nothing) if *not* building with CMake.

After that, I believe we have a chance of passing the tests :) Sorry about all the back and forth, if you prefer I can also continue with your patch until it works internally, and then send a pull request which you could then copy and re-submit for correct attribution?

malaterre:
@jan-wassenberg Thanks for the feedback, I've been a bit sloppy at supporting windows. I should have double-check at least with a simple compilation. I can reproduce the issue here. Let me fix my PR ASAP.

malaterre:
@jan-wassenberg Could you please comment on 3b64be76b1f20b29592891b68a698d71fca8dae4 I am not 100% sure of what I did. It seems one particular source code was included both in the library and the executable itself.

hwy/highway_export.h
```diff
@@ -0,0 +1,70 @@
...
+/* We are using this library */
+#ifdef _WIN32
+#define HWY_DLLEXPORT __declspec(dllimport)
+#else
+#define HWY_DLLEXPORT __attribute__((visibility("default")))
+#endif
+#endif
+#endif
+
+#ifndef HWY_NO_EXPORT
```
jan-wassenberg, line 39: I think we can remove the NO_EXPORT, DEPRECATED_EXPORT, DEPRECATED_NO_EXPORT, and NO_DEPRECATED blocks? (DEPRECATED is potentially useful, might move it to base.h but it's fine for now)

CMakeLists.txt
```diff
@@ -192,21 +197,48 @@ else()
...
+option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
+# only expose shared option to advanced users:
+mark_as_advanced(BUILD_SHARED_LIBS)
+# Define visibility settings globally:
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
+
+# This preprocessor define will drive the build, also used in the *.pc files:
+if(BUILD_SHARED_LIBS)
+  set(HWY_DLLEXPORT_STRING "HWY_SHARED_DEFINE")
```
jan-wassenberg, line 210: We might consider renaming? Generic terms such as string or manager don't add much value. Maybe something like HWY_WHICH_DLLEXPORT or DLLEXPORT_TO_DEFINE?

jan-wassenberg:
@malaterre 
> Could you please comment on 3b64be7 I am not 100% sure of what I did. It seems one particular source code was included both in the library and the executable itself.

Ah, indeed. Thanks for fixing that. It's a separate library in bazel but I think it's reasonable to keep in the main highway library to avoid more boilerplate in the CMake build.

All good, I left two minor suggestions if you'd like to rename _STRING and shorten the highway_export.h? In parallel, I'll run the internal tests again.

jan-wassenberg:
Good news, the internal checks pass. Would you like to make those changes, or shall we submit as-is? (I can also help do that after this goes in.)

CMakeLists.txt
malaterre: Sure, why not. I picked `DLLEXPORT_TO_DEFINE`.

jan-wassenberg (approved):
Nice, thanks for making the changes! Beginning the merge process.

hmaarrfk:
Is it true to assume that MAC also has support for `--version-script` I am getting linking errors when buliding for OSX.

jan-wassenberg:
@hmaarrfk Sorry to hear. Indeed looks like this is unsupported on OSX and Clang: https://stackoverflow.com/questions/13793325/hiding-symbols-in-a-shared-library-on-mac-os-x

I think we can also skip this on OSX?

malaterre:
Do you want me to submit a PR ?

malaterre:
Here is how jxl is doing it:

https://github.com/libjxl/libjxl/blob/main/lib/jxl.cmake#L581-L591

jan-wassenberg:
@malaterre oh, thanks for offering :) I'd just have changed line 224 to
```
# not supported by MSVC/Clang, safe to skip
if(UNIX AND NOT APPLE)
```

For such a small patch it's quicker if I make the change.
But is it actually safe to omit the hwy.version? We rely on the DLLEXPORT macro and hiding std::* is just an optimization, right?

It would be nice not to have to maintain a separate visibility file for Apple like JXL does.

malaterre:
@jan-wassenberg you're right let's use your suggested patch. And we'll see if highway packagers (brew, ports...) ever complains that there are extra stl::* symbols that needs to be hidden on this arch.

PR #491, opened 2022-01-05
Replace usage of __MACH__ with __APPLE__

malaterre:
It turns out that __MACH__ macro refer to the Mach kernel, which is also
used for GNU/Hurd. Prefer usage of macro __APPLE__ in this case.

This should solve the following compilation error:

nanobenchmark.cc:42:10: fatal error: mach/mach_time.h: No such file or directory

For reference:

* https://developer.apple.com/library/archive/documentation/Porting/Conceptual/PortingUnix/compiling/compiling.html#//apple_ref/doc/uid/TP40002850-SW13

### DIFF
--- a/hwy/nanobenchmark.cc
+++ b/hwy/nanobenchmark.cc
@@ -37,7 +37,7 @@
 #include <windows.h>
 #endif
 
-#if defined(__MACH__)
+#if defined(__APPLE__)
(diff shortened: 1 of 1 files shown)


jan-wassenberg:
@malaterre makes sense, thanks, beginning import :)

PR #499, opened 2022-01-10
Add missing include of stdlib.h for exit() prototype.

malaterre:
This should fix the following compilation error seen on riscv64:

/<<PKGBUILDDIR>>/hwy/targets.cc:205:3: error: ‘exit’ was not declared in this scope
  205 |   exit(1);  // trap/abort just freeze Spike
      |   ^~~~
/<<PKGBUILDDIR>>/hwy/targets.cc:209:1: error: ‘noreturn’ function does return [-Werror]

### DIFF
--- a/hwy/targets.cc
+++ b/hwy/targets.cc
@@ -18,6 +18,7 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <stdlib.h> // exit()
 
 #include <atomic>
 


jan-wassenberg (approved):

jan-wassenberg:
@malaterre Thanks!

jan-wassenberg:
I think this is subsumed by https://github.com/google/highway/pull/502 once that goes in?

malaterre:
> I think this is subsumed by #502 once that goes in?

Oh, right ! Thanks for the notice. I'll remove the local Debian patch, once next release goes out. FYI, riscv64 is currently building fine:

* https://buildd.debian.org/status/package.php?p=highway

jan-wassenberg:
Nice :D

PR #524, opened 2022-01-30
Fix inlcude headers for c++ to avoid compilation errors with PRIu64

hmaarrfk:
Without this, you get an error about undefined `PRIu64`

You probably only need the `<inttypes.h>` -> `<cinttypes>` but I think this is going to help in the long run.

See: https://stackoverflow.com/questions/8132399/how-to-printf-uint64-t-fails-with-spurious-trailing-in-format/8132440#comment9979590_8132440
See: https://stackoverflow.com/questions/12497894/d-stdc-format-macros-gnu-compiler-option

### DIFF
--- a/hwy/contrib/image/image.h
+++ b/hwy/contrib/image/image.h
@@ -17,10 +17,10 @@
 
 // SIMD/multicore-friendly planar image representation with row accessors.
 
-#include <inttypes.h>
-#include <stddef.h>
-#include <stdint.h>
-#include <string.h>
+#include <cinttypes>
+#include <cstddef>
+#include <cstdint>
+#include <cstring>
 
 #include <cstddef>
 #include <utility>  // std::move
--- a/hwy/contrib/sort/sort-inl.h
+++ b/hwy/contrib/sort/sort-inl.h
@@ -22,7 +22,7 @@
 #define HIGHWAY_HWY_CONTRIB_SORT_SORT_INL_H_
 #endif
 
-#include <inttypes.h>
+#include <cinttypes>
 
 #include "hwy/aligned_allocator.h"
 #include "hwy/highway.h"
--- a/hwy/contrib/sort/sort_test.cc
+++ b/hwy/contrib/sort/sort_test.cc
@@ -12,10 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <inttypes.h>
-#include <stdint.h>
-#include <stdio.h>
-#include <stdlib.h>
+#include <cinttypes>
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
 
 #include "hwy/tests/include_farm_sve.h"
 // ^ must come before highway.h.
--- a/hwy/examples/benchmark.cc
+++ b/hwy/examples/benchmark.cc
@@ -16,10 +16,10 @@
 #define HWY_TARGET_INCLUDE "hwy/examples/benchmark.cc"
 #include "hwy/foreach_target.h"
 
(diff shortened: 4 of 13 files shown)


jan-wassenberg:
@hmaarrfk Thanks for reaching out and sending PRs :D 👍 Can you share more about which compilers are affected? From your links, looks like older GCC?
One complication is that using only the C++ headers technically means we also have to change size_t to std::size_t etc, which would be a drag.

[Reference](https://en.cppreference.com/w/cpp/header):
> These headers are allowed to also declare the same names in the std namespace, and the corresponding cxxx headers are allowed to also declare the same names in the global namespace: including <cstdlib> definitely provides std::malloc and may also provide ::malloc. Including <stdlib.h> definitely provides ::malloc and may also provide std::malloc. This applies even to functions and function overloads that are not part of C standard library. Notes: xxx.h headers are deprecated in C++98 and undeprecated in C++23. These headers are discouraged for pure C++ code, but not subject to future removal.

We have at least three possibilities:
1) we use only cinttype alongside the other .h headers (a bit inconsistent)
2) we do not support compilers with this problem, assuming they are too old anyway. (We do try to support as many compilers as possible and the fix seems less painful than excluding a compiler for that reason alone.)
3) we define __STDC_FORMAT_MACROS if that is sufficient for the compiler with this issue.

Do you see other options?

hmaarrfk:
Hey Thanks for the quick reply.

I guess I haven't kept uo with the changes in C++23! They definitely reversed my 25 year old wisdom!

> size_t to std::size_t etc,

We could use `using namespace std` which would be more correct :/.

> what version of GCC

I believe I'm using GCC 9.4.0 

As for the best of 3 solutions, if you are against the c++ only headers in a consistent way, then I would just add __STDC_FORMAT_MACROS to some human building instructions explaining what you explained to me.

I think I was mostly being a purist and didn't want to change CFLAGS when I thought the inclusion of the c headers was "wrong".

hmaarrfk:
I think my issue was setting `CFLAGS` instead of `CXXFLAGS`.

Using
```
CXXFLAGS="-D__STDC_FORMAT_MACROS ${CXXFLAGS}"
```

before the `cmake` commands seems to be enough.

hmaarrfk:
Ok, this is definitely a conda-forge problem working by default with glibc 2.12 which doesn't have the patch above.

The header from the next version that conda-forge will use (not yet fully migrated), glibc 2.17 doesn't have this problem. 

Seeing as glibc 2.12 is from Dec 2010, I don't think you need to do anything.

jan-wassenberg:
Yeah I was also surprised to see that un-deprecated. And thanks for pointing me to the issue, wasn't aware this was a problem. It's regrettable we cannot just use %zu.

Our style guide disallows using namespace especially in headers. I'm definitely not against using cinttypes, that looks like the simplest fix. But if this only affects super old glibc as you say, then I agree we can leave things as-is.

hmaarrfk:
I think using `using namespace std` in headers is also very bad. 

I've typically gotten used to using `std::...` in headers.

hmaarrfk:
Oh, this seems to be the patch:
https://sourceware.org/git/?p=glibc.git;a=commit;h=1ef74943ce2f114c78b215af57c2ccc72ccdb0b7

hmaarrfk:
>  It's regrettable we cannot just use %zu.

Is there a portability concern with %zu?

I don't really use C/C++ too much so I'm rusty on the details.

jan-wassenberg:
Unfortunately %zu doesn't work with some earlier MSVC.

Issue #525, opened 2022-01-30
unresolved external symbol "struct hwy::ChosenTarget hwy::chosen_target"


hmaarrfk:
Thank you for the good work.

I'm trying to build highway for conda-forge where shared libraries are preferred.

I'm running into the following error:
```
[8/10] Linking CXX shared library hwy_contrib.dll
FAILED: hwy_contrib.dll hwy_contrib.lib 
cmd.exe /C "cd . && %BUILD_PREFIX%\Library\bin\cmake.exe -E vs_link_dll --intdir=CMakeFiles\hwy_contrib.dir --rc=C:\PROGRA~2\WI3CF2~1\10\bin\100220~1.0\x64\rc.exe --mt=C:\PROGRA~2\WI3CF2~1\10\bin\100220~1.0\x64\mt.exe --manifests  -- C:\PROGRA~2\MICROS~1\2019\ENTERP~1\VC\Tools\MSVC\1416~1.270\bin\HostX64\x64\link.exe /nologo CMakeFiles\hwy_contrib.dir\hwy\contrib\image\image.cc.obj  /out:hwy_contrib.dll /implib:hwy_contrib.lib /pdb:hwy_contrib.pdb /dll /version:0.15 /machine:x64 /INCREMENTAL:NO  hwy.lib  kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib  && cd ."
LINK: command "C:\PROGRA~2\MICROS~1\2019\ENTERP~1\VC\Tools\MSVC\1416~1.270\bin\HostX64\x64\link.exe /nologo CMakeFiles\hwy_contrib.dir\hwy\contrib\image\image.cc.obj /out:hwy_contrib.dll /implib:hwy_contrib.lib /pdb:hwy_contrib.pdb /dll /version:0.15 /machine:x64 /INCREMENTAL:NO hwy.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib /MANIFEST /MANIFESTFILE:hwy_contrib.dll.manifest" failed (exit code 1120) with the following output:
   Creating library hwy_contrib.lib and object hwy_contrib.exp
[...]

PR #526, opened 2022-01-30
Fix exporting symbols between contrib and main dll

hmaarrfk:
I believe this fixes https://github.com/google/highway/issues/525

By exporting the different symbols based on the different libraries that are built.

I'm not sure if you a test for a dynamically linked build.


jan-wassenberg (approved):
Thanks! Looks good to me, beginning the import process.

PR #527, opened 2022-01-31
Remove erroneous install destination for share libraries

hmaarrfk:
With this patch, CMake seems to choose the correct installation directories for shared libraries

* .dll in bin
* .lib in lib

I think it should be correct for the static builds.

Otherwise, both dll and lib were going in lib.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -266,8 +266,7 @@ endif()
 # -------------------------------------------------------- install library
 if (HWY_ENABLE_INSTALL)
 
-install(TARGETS hwy
-  DESTINATION "${CMAKE_INSTALL_LIBDIR}")
+install(TARGETS hwy)
 # Install all the headers keeping the relative path to the current directory
 # when installing them.
 foreach (source ${HWY_SOURCES})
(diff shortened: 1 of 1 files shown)


jan-wassenberg:
@hmaarrfk hm, I think this might cause trouble on Linux builds. Would it work for you if we instead add the RUNTIME and/or LIBRARY keywords before DESTINATION? Or perhaps two lines with both?
Related example: https://github.com/facebook/zstd/issues/2160

hmaarrfk:
How will we test the new proposal? Do you have CIs where we can add checks to? I've seen alot of changes to the build process recently (as users have been adopting the library)

hmaarrfk:
https://github.com/google/highway/pull/536 is an alternative. I'll test it on our conda-forge build system. But.... all i'm doing is reinforcing the defaults that GNUInstallDirs does 

image
https://cmake.org/cmake/help/latest/command/install.html
https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

hmaarrfk:
So #536 is a more explicit fix if you prefer. Feel free to advance one or the other or none. Thanks for considering!

jan-wassenberg:
> Do you have CIs where we can add checks to?

For testing/CI, we have Github actions. Is there anything in particular you or others would like to see added?

@hmaarrfk thanks, #536 looks good, let's go with that.

hmaarrfk:
Adding actions for shared builds would be great! As well as windows and OSX.

However I do understand that everybody's time is limited.

PR #536, opened 2022-02-01
WIP: Specify the location of the bindir for windows dll

hmaarrfk:


hmaarrfk:
Being tested in https://github.com/conda-forge/libhwy-feedstock/pull/1

jan-wassenberg:
Thanks, this looks good to me. I'm not very knowledgeable about CMake but it's good to know @malaterre and @kmilos also thumbed-up.

Let's re-run the checks and then I can start the import process.

hmaarrfk:
Thank you all.

I'll change  the patches I made downstream to reflect the desires here.

PR #552, opened 2022-02-14
Add an option to override "BUILD_SHARED_LIBS".

eustas:
libjxl sometimes mixes static and shared targets... So it has BUILD_SHARED_LIBS=ON, but wants to use libhwy.a...

Tested on libjxl: https://github.com/libjxl/libjxl/actions/runs/1841037660

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,20 +223,28 @@ endif()  # !MSVC
 
 # By default prefer STATIC build (legacy behavior)
 option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
-# only expose shared option to advanced users:
+option(HWY_FORCE_STATIC_LIBS "Ignore BUILD_SHARED_LIBS" OFF)
+# only expose shared/static options to advanced users:
 mark_as_advanced(BUILD_SHARED_LIBS)
+mark_as_advanced(HWY_FORCE_STATIC_LIBS)
 # Define visibility settings globally:
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
 
+# Copy-cat "add_library" logic + add override.
+set(HWY_LIBRARY_TYPE "SHARED")
+if (NOT BUILD_SHARED_LIBS OR HWY_FORCE_STATIC_LIBS)
+  set(HWY_LIBRARY_TYPE "STATIC")
+endif()
+
 # This preprocessor define will drive the build, also used in the *.pc files:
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -221,14 +221,19 @@ else()
 
 endif()  # !MSVC
 
-# By default prefer STATIC build (legacy behavior)
-option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
-# only expose shared option to advanced users:
+option(HWY_FORCE_STATIC_LIBS "Ignore BUILD_SHARED_LIBS" OFF)
+option(BUILD_SHARED_LIBS "Build shared libraries" ON)
```
jan-wassenberg, line 225: Please change the default to OFF so that there's no change in behavior for users.

CMakeLists.txt
```diff
@@ -221,14 +221,19 @@ else()
...
+# only expose shared/static options to advanced users:
+mark_as_advanced(HWY_FORCE_STATIC_LIBS)
 mark_as_advanced(BUILD_SHARED_LIBS)
 # Define visibility settings globally:
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
 
+if (HWY_FORCE_STATIC_LIBS)
+  set(BUILD_SHARED_LIBS OFF)
```
jan-wassenberg, line 234: How does this interact with the parent project when we're including this as a subdirectory?
Is it scoped to the current directory?

eustas: The change, IIUC, is local - we do not pass `CACHE BOOL 'Description' FORCE` mantra to `set`

eustas: Going to investigate if there is a way to emulate BUILD_SHARED_LIBS effects without modifying it...

jan-wassenberg: I see, it seems reasonable to override this locally, but let's at least change the default value to OFF in line 224, as it was originally?

CMakeLists.txt
eustas: Done and reworked how this is applied. Now it does not change `BUILD_SHARED_LIBS`. Instead it instructs `add_library`.
Tested without options, with `-DBUILD_SHARED_LIBS=ON` and with `-DHWY_FORCE_STATIC_LIBS=ON`.

PTAL

CMakeLists.txt
eustas: Done

jan-wassenberg:
Very nice, thanks for making the change!

jan-wassenberg (approved):

PR #613, opened 2022-03-28
fix build errors in Visual Studio

oniliste:
blockwise_test.cc : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj
rithmetic_test.cc : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj


jan-wassenberg:
@oniliste 
Thanks for patching this! For consistency, would you mind moving this to replace  "# TODO(janwas): add flags", perhaps like:
```
  set(HWY_FLAGS
    # workaround for C1128 in blockwise_test
    /bigobj
  )
```
?

PR #628, opened 2022-04-07
fix build error  C1128 for MSVC

oniliste:
blockwise_test.cc : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj
rithmetic_test.cc : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj


jan-wassenberg:
@oniliste Super, thank you for making the change!

PR #651, opened 2022-04-21
hwy/highway_test.cc: include string.h for memcpy

thebombzen:
Fixes a compile error caused by implicit function declaration
of memcpy, which is defined in string.h.


jan-wassenberg:
@thebombzen Thanks!

PR #687, opened 2022-05-09
Wrap build system include directories with generator expression

madebr:
This allows highway targets to be exported at install time (by dependencies).
CMake does not allow build system directories for interface include directories.

https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_INCLUDE_DIRECTORIES.html#creating-relocatable-packages

Required by https://github.com/libsdl-org/SDL_image/pull/189
(libhighway is a dependency of libjxl, which is a dependency of sdl_image)


jan-wassenberg (approved):
@madebr Thank you :) Beginning the import process.

PR #732, opened 2022-06-04
fix CMakeLists.txt

mattn:


jan-wassenberg:
Hi @mattn, sorry I just saw this, we've already fixed this in the same way, but thank you for sending the patch :)

mattn:
Okay, closing

PR #780, opened 2022-06-20
preventing unused variable N_sn in debug build

mo271:


jan-wassenberg:
Oh, sorry, this code is anyway [changing](https://github.com/google/highway/commit/a2ebda3aee0800bff4a39576c7c815b961c453b2#diff-adcb5fe8d4a8b73787547a9bff2dd40a1d60e93147153747d8ed317dfc37d670) at the moment in a way that would also fix the warning :)

mo271:
ok, cool that it is fixed already!

PR #844, opened 2022-07-08
Add documentation for HWY_COMPILER_GCC macro

malaterre:


jan-wassenberg (approved):
Thanks! FYI I was considering later adding HWY_COMPILER_GCC_ACTUAL that's GCC only.

malaterre:
Unless you think `HWY_COMPILER_GCC` has been used in the wild (public macro?), I would have done on the entire project:

sed -i -e 's/HWY_COMPILER_GCC/HWY_COMPILER_GNU/g' *

And then add back an actual HWY_COMPILER_GCC macro...2cts

jan-wassenberg:
It's quite possible this one has been used elsewhere, so better not to change its behavior :)

Issue #846, opened 2022-07-08
base.h:249:19: error: _Float16 is not supported on this target


malaterre:
It seems I can no longer build on s390x:

* https://buildd.debian.org/status/fetch.php?pkg=highway&arch=s390x&ver=0.17.1%7Egit20220708.70d2be7-1&stamp=1657291435&raw=0

```
[1/103] /usr/bin/g++ -DHWY_SHARED_DEFINE -Dhwy_EXPORTS -I"/<<PKGBUILDDIR>>" -g -O2 -ffile-prefix-map=/<<PKGBUILDDIR>>=. -fstack-protector-strong -Wformat -Werror=format-security -DHWY_BROKEN_EMU128=0 -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -Wno-builtin-macro-redefined -D__DATE__=\"redacted\" -D__TIMESTAMP__=\"redacted\" -D__TIME__=\"redacted\" -fmerge-all-constants -Wall -Wextra -Wconversion -Wsign-conversion -Wvla -Wnon-virtual-dtor -fmath-errno -fno-exceptions -MD -MT CMakeFiles/hwy.dir/hwy/aligned_allocator.cc.o -MF CMakeFiles/hwy.dir/hwy/aligned_allocator.cc.o.d -o CMakeFiles/hwy.dir/hwy/aligned_allocator.cc.o -c '/<<PKGBUILDDIR>>/hwy/aligned_allocator.cc'
FAILED: CMakeFiles/hwy.dir/hwy/aligned_allocator.cc.o 
[...]

PR #847, opened 2022-07-09
Fix for GCC

Jamaika1:


jan-wassenberg:
Hi @Jamaika1 , thanks for sending. We will make a larger change that only uses _Float16 when required, instead of opting out like now. Please see #846.

PR #904, opened 2022-08-05
Run convert_test/TestIntFromFloatHuge again on neon

malaterre:


malaterre:
The above patch has been integrated in debian package since 7/72022 with no failures neither on gcc-11 nor on gcc-12:

* https://salsa.debian.org/debian-phototools-team/highway/-/blob/master/debian/patches/neon.patch

jan-wassenberg:
Nice, thanks for re-enabling it. Would you mind updating the comment before it as well (removing mention of Arm?)

malaterre:
would be nice to have this for 1.0.1. thanks

jan-wassenberg:
Thanks for following up :) I was hoping we'd also change the comment (which still mentions ArmV7 as being broken). I'll do that in a separate PR shortly.

jan-wassenberg:
Unfortunately this is still failing for us in clang + Arm7 32-bit. Let's re-enable only for GCC then.

malaterre:
closing since fixed in 0a9ec963b8478c45af963e78eb65e332fb4d07d7

Issue #907, opened 2022-08-05
64 bit integer arithmetic?


chriselrod:
For example, this works:
https://gcc.godbolt.org/z/Kn8jce6nb
but replacing `using T = int32_t` with `using T = int64_t` results in compilation errors.

I know that many architectures don't support it (e.g., only AVX3/AVX512 supports 64 bit integer multiplication), but it can be replicated with high/low sequences of 32 bit operations.

Would be nice to have this work automatically.

jan-wassenberg:
Hi @chriselrod, the good news is that SVE and RVV also support i64 * i64, so an argument can be made that previous ISAs should emulate the modern ones.

Would you be willing to send a patch for this? Here's one example of adding an op: https://github.com/google/highway/commit/18f98b9133e1c25633282f949d218fffec08b3aa
One could also use the approach from https://github.com/google/highway/commit/7d080f1dcdb798dd8661951aca3aa9b0a5dd352c to implement the fallback only once using generic ops (MulEven?), if that makes sense.

PR #908, opened 2022-08-05
Implement i64mul

chriselrod:
I'm experiencing compilation and test failures at the moment.

I have some experience with C++20, where I can use `requires`, `concepts`, and `if constexpr` for generic programming, but I'm unfamiliar with C++11 or SFINAE.

Odds are the approach is rather poor, and focused on x86. Suggestions would be appreciated.

Fixes https://github.com/google/highway/issues/907.

### DIFF
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -1250,7 +1250,7 @@ HWY_API V PopulationCount(V v) {
 #endif
 
 template <class V, class D = DFromV<V>, HWY_IF_LANE_SIZE_D(D, 8),
-          HWY_IF_LT128_D(D), HWY_IF_NOT_FLOAT(D)>
+          HWY_IF_LT128_D(D)>
 HWY_API V operator*(V x, V y) {
   return Set(D(), GetLane(x) * GetLane(y));
 }
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -16,7 +16,6 @@
 // Target-independent types/functions defined after target-specific ops.
 
 // Relies on the external include guard in highway.h.
-#include <cstdint>
 HWY_BEFORE_NAMESPACE();
 namespace hwy {
 namespace HWY_NAMESPACE {
@@ -1259,32 +1258,27 @@ HWY_API Vec128<uint64_t,1> operator*(Vec128<uint64_t,1> x, Vec128<uint64_t,1> y)
 
 #ifndef HWY_NATIVE_I64MULLO
 
+using V64 = Vec128<uint64_t>;
+using V32 = Vec128<uint32_t>;
 
-HWY_API Vec128<uint64_t> operator*(Vec128<uint64_t> x, Vec128<uint64_t> y) {
-  Vec128<uint32_t> x32 = BitCast(Simd<uint32_t,4,0>(), x);
-  Vec128<uint32_t> y32 = BitCast(Simd<uint32_t,4,0>(), y);
-  Vec128<uint64_t> lolo = MulEven(x32, y32);
-  Vec128<uint32_t> lohi = BitCast(Simd<uint32_t,4,0>(), MulEven(x32, BitCast(Simd<uint32_t,4,0>(), ShiftRight<32>(y))));
-  Vec128<uint32_t> hilo = BitCast(Simd<uint32_t,4,0>(), MulEven(BitCast(Simd<uint32_t,4,0>(), ShiftRight<32>(x)), y32));
-  return lolo + ShiftLeft<32>(BitCast(Simd<uint64_t,2,0>(), lohi + hilo));
+template <class V, typename T = LaneType<V>, HWY_IF_LANE_SIZE(T, 8), HWY_IF_UNSIGNED(T)>
+HWY_API V operator*(V x, V y) {
+  DFromV<V> d64;
+  RepartitionToNarrow<decltype(d64)> d32;
+  auto x32 = BitCast(d32, x);
+  auto y32 = BitCast(d32, y);
+  V lolo = MulEven(x32, y32);
+  auto lohi = BitCast(d32, MulEven(x32, BitCast(d32, ShiftRight<32>(y))));
+  auto hilo = BitCast(d32, MulEven(BitCast(d32, ShiftRight<32>(x)), y32));
+  return lolo + ShiftLeft<32>(BitCast(d64, lohi + hilo));
 }
-HWY_API Vec128<int64_t> operator*(Vec128<int64_t> x, Vec128<int64_t> y) {
-  return BitCast(Simd<int64_t,2,0>(), BitCast(Simd<uint64_t,2,0>(), x) * BitCast(Simd<uint64_t,2,0>(), y));
+template <class V, typename T = LaneType<V>, HWY_IF_LANE_SIZE(T, 8), HWY_IF_SIGNED(T)>
+HWY_API V operator*(V x, V y) {
+  DFromV<V> di64;
+  Repartition<uint64_t, decltype(di64)> du64;
+  return BitCast(di64, BitCast(du64, x) * BitCast(du64, y));
 }
-#if HWY_TARGET == HWY_AVX2
 
-HWY_API Vec256<uint64_t> operator*(Vec256<uint64_t> x, Vec256<uint64_t> y) {
-  Vec256<uint32_t> x32 = BitCast(Simd<uint32_t,8,0>(), x);
-  Vec256<uint32_t> y32 = BitCast(Simd<uint32_t,8,0>(), y);
-  Vec256<uint64_t> lolo = MulEven(x32, y32);
-  Vec256<uint32_t> lohi = BitCast(Simd<uint32_t,8,0>(), MulEven(x32, BitCast(Simd<uint32_t,8,0>(), ShiftRight<32>(y))));
-  Vec256<uint32_t> hilo = BitCast(Simd<uint32_t,8,0>(), MulEven(BitCast(Simd<uint32_t,8,0>(), ShiftRight<32>(x)), y32));
-  return lolo + ShiftLeft<32>(BitCast(Simd<uint64_t,4,0>(), lohi + hilo));
-}
-HWY_API Vec256<int64_t> operator*(Vec256<int64_t> x, Vec256<int64_t> y) {
-  return BitCast(Simd<int64_t,4,0>(), BitCast(Simd<uint64_t,4,0>(), x) * BitCast(Simd<uint64_t,4,0>(), y));
-}
-#endif 
 #endif // HWY_NATIVE_I64MULLO
 
 #endif  // HWY_NATIVE_POPCNT
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -1258,9 +1258,6 @@ HWY_API Vec128<uint64_t,1> operator*(Vec128<uint64_t,1> x, Vec128<uint64_t,1> y)
 
 #ifndef HWY_NATIVE_I64MULLO
 
-using V64 = Vec128<uint64_t>;
-using V32 = Vec128<uint32_t>;
-
 template <class V, typename T = LaneType<V>, HWY_IF_LANE_SIZE(T, 8), HWY_IF_UNSIGNED(T)>
 HWY_API V operator*(V x, V y) {
   DFromV<V> d64;
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -1249,6 +1249,46 @@ HWY_API V PopulationCount(V v) {
 }
 #endif
 
+#ifndef HWY_NATIVE_I64MULLO
+
+HWY_API Vec128<int64_t> operator*(Vec128<int64_t> x, Vec128<int64_t> y) {
+  Vec128<int32_t> x32 = BitCast(Simd<int32_t,4,0>(), x);
+  Vec128<int32_t> y32 = BitCast(Simd<int32_t,4,0>(), y);
+  Vec128<int64_t> lolo = BitCast(Simd<int64_t,2,0>(), MulEven(x32, y32));
+  Vec128<int32_t> lohi = BitCast(Simd<int32_t,4,0>(), MulEven(x32, BitCast(Simd<int32_t,4,0>(), ShiftRight<32>(y))));
+  Vec128<int32_t> hilo = BitCast(Simd<int32_t,4,0>(), MulEven(BitCast(Simd<int32_t,4,0>(), ShiftRight<32>(x)), x32));
+  return lolo + ShiftLeft<32>(BitCast(Simd<int64_t,2,0>(), lohi + hilo));
+}
+
+HWY_API Vec128<uint64_t> operator*(Vec128<uint64_t> x, Vec128<uint64_t> y) {
+  Vec128<uint32_t> x32 = BitCast(Simd<uint32_t,4,0>(), x);
+  Vec128<uint32_t> y32 = BitCast(Simd<uint32_t,4,0>(), y);
+  Vec128<uint64_t> lolo = BitCast(Simd<uint64_t,2,0>(), MulEven(x32, y32));
+  Vec128<uint32_t> lohi = BitCast(Simd<uint32_t,4,0>(), MulEven(x32, BitCast(Simd<uint32_t,4,0>(), ShiftRight<32>(y))));
+  Vec128<uint32_t> hilo = BitCast(Simd<uint32_t,4,0>(), MulEven(BitCast(Simd<uint32_t,4,0>(), ShiftRight<32>(x)), x32));
+  return lolo + ShiftLeft<32>(BitCast(Simd<uint64_t,2,0>(), lohi + hilo));
+}
+#if HWY_TARGET == HWY_AVX2
+HWY_API Vec256<int64_t> operator*(Vec256<int64_t> x, Vec256<int64_t> y) {
+  Vec256<int32_t> x32 = BitCast(Simd<int32_t,8,0>(), x);
+  Vec256<int32_t> y32 = BitCast(Simd<int32_t,8,0>(), y);
+  Vec256<int64_t> lolo = BitCast(Simd<int64_t,4,0>(), MulEven(x32, y32));
+  Vec256<int32_t> lohi = BitCast(Simd<int32_t,8,0>(), MulEven(x32, BitCast(Simd<int32_t,8,0>(), ShiftRight<32>(y))));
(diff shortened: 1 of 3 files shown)


chriselrod:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

I just signed it.

chriselrod:
Tests now pass locally on an AVX3 machine (skylake-avx512), which also tests down through SSE3.

Note that it is still missing SVE and RVV.

hwy/ops/generic_ops-inl.h
```diff
@@ -1249,6 +1250,43 @@ HWY_API V PopulationCount(V v) {
...
+}
+#if HWY_TARGET == HWY_AVX2
+
+HWY_API Vec256<uint64_t> operator*(Vec256<uint64_t> x, Vec256<uint64_t> y) {
+  Vec256<uint32_t> x32 = BitCast(Simd<uint32_t,8,0>(), x);
+  Vec256<uint32_t> y32 = BitCast(Simd<uint32_t,8,0>(), y);
+  Vec256<uint64_t> lolo = MulEven(x32, y32);
+  Vec256<uint32_t> lohi = BitCast(Simd<uint32_t,8,0>(), MulEven(x32, BitCast(Simd<uint32_t,8,0>(), ShiftRight<32>(y))));
+  Vec256<uint32_t> hilo = BitCast(Simd<uint32_t,8,0>(), MulEven(BitCast(Simd<uint32_t,8,0>(), ShiftRight<32>(x)), y32));
+  return lolo + ShiftLeft<32>(BitCast(Simd<uint64_t,4,0>(), lohi + hilo));
```
chriselrod, line 1282: The casts back to `uint32_t` before adding and casting again to `uint64_t` and shifting by 32 may look funny, because we'd get the exact same answer out of the exact same number of instructions by avoiding the casts on x86_64 by dropping the casts and doing the add on `uint64_t` directly.

I did this because I figured in theory doing 2X 32 bit adds should be cheaper than doing X 64 bit adds. While latencies and throughputs are the same, perhaps it uses slightly less power.
Also, on Haswell, `vpaddq` requires the complex decoder, while `vpaddd` does not. Thus in unrolled code (not operating out of the uop cache), `vpaddd` gets better throughput:
https://uops.info/html-instr/VPADDD_YMM_YMM_YMM.html
https://uops.info/html-instr/VPADDQ_YMM_YMM_YMM.html

hwy/ops/generic_ops-inl.h
```diff
@@ -16,6 +16,7 @@
 // Target-independent types/functions defined after target-specific ops.
 
 // Relies on the external include guard in highway.h.
+#include <cstdint>
```
jan-wassenberg, line 19: Please remove (I know that clang-tidy or other tools like to re-insert them).

hwy/ops/generic_ops-inl.h
```diff
@@ -1249,6 +1250,43 @@ HWY_API V PopulationCount(V v) {
...
+}
+HWY_API Vec128<uint64_t,1> operator*(Vec128<uint64_t,1> x, Vec128<uint64_t,1> y) {
+    return Set(Simd<uint64_t,1,0>(), GetLane(x)*GetLane(y));
+}
+
+#ifndef HWY_NATIVE_I64MULLO
+
+
+HWY_API Vec128<uint64_t> operator*(Vec128<uint64_t> x, Vec128<uint64_t> y) {
+  Vec128<uint32_t> x32 = BitCast(Simd<uint32_t,4,0>(), x);
```
jan-wassenberg, line 1264: To make this more readable I'd recommend a typedef V32 (`using V32 = Vec128<uint32_t>;`) and an lvalue for the Simd: `Full128<uint32_t> d32;`).

hwy/ops/generic_ops-inl.h
```diff
@@ -1249,6 +1250,43 @@ HWY_API V PopulationCount(V v) {
...
+  Vec128<uint32_t> lohi = BitCast(Simd<uint32_t,4,0>(), MulEven(x32, BitCast(Simd<uint32_t,4,0>(), ShiftRight<32>(y))));
+  Vec128<uint32_t> hilo = BitCast(Simd<uint32_t,4,0>(), MulEven(BitCast(Simd<uint32_t,4,0>(), ShiftRight<32>(x)), y32));
+  return lolo + ShiftLeft<32>(BitCast(Simd<uint64_t,2,0>(), lohi + hilo));
+}
+HWY_API Vec128<int64_t> operator*(Vec128<int64_t> x, Vec128<int64_t> y) {
+  return BitCast(Simd<int64_t,2,0>(), BitCast(Simd<uint64_t,2,0>(), x) * BitCast(Simd<uint64_t,2,0>(), y));
+}
+#if HWY_TARGET == HWY_AVX2
+
+HWY_API Vec256<uint64_t> operator*(Vec256<uint64_t> x, Vec256<uint64_t> y) {
```
jan-wassenberg, line 1276: I think this can share the same implementation as the Vec128 version if we make it a template with `<class V>` argument, where V is the arg and return type.
To get the `Simd<>` you can then use `DFromV<V> d32` and `RepartitionToWide<decltype(d32)> d64`.

(This assumes the Vec512 version is not a template, which it currently isn't. If it were, it might be ambiguous.)

hwy/ops/generic_ops-inl.h
jan-wassenberg: Interesting. I'd think the uop cache is the common case. To better predict performance, we can enable the llvm-mca tool in godbolt to predict throughput of an actual code block, not just one instruction.
[Example](https://gcc.godbolt.org/z/n-KcQ-)

I suspect that a 64-bit add is OK here.

hwy/tests/mul_test.cc
```diff
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <cstdint>
```
jan-wassenberg, line 16: Please remove.

hwy/tests/mul_test.cc
```diff
@@ -58,7 +59,7 @@ struct TestUnsignedMul {
     HWY_ASSERT_VEC_EQ(d, vmax, Mul(v1, vmax));
 
     const size_t bits = sizeof(T) * 8;
-    const uint64_t mask = (1ull << bits) - 1;
+    const uint64_t mask = bits==64 ? uint64_t(-1) : (1ull << bits) - 1;
```
jan-wassenberg, line 62: We can avoid warnings here by saying ~uint64_t{0}, or ~0ULL if you prefer.

hwy/ops/x86_512-inl.h
```diff
@@ -1164,15 +1164,33 @@ HWY_API Vec512<uint16_t> operator*(Vec512<uint16_t> a, Vec512<uint16_t> b) {
 HWY_API Vec512<uint32_t> operator*(Vec512<uint32_t> a, Vec512<uint32_t> b) {
   return Vec512<uint32_t>{_mm512_mullo_epi32(a.raw, b.raw)};
 }
-
+HWY_API Vec512<uint64_t> operator*(Vec512<uint64_t> a, Vec512<uint64_t> b) {
+  return Vec512<uint64_t>{_mm512_mullo_epi64(a.raw, b.raw)};
+}
+HWY_API Vec256<uint64_t> operator*(Vec256<uint64_t> a, Vec256<uint64_t> b) {
```
jan-wassenberg, line 1170: FYI normally we would define Vec256 and Vec128 overloads in x86_256/128, but this seems fine as-is because these overloads are only defined when we're targeting AVX3 or AVX3_DL, i.e. this header is included.

jan-wassenberg (changes requested):

jan-wassenberg:
Congrats, looks like good progress :) What are the compile errors with GCC? Happy to help figure out a workaround.

hwy/ops/generic_ops-inl.h
chriselrod: Note that llvm-mca does not model the front end, thus if the difference is expected to mostly be in the front end, we wouldn't expect llvm-mca to help.

I tried with a simple example, but I'd probably have to try a specific degenerate example to get integer adds to actually become the bottleneck.

hwy/ops/generic_ops-inl.h
chriselrod: How do I restrict it to only be for 64 bit integers?
```c++
template<class V>
HWY_API V operator*(V x, V y){
    ...
}
```
How do I restrict this to be called for `V` such that `LaneType<V>` is an 8 byte integer?

hwy/ops/x86_512-inl.h
chriselrod: Yes, they may be 256 or 128 bits, but they're AVX512 instructions / they require AVX3 and thus I didn't want them defined without it.

chriselrod:
> Congrats, looks like good progress :) What are the compile errors with GCC? Happy to help figure out a workaround.

I had used `runtests.sh` instead of the cmake build (I've since switched to the latter to make rebuilding much faster).
`runtests.sh` turns on Werror, causing this to throw:
```
In file included from /usr/lib/gcc/x86_64-redhat-linux/12/include/emmintrin.h:31,
                 from /home/chriselrod/Documents/progwork/cxx/simd/highway/hwy/ops/x86_128-inl.h:30,
                 from /home/chriselrod/Documents/progwork/cxx/simd/highway/hwy/highway.h:350,
                 from /home/chriselrod/Documents/progwork/cxx/simd/highway/hwy/highway_test.cc:27,
                 from /home/chriselrod/Documents/progwork/cxx/simd/highway/hwy/foreach_target.h:147,
                 from /home/chriselrod/Documents/progwork/cxx/simd/highway/hwy/highway_test.cc:26:
In function ‘void _mm_storel_pi(__m64*, __m128)’,
    inlined from ‘void hwy::N_SSSE3::Store(Vec128<float, 2>, Full64<float>, float*)’ at /home/chriselrod/Documents/progwork/cxx/simd/highway/hwy/ops/x86_128-inl.h:2133:16,
    inlined from ‘void hwy::N_SSSE3::TestSet::operator()(T, D) [with T = float; D = hwy::N_SSSE3::Simd<float, 2, 0>]’ at /home/chriselrod/Documents/progwork/cxx/simd/highway/hwy/highway_test.cc:140:10:
/usr/lib/gcc/x86_64-redhat-linux/12/include/xmmintrin.h:809:27: warning: ‘__Y’ may be used uninitialized [-Wmaybe-uninitialized]
  809 |   __builtin_ia32_storelps ((__v2sf *)__P, (__v4sf)__A);
      |   ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/lib/gcc/x86_64-redhat-linux/12/include/xmmintrin.h: In function ‘void hwy::N_SSSE3::TestSet::operator()(T, D) [with T = float; D = hwy::N_SSSE3::Simd<float, 2, 0>]’:
/usr/lib/gcc/x86_64-redhat-linux/12/include/xmmintrin.h:112:10: note: ‘__Y’ was declared here
  112 |   __m128 __Y = __Y;
      |          ^~~
In function ‘void _mm_storel_pi(__m64*, __m128)’,
    inlined from ‘void hwy::N_SSE4::Store(Vec128<float, 2>, Full64<float>, float*)’ at /home/chriselrod/Documents/progwork/cxx/simd/highway/hwy/ops/x86_128-inl.h:2133:16,
    inlined from ‘void hwy::N_SSE4::TestSet::operator()(T, D) [with T = float; D = hwy::N_SSE4::Simd<float, 2, 0>]’ at /home/chriselrod/Documents/progwork/cxx/simd/highway/hwy/highway_test.cc:140:10:
/usr/lib/gcc/x86_64-redhat-linux/12/include/xmmintrin.h:809:27: warning: ‘__Y’ may be used uninitialized [-Wmaybe-uninitialized]
  809 |   __builtin_ia32_storelps ((__v2sf *)__P, (__v4sf)__A);
      |   ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/lib/gcc/x86_64-redhat-linux/12/include/xmmintrin.h: In function ‘void hwy::N_SSE4::TestSet::operator()(T, D) [with T = float; D = hwy::N_SSE4::Simd<float, 2, 0>]’:
/usr/lib/gcc/x86_64-redhat-linux/12/include/xmmintrin.h:112:10: note: ‘__Y’ was declared here
  112 |   __m128 __Y = __Y;
      |          ^~~
```
This is also a problem on master, but the obvious work around is to not `-DHWY_WARNINGS_ARE_ERRORS:BOOL=ON`.

hwy/ops/generic_ops-inl.h
jan-wassenberg: Ah yes, I should have mentioned that :)
`template <class V, HWY_IF_LANE_SIZE_V(V, 8)>` will do that.

hwy/ops/x86_512-inl.h
jan-wassenberg: Right. Usually the way we specialize AVX3 implementations of 256-bit vectors is to have one function in x86_256 with `#if HWY_TARGET <= HWY_AVX3` inside, but as mentioned it's fine as-is in this case.

jan-wassenberg:
Oh, I had hoped the recent https://github.com/google/highway/commit/c06d11ff88801f90c9eceb40ae6448e10777b370 fixes that, but I see now we require the same patch there as in x86_256. Sending that shortly.

hwy/ops/generic_ops-inl.h
chriselrod: I looked through the source, and tried to follow the approach you took elsewhere.

hwy/ops/generic_ops-inl.h
```diff
@@ -1249,6 +1249,35 @@ HWY_API V PopulationCount(V v) {
...
+  auto y32 = BitCast(d32, y);
+  V lolo = MulEven(x32, y32);
+  auto lohi = BitCast(d32, MulEven(x32, BitCast(d32, ShiftRight<32>(y))));
+  auto hilo = BitCast(d32, MulEven(BitCast(d32, ShiftRight<32>(x)), y32));
+  return lolo + ShiftLeft<32>(BitCast(d64, lohi + hilo));
+}
+template <class V, typename T = LaneType<V>, HWY_IF_LANE_SIZE(T, 8), HWY_IF_SIGNED(T)>
+HWY_API V operator*(V x, V y) {
+  DFromV<V> di64;
+  Repartition<uint64_t, decltype(di64)> du64;
```
jan-wassenberg, line 1275: Minor: Repartition is intended for changes of lane size, in this case we want the same number of lanes, so you can use `RebindToUnsigned<decltype(di64)> du64`.

jan-wassenberg:
Nice, almost there, just one minor comment. If you prefer, I can take care of SVE/RVV, or give some pointers if you'd prefer to?

jan-wassenberg:
Wow, that was quick :) Ready to start importing this change if you're happy with it?

jan-wassenberg (approved):

chriselrod:
> Ready to start importing this change if you're happy with it?

Thanks, I'm happy with it. You're welcome to make the SVE/RVV change. If they're not already supported, I'm guessing it's just changing some macro invocation?

chriselrod:
I forgot a semicolon. Rerunning tests locally.

jan-wassenberg:
Sure, no problem. I will add the SVE/RVV, it's just a matter of adding some overloads in arm_sve.h etc.
Thanks for implementing this!

jan-wassenberg:
@chriselrod our CI has caught an issue with the HWY_SCALAR target, which lacks Vec128. Would you mind changing the two overloads for single-entry vectors to the following?
```
template <class V, class D = DFromV<V>, HWY_IF_LANE_SIZE_D(D, 8),
          HWY_IF_LT128_D(D)>
HWY_API V operator*(V x, V y) {
  return Set(D(), GetLane(x) * GetLane(y));
}
```

chriselrod:
Does it need any sort of integer check?

jan-wassenberg:
Did you mean `HWY_IF_NOT_FLOAT_D(D)` instead of `HWY_IF_NOT_FLOAT(D)`? I would expect the latter to be a no-op. Do we require it at all?

chriselrod:
> Did you mean `HWY_IF_NOT_FLOAT_D(D)` instead of `HWY_IF_NOT_FLOAT(D)`? I would expect the latter to be a no-op. Do we require it at all?

I'll try dropping it.

chriselrod:
We can about the `typename T = LaneType<V>` via
```c++
template <class V, class D64 = DFromV<V>, HWY_IF_LANE_SIZE_D(D64, 8),
	  HWY_IF_UNSIGNED_D(D64), HWY_IF_GE128_D(D64)>
HWY_API V operator*(V x, V y) {
  RepartitionToNarrow<D64> d32;
  auto x32 = BitCast(d32, x);
  auto y32 = BitCast(d32, y);
  auto lolo = BitCast(d32, MulEven(x32, y32));
  auto lohi = BitCast(d32, MulEven(x32, BitCast(d32, ShiftRight<32>(y))));
  auto hilo = BitCast(d32, MulEven(BitCast(d32, ShiftRight<32>(x)), y32));
  auto hi = BitCast(d32, ShiftLeft<32>(BitCast(D64{}, lohi + hilo)));
  return BitCast(D64{}, lolo + hi);
}
template <class V, class DI64 = DFromV<V>, HWY_IF_LANE_SIZE_D(DI64, 8),
	  HWY_IF_SIGNED_D(DI64), HWY_IF_GE128_D(DI64)>
HWY_API V operator*(V x, V y) {
  RebindToUnsigned<DI64> du64;
  return BitCast(DI64{}, BitCast(du64, x) * BitCast(du64, y));
}
```
Is this preferred?

jan-wassenberg:
Your latest change looks great :) But let's give our internal CI (rather slow) a chance to finish testing the dozens of configurations and see if tests pass with that.

chriselrod:
> Your latest change looks great :) But let's give our internal CI (rather slow) a chance to finish testing the dozens of configurations and see if tests pass with that.

Latest change meaning my comment (which passed locally, but I didn't push yet), or the remote/this PR?

jan-wassenberg:
Sorry to be unclear :) I like what you posted right before "Is this preferred?" but IMO the removal of typename T is not enough benefit to rerun all the tests, so let's go with what is currently being tested (https://github.com/google/highway/pull/908/commits/a3795f822718c4da70d93cad8e0f5f4689ebdf00) if that passes.

chriselrod:
Sounds good!

jan-wassenberg:
Thanks for updating your patch! This has now landed and I've updated your change to support SVE/RVV.

BTW I gained some experience on how our iterate-on-Github-change workflow is meant to work, the CI will be faster in future 👍

PR #922, opened 2022-08-11
Unsigned integer to floating-point conversion

### LINKED ISSUE #917: Missing unsigned integer => floating-point conversions
The current implementation of `ConvertTo` does not have an overload for unsigned integer to floating-point conversion. For my specific application I need to convert uint32 => float32 and uint64 => float64.

I'm willing to try implementing it myself, but suggestions are welcome as I am not (yet) an expert on CPU intrinsics.

nielskm:
Implementing unsigned integer to floating-point conversions. This fixes #917 .

### DIFF
--- a/hwy/ops/arm_sve-inl.h
+++ b/hwy/ops/arm_sve-inl.h
@@ -1502,6 +1502,11 @@ HWY_API svint32_t DemoteTo(Simd<int32_t, N, kPow2> d, const svfloat64_t v) {
       NAME(HWY_SVE_D(BASE, BITS, N, kPow2) /* d */, HWY_SVE_V(int, BITS) v) { \
     return sv##OP##_##CHAR##BITS##_s##BITS##_x(HWY_SVE_PTRUE(BITS), v);       \
   }                                                                           \
+  template <size_t N, int kPow2>                                              \
+  HWY_API HWY_SVE_V(BASE, BITS)                                               \
+      NAME(HWY_SVE_D(BASE, BITS, N, kPow2) /* d */, HWY_SVE_V(int, BITS) v) { \
+    return sv##OP##_##CHAR##BITS##_u##BITS##_x(HWY_SVE_PTRUE(BITS), v);       \
+  }                                                                           \
   /* Truncates (rounds toward zero). */                                       \
   template <size_t N, int kPow2>                                              \
   HWY_API HWY_SVE_V(int, BITS)                                                \
--- a/hwy/ops/arm_sve-inl.h
+++ b/hwy/ops/arm_sve-inl.h
@@ -1504,13 +1504,13 @@ HWY_API svint32_t DemoteTo(Simd<int32_t, N, kPow2> d, const svfloat64_t v) {
 // ------------------------------ ConvertTo F
 
 #define HWY_SVE_CONVERT(BASE, CHAR, BITS, HALF, NAME, OP)                     \
-  /* signed integers */
+  /* signed integers */                                                       \
   template <size_t N, int kPow2>                                              \
   HWY_API HWY_SVE_V(BASE, BITS)                                               \
       NAME(HWY_SVE_D(BASE, BITS, N, kPow2) /* d */, HWY_SVE_V(int, BITS) v) { \
     return sv##OP##_##CHAR##BITS##_s##BITS##_x(HWY_SVE_PTRUE(BITS), v);       \
   }                                                                           \
-  /* unsigned integers */
+  /* unsigned integers */                                                     \
   template <size_t N, int kPow2>                                              \
   HWY_API HWY_SVE_V(BASE, BITS)                                               \
       NAME(HWY_SVE_D(BASE, BITS, N, kPow2) /* d */, HWY_SVE_V(uint, BITS) v) { \
--- a/hwy/ops/arm_sve-inl.h
+++ b/hwy/ops/arm_sve-inl.h
@@ -1504,14 +1504,16 @@ HWY_API svint32_t DemoteTo(Simd<int32_t, N, kPow2> d, const svfloat64_t v) {
 // ------------------------------ ConvertTo F
 
 #define HWY_SVE_CONVERT(BASE, CHAR, BITS, HALF, NAME, OP)                     \
+  /* signed integers */
   template <size_t N, int kPow2>                                              \
   HWY_API HWY_SVE_V(BASE, BITS)                                               \
       NAME(HWY_SVE_D(BASE, BITS, N, kPow2) /* d */, HWY_SVE_V(int, BITS) v) { \
     return sv##OP##_##CHAR##BITS##_s##BITS##_x(HWY_SVE_PTRUE(BITS), v);       \
   }                                                                           \
+  /* unsigned integers */
   template <size_t N, int kPow2>                                              \
   HWY_API HWY_SVE_V(BASE, BITS)                                               \
-      NAME(HWY_SVE_D(BASE, BITS, N, kPow2) /* d */, HWY_SVE_V(int, BITS) v) { \
+      NAME(HWY_SVE_D(BASE, BITS, N, kPow2) /* d */, HWY_SVE_V(uint, BITS) v) { \
     return sv##OP##_##CHAR##BITS##_u##BITS##_x(HWY_SVE_PTRUE(BITS), v);       \
   }                                                                           \
   /* Truncates (rounds toward zero). */                                       \
--- a/hwy/ops/x86_128-inl.h
+++ b/hwy/ops/x86_128-inl.h
@@ -5593,9 +5593,21 @@ HWY_API Vec128<float, N> ConvertTo(Simd<float, N, 0> /* tag */,
 }
 
 template <size_t N>
-HWY_API Vec128<float, N> ConvertTo(Simd<float, N, 0> /* tag */,
+HWY_API Vec128<float, N> ConvertTo(HWY_MAYBE_UNUSED Simd<float, N, 0> dd,
                                    const Vec128<uint32_t, N> v) {
+#if HWY_TARGET <= HWY_AVX3
   return Vec128<float, N>{_mm_cvtepu32_ps(v.raw)};
+#else
+  // Based on wim's approach (https://stackoverflow.com/questions/34066228/)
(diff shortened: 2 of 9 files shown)


hwy/ops/arm_sve-inl.h
```diff
@@ -1509,6 +1509,11 @@ HWY_API svint32_t DemoteTo(Simd<int32_t, N, kPow2> d, const svfloat64_t v) {
       NAME(HWY_SVE_D(BASE, BITS, N, kPow2) /* d */, HWY_SVE_V(int, BITS) v) { \
     return sv##OP##_##CHAR##BITS##_s##BITS##_x(HWY_SVE_PTRUE(BITS), v);       \
   }                                                                           \
+  template <size_t N, int kPow2>                                              \
```
jan-wassenberg, line 1512: Would be nice to add a comment `/* unsigned */` and perhaps also above `/* signed */`, it wasn't immediately visible what this change was doing.

hwy/ops/x86_128-inl.h
```diff
@@ -5592,6 +5592,24 @@ HWY_API Vec128<float, N> ConvertTo(Simd<float, N, 0> /* tag */,
   return Vec128<float, N>{_mm_cvtepi32_ps(v.raw)};
 }
 
+template <size_t N>
+HWY_API Vec128<float, N> ConvertTo(HWY_MAYBE_UNUSED Simd<float, N, 0> dd,
```
jan-wassenberg, line 5596: Optional: we usually use df float float, perhaps dd for double. Might be nice to rename to df.

hwy/ops/x86_128-inl.h
```diff
@@ -5592,6 +5592,24 @@ HWY_API Vec128<float, N> ConvertTo(Simd<float, N, 0> /* tag */,
...
+                                   const Vec128<uint32_t, N> v) {
+#if HWY_TARGET <= HWY_AVX3
+  return Vec128<float, N>{_mm_cvtepu32_ps(v.raw)};
+#else
+  // Based on wim's approach (https://stackoverflow.com/questions/34066228/)
+  const Repartition<uint32_t, decltype(dd)> du32;
```
jan-wassenberg, line 5602: This will work but `RebindToUnsigned<decltype(dd)>` would be a bit more clear and emphasizes we're not changing the number of lanes.

hwy/ops/x86_128-inl.h
```diff
@@ -5592,6 +5592,24 @@ HWY_API Vec128<float, N> ConvertTo(Simd<float, N, 0> /* tag */,
...
+  const Repartition<int32_t, decltype(dd)> d32;
+
+  const auto msk_lo = Set(du32, 0xFFFF);
+  const auto cnst2_16_flt = Set(dd, 65536.0f); /* 2^16 */
```
jan-wassenberg, line 5606: Please use // comments (unless we're in a macro, where /* is required) and put longer comments on their own line - our autoformatter will enforce an 80 char line limit, which would break up the "Extract" comment below.

hwy/ops/x86_128-inl.h
```diff
@@ -5616,6 +5634,32 @@ HWY_API Vec128<double, N> ConvertTo(Simd<double, N, 0> dd,
...
 }
 
+template <size_t N>
+HWY_API Vec128<double, N> ConvertTo(HWY_MAYBE_UNUSED Simd<double, N, 0> dd,
+                                    const Vec128<uint64_t, N> v) {
+#if HWY_TARGET <= HWY_AVX3
+  return Vec128<double, N>{_mm_cvtepu64_pd(v.raw)};
+#else
+  // Based on wim's approach (https://stackoverflow.com/questions/41144668/)
+  const Repartition<uint64_t, decltype(dd)> d64;
```
jan-wassenberg, line 5644: Also consider `RebindToUnsigned<decltype(dd)>` here.

hwy/ops/x86_128-inl.h
```diff
@@ -5616,6 +5634,32 @@ HWY_API Vec128<double, N> ConvertTo(Simd<double, N, 0> dd,
...
+
+  const auto msk_lo = Set(d64, 0xFFFFFFFF);
+  const auto cnst2_32_dbl = Set(dd, 4294967296.0); /* 2^32 */
+  const auto v_lo = And(v, msk_lo); /* Extract the 32 lowest significant bits of v */
+  const auto v_hi = ShiftRight<32>(v);
+
+  using vec_type = typename std::decay<decltype(v_lo)>::type;
```
jan-wassenberg, line 5651: You can also say `using VU = VFromD<decltype(d64)>;`. Would be nice to then use this VU type instead of auto for a bit more clarity/readability.

hwy/ops/x86_256-inl.h
```diff
@@ -4207,6 +4207,47 @@ HWY_API Vec256<double> ConvertTo(Full256<double> dd, const Vec256<int64_t> v) {
 #endif
 }
 
+HWY_API Vec256<float> ConvertTo(HWY_MAYBE_UNUSED Full256<float> dd,
+                                const Vec256<uint32_t> v) {
+#if HWY_TARGET <= HWY_AVX3
+  return Vec256<float>{_mm256_cvtepu32_ps(v.raw)};
+#else
+  // Based on wim's approach (https://stackoverflow.com/questions/34066228/)
+  const Repartition<uint32_t, decltype(dd)> du32;
```
jan-wassenberg, line 4216: Also RebindToUnsigned/Signed here?

hwy/ops/x86_256-inl.h
```diff
@@ -4207,6 +4207,47 @@ HWY_API Vec256<double> ConvertTo(Full256<double> dd, const Vec256<int64_t> v) {
 #endif
 }
 
+HWY_API Vec256<float> ConvertTo(HWY_MAYBE_UNUSED Full256<float> dd,
```
jan-wassenberg, line 4210: df?

hwy/tests/convert_test.cc
```diff
@@ -532,6 +532,32 @@ HWY_NOINLINE void TestAllFloatFromInt() {
   ForFloatTypes(ForPartialVectors<TestFloatFromInt>());
 }
 
+struct TestFloatFromUint {
+  template <typename TF, class DF>
+  HWY_NOINLINE void operator()(TF /*unused*/, const DF df) {
+    using TI = MakeSigned<TF>;
```
jan-wassenberg, line 538: Probably want MakeUnsigned here so that we're testing the full unsigned range?

jan-wassenberg (changes requested):

hwy/ops/x86_256-inl.h
```diff
@@ -4207,6 +4207,47 @@ HWY_API Vec256<double> ConvertTo(Full256<double> dd, const Vec256<int64_t> v) {
...
+  const auto cnst2_16_flt = Set(dd, 65536.0f); /* 2^16 */
+  const auto v_lo = BitCast(d32, And(v, msk_lo)); /* Extract the 16 lowest significant bits of v and cast to signed int */
+  const auto v_hi = BitCast(d32, ShiftRight<16>(v));
+  return MulAdd(cnst2_16_flt, ConvertTo(dd, v_hi), ConvertTo(dd, v_lo));
+#endif
+}
+
+HWY_API Vec256<double> ConvertTo(HWY_MAYBE_UNUSED Full256<double> dd,
+                                  const Vec256<uint64_t> v) {
+#if HWY_TARGET <= HWY_AVX3
```
chriselrod, line 4229: Should this be `>=`, or is what the comparison does unintuitive?
I.e., without AVX3 (e.g. only AVX2) you'd need to run the fallback. Only AVX3(+) provides `_mm256_cvtepu64_pd`.

nielskm: Yes, but the highest-priority target has the smallest numerical value :-)

hwy/ops/arm_sve-inl.h
nielskm: Done. And thank you for pointing my attention to this file - there was a bug.

hwy/ops/x86_256-inl.h
nielskm: I was also confused about it at first.

jan-wassenberg: That's right :) Thanks for raising this point, I will document it and explain why `<= HWY_AVX3` works and is desirable.

jan-wassenberg (approved):
Super, thanks for making these changes. Looks good, I will launch the import/testing process :)

hwy/ops/x86_256-inl.h
```diff
@@ -4207,6 +4207,53 @@ HWY_API Vec256<double> ConvertTo(Full256<double> dd, const Vec256<int64_t> v) {
...
+  return Vec256<double>{_mm256_cvtepu64_pd(v.raw)};
+#else
+  // Based on wim's approach (https://stackoverflow.com/questions/41144668/)
+  const RebindToUnsigned<decltype(dd)> d64;
+  using VU = VFromD<decltype(d64)>;
+
+  const VU msk_lo = Set(d64, 0xFFFFFFFF);
```
jan-wassenberg, line 4239: Should we make this 0xFFFFFFFFull?

jan-wassenberg:
The internal tests found some issues. Sorry I didn't catch them during review, but it's nice to have an additional safety net.

1) A quick one: the lambda in x86_128 and x86_256 requires HWY_ATTR before the opening brace (this is a compiler limitation briefly mentioned in the readme).
2) for arm_sve, a backslash character is still required after the new /* */ comments.
3) it looks like we might be stumbling over an MSVC compiler bug. It causes the f64x4 (AVX2 target) test to fail - not sure about f32 because that comes later. As annoying as it is to repeat code, it might be caused by the lambda. Maybe we can try inlining the macro (we can also declare an lvalue for the constants)? Or feel free to wrap the new `HWY_EXPORT_AND_TEST_P` in `#if !HWY_COMPILER_MSVC` and I can help debug/work around this.

hwy/ops/x86_256-inl.h
nielskm: Let's :-)

jan-wassenberg (approved):
Nice, thanks for making the changes. I'll rerun the tests :)

PR #934, opened 2022-08-19
Fixing unroll pragma for GCC

### LINKED ISSUE #933: GCC unknown pragma '#pragma unroll'
The added `#pragma unroll` in `vqsort-inl.h` is not supported by GCC. Instead use `#pragma GCC unroll <value>`. I will make a PR shortly.

nielskm:
- Changing `#pragma unroll` to `#pragma GCC unroll 4` for GCC compiler. The value of 4 can be changed if needed.

This fixes #933 .

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -138,8 +138,11 @@
 // nielskm: GCC does not support '#pragma GCC unroll' without the factor.
 #define HWY_UNROLL(factor) HWY_PRAGMA(GCC unroll factor)
 #define HWY_DEFAULT_UNROLL HWY_UNROLL(4)
-#else
+#elif HWY_COMPILER_CLANG || HWY_COMPILER_ICC || HWY_COMPILER_ICX
 #define HWY_UNROLL(factor) HWY_PRAGMA(unroll factor)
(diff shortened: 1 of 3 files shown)


jan-wassenberg:
Thanks for the fix! Might be useful in other places, too. Would you like to add a `#define HWY_UNROLL(factor)` to base.h?
Something like: `#if HWY_COMPILER_GCC_ACTUAL` then `HWY_PRAGMA(GCC unroll factor)` `#elif HWY_COMPILER_CLANG` then `HWY_PRAGMA(unroll factor)` else expand to nothing?

I'm also happy to do this if you prefer.

nielskm:
@jan-wassenberg Thanks for the suggestion - I just pushed the changes. One thing to consider is that GCC does not support unrolling without a factor, while e.g. clang does (it tries to unroll the entire loop if the number of iterations is known at compile time and presumably has a good default in other cases). I therefore implemented both `HWY_UNROLL(factor)` and a `HWY_DEFAULT_UNROLL`, where the latter uses factor=4 for GCC.

jan-wassenberg (approved):
Nice, thanks for upgrading this :)

PR #961, opened 2022-09-02
Replace rdcycle instruction with rdtime

malaterre:
The original rdcycle intent was for timestamp. In RISC-V the proper
equivalent of x86 "rdtsc" is "rdtime".

* http://lists.infradead.org/pipermail/linux-riscv/2022-September/018862.html

### DIFF
--- a/hwy/nanobenchmark.cc
+++ b/hwy/nanobenchmark.cc
@@ -149,11 +149,8 @@ inline Ticks Start() {
       // "memory" avoids reordering. rdx = TSC >> 32.
       // "cc" = flags modified by SHL.
       : "rdx", "memory", "cc");
-  // TODO(janwas): the cycle counter and even the timer CSR are no longer in the
-  // base spec and are part of the Zicntr extension, which is not yet ratified
-  // as of 2022-09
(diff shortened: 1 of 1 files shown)


jan-wassenberg:
Let's discuss this one further via email. For benchmarking we really would prefer a cycle count, rather than "real time clock ticks", for which the spec makes no guarantees. If/when those turn out to be millisecond-resolution, they are not very useful.

malaterre:
For reference, this `rdtime` patch has been tested on:

```
[    0.000000] Linux version 5.18.0-2-riscv64 (debian-kernel@lists.debian.org) (gcc-11 (Debian 11.2.0-20) 11.2.0, GNU ld (GNU Binutils for Debian) 2.38) #1 SMP Debian 5.18.5-1 (2022-06-16)
[    0.000000] Machine model: SiFive HiFive Unmatched A00
```

`rdcycle` gives a SIGILL in this particular combo.

jan-wassenberg (approved):
Per email discussion, it should be possible to regain access to the cycle counter. Until then, let's merge this so your build will be green again.

PR #978, opened 2022-09-09
Add vcpkg installation instructions

JonLiu1993:
highway is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for highway and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build highway, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/highway/portfile.cmake). We try to keep the library maintained as close as possible to the original library. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,18 @@ Or you can run `run_tests.sh` (`run_tests.bat` on Windows).
 
 Bazel is also supported for building, but it is not as widely used/tested.
 
+## Building highway - Using vcpkg
+
+You can download and install highway using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    ./vcpkg install highway
+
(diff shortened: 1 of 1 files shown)


jan-wassenberg:
@JonLiu1993 Oh, great to hear Highway is in vcpkg, thanks for reaching out :)

We'd be happy to add this to the readme, are you able to sign the CLA?

jan-wassenberg:
Hi @JonLiu1993, we'd be happy to add this to the readme but for 100% correctness would require the CLA. Are you able to sign it at https://cla.developers.google.com/ ?

JonLiu1993:
@jan-wassenberg, I'd love to contribute to you guys, I signed the CLA a long time ago, but for some reason I'm still showing unsigned here, do you know why?
image

jan-wassenberg:
Nice, thank you @JonLiu1993  for signing it. Usually the cause is that the email used for signing (can be seen [here](https://cla.developers.google.com/clas)) differs from that in the commit, which looks to have been "<137******33​@163.com>".

Issue #988, opened 2022-09-19
Debian: executable stack warning


boxerab:
Hi! I notice that the [debian package](https://tracker.debian.org/pkg/highway) for highway has the following lintian warnings :

shlib-with-executable-stack

https://udd.debian.org/lintian/?packages=highway

Is this due to dynamic execution of different SIMD instruction sets at runtime, based on actual architecture ?
It seems to only be a problem on mips hardware.

jan-wassenberg:
Hi, interesting. I'd be happy to help fix this, but am not sure why it's happening.
The [docs](https://lintian.debian.org/tags/executable-stack-in-shared-library) mention that assembly language files might be a cause (and indeed another project fixed that [like so](https://github.com/intel/intel-ipsec-mb/commit/94f3ebb21e09e9923dfc28e8af0ddda405e67d20)), but we do not have any standalone .s files.

Does anyone have any idea what might be causing this?

jan-wassenberg:
Oh, interesting. That seems reasonable. Would you like to send a PR for this?
Would be good to rename ARCH to HWY_ARCH to reduce likelihood of clashes when included as a cmake subdirectory.

jrtc27:
This is because, prior to 4.8.0, kernel FPU emulation abused the process's stack to execute code in the user context (similar to how in the old days sigreturn trampolines would be put on the stack rather than use the vDSO or equivalent) and Debian's glibc is configured with upstream's default baseline, older than 4.8.0. Using `-z noexecstack` will override that and should be safe so long as you're not linking against something which specifically needs an executable stack (rare these days) and have 4.8.0 or above as your baseline. See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1022787.

jan-wassenberg:
Thanks @jrtc27 for the background! Sounds like you are saying @boxerab's workaround is fine (even if lintian shouldn't have warned about it) and no further action is required at the moment.

PR #991, opened 2022-09-19
build: detect architecture and add noexecstack link flag for mips

boxerab:
see #988 for details on why this fix is necessary.

Verified on `mipsel` chroot that this patch removes the `E` flag from the stack for the shared library build.


jan-wassenberg (changes requested):
We have one remaining instance of ARCH (line 298), let's also rename to HWY_ARCH :)
(Hard to see in the verbose cmake language..)

jan-wassenberg (approved):

PR #995, opened 2022-09-22
cmake: make architecture message a STATUS message

boxerab:
Default message seems to be ERROR message, appearing in log in red


jan-wassenberg (approved):

jan-wassenberg:
FYI this hasn't gone in yet because our Copybara workflow had a conflict (attempted to change same file at the same time).

PR #996, opened 2022-09-22
cmake: add option to enable all tests

boxerab:
This option is for projects that include highway as a subproject and do not want to build highway tests or download dependencies


jan-wassenberg (changes requested):
Would you mind moving the set() to line 54 with the others, and changing the sense from DISABLE to ENABLE (with default value ON) for consistency?

jan-wassenberg (approved):
Great, thanks for making the change :)

PR #1008, opened 2022-10-05
cmake: Detect whether or not atomic is needed

### LINKED ISSUE #1003: -latomic not present on x86 platforms
Since #889, the build seems to no longer work on x86 GCC10, where it previously did build just fine.

/opt/rh/devtoolset-10/root/usr/libexec/gcc/x86_64-redhat-linux/10/ld: cannot find -latomic

malaterre:
It turns out that GCC requires explicitly linking to a library `atomic` in order to support c11/atomics. gcc spec file will not handle it directly at least not on the following Debian archs: armel, m68k, mipsel, powerpc, sh4, sparc64 and riscv64.

Introduce a new cmake module to detect usage of gcc/atomic and add missing library to the link step.

For reference:
* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104248

Fixes #1003

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,10 @@ set(LIBRARY_SOVERSION ${hwy_VERSION_MAJOR})
 
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+# Search for Atomics implementation:
+find_package(Atomics REQUIRED)
+
 # Enabled PIE binaries by default if supported.
 include(CheckPIESupported OPTIONAL RESULT_VARIABLE CHECK_PIE_SUPPORTED)
 if(CHECK_PIE_SUPPORTED)
@@ -281,11 +285,9 @@ target_include_directories(hwy PUBLIC
 target_compile_features(hwy PUBLIC cxx_std_11)
 set_target_properties(hwy PROPERTIES
   LINK_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/hwy/hwy.version)
+# For GCC __atomic_store_8, see #887
+target_link_libraries(hwy PRIVATE ${ATOMICS_LIBRARIES})
 if(UNIX AND NOT APPLE)
-  # For GCC __atomic_store_8, see #887
-  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND NOT HWY_EMSCRIPTEN)
-    target_link_libraries(hwy atomic)
-  endif()
   # not supported by MSVC/Clang, safe to skip (we use DLLEXPORT annotations)
   set_property(TARGET hwy APPEND_STRING PROPERTY
     LINK_FLAGS " -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/hwy/hwy.version")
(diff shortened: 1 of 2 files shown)


jan-wassenberg:
Nice, thanks for adding this. It looks like the clang-5 builder/check is still [failing](https://github.com/google/highway/actions/runs/3189784765/jobs/5204006133), though it does seem fine on [godbolt](https://gcc.godbolt.org/z/P9PT7fWPf). Any ideas on how we might work around that?

malaterre:
@jan-wassenberg I cannot get my hand on a clang-5 compiler. Any way you could add something like:

```
cat /home/runner/work/highway/highway/out/CMakeFiles/CMakeOutput.log
cat /home/runner/work/highway/highway/out/CMakeFiles/CMakeError.log
```

somewhere in your CI setup (in case of error)

malaterre:
@jan-wassenberg Two things:

1. I seems the CI has stopped working on this PR
2. This code has been working nicely in libjxl: https://github.com/libjxl/libjxl/commit/87fe7c16e1fb2e21b6a1dca26782950ae1559d99

I agree that on libjxl I used `set(CMAKE_CXX_STANDARD 11)` instead of a direct `-std=c++11` but I fail to see why it would make any difference.

malaterre:
lol, someone is adding -std=c++98 in my back:

```
/usr/bin/clang++-5.0 -DATOMICS_IN_LIBRARY  -Werror -std=c++11 -fPIE -std=c++98 -MD -MT CMakeFiles/cmTC_16d49.dir/src.cxx.o -MF CMakeFiles/cmTC_16d49.dir/src.cxx.o.d -o CMakeFiles/cmTC_16d49.dir/src.cxx.o -c /home/runner/work/highway/highway/out/CMakeFiles/CMakeTmp/src.cxx
```

@jan-wassenberg any comment ?

jan-wassenberg:
Good old CMake, it often makes a mess, sigh.

Very difficult to debug without access to the logs, that's a great trick with reading+printing it, nice!

It seems to me that CMAKE_CXX_STANDARD would be the correct way to avoid the problem, but it's only supported in 3.1. There's also target_compile_features but that seems to only be for targets (not sure it would work with try_compile), and it's also 3.1 only. I wonder if the CMake running is older than 3.1? The logs don't seem to mention CMake version.

malaterre:
FYI:

* https://discourse.cmake.org/t/cmake-cxx-standard-cmake-3-16-vs-3-24/6623

jan-wassenberg:
hm, so a CMake misbehavior indeed. Nice that you've raised it with them, let's see if there's a workaround.

malaterre:
@jan-wassenberg Seems like this one is not my fault:

`vqsort-inl.h:73:49: error: unused parameter 'p' [-Werror,-Wunused-parameter]`

jan-wassenberg:
Oh, indeed :) Fixing, thanks for pointing that out. (Our internal CI is still somehow suppressing unused arg warnings.)

I appreciate your heroic efforts to get this working, looks like we're almost there?

jan-wassenberg:
:) The fix is in.

jan-wassenberg:
Congrats, seems you have found a decent workaround for the CMake issue. Are you happy with the current code, should we merge it?

malaterre:
> Are you happy with the current code, should we merge it?

I believe this is a good improvement. As mentionned in the comment, I still need to dig in cmake to understand the difference in behavior between 3.16 vs 3.24, but I suspect this should not hold this PR. So please merge, thanks !

jan-wassenberg (approved):

jan-wassenberg:
"GitHub Actions has encountered an internal error when running your job."
This seems to happen often:
https://github.com/actions/runner-images/issues/6131
https://github.com/actions/runner-images/issues/2179
https://github.com/actions/runner-images/issues/4627
https://github.com/actions/runner-images/issues/2579

PR #1019, opened 2022-10-17
neon: fix i16-based version of ReorderDemote2To

kleisauke:
The `hi`/`lo` arguments passed to the `Combine` operation were reversed.

This was noticed while testing the [`simd-highway`](https://github.com/kleisauke/libvips/tree/simd-highway) branch of libvips, which caused an RGB -> BRG conversion during thumbnailing on ARMv7.

Test case:
<details>
  <summary>Details</summary>

```cpp
#include <cstdint>
#include <iostream>

#undef HWY_TARGET_INCLUDE
#define HWY_TARGET_INCLUDE "/src/main.cc"
#include <hwy/foreach_target.h> // IWYU pragma: keep

// Must come after foreach_target.h to avoid redefinition errors.
#include <hwy/highway.h>
#include <hwy/print-inl.h>

HWY_BEFORE_NAMESPACE();
namespace hwy {
namespace HWY_NAMESPACE {

using namespace hwy::HWY_NAMESPACE;

void Test() {
  int32_t in1[4] = {0, 1, 2, 3};
  int32_t in2[4] = {4, 5, 6, 7};

  const ScalableTag<int32_t> di32;
  const ScalableTag<int16_t> di16;

  auto top = LoadU(di32, in1);
  auto bottom = LoadU(di32, in2);

  Print(di32, "top", top, 0, Lanes(di32));
  Print(di32, "bottom", bottom, 0, Lanes(di32));
  Print(di16, "sum", ReorderDemote2To(di16, top, bottom), 0, Lanes(di16));
}

void RunTest() {
  printf("------------------------ %s\n", TargetName(HWY_TARGET));
  Test();
}

// NOLINTNEXTLINE(google-readability-namespace-comments)
} // namespace HWY_NAMESPACE
} // namespace hwy
HWY_AFTER_NAMESPACE();

#if HWY_ONCE
namespace hwy {
HWY_EXPORT(RunTest);

void Run() {
  for (int64_t target : SupportedAndGeneratedTargets()) {
    SetSupportedTargetsForTest(target);
    HWY_DYNAMIC_DISPATCH(RunTest)();
  }
  SetSupportedTargetsForTest(0); // Reset the mask afterwards.
}

} // namespace hwy

int main(int /*argc*/, char ** /*argv*/) {
  hwy::Run();
  return 0;
}
#endif // HWY_ONCE
```
</details>

Raspberry Pi 3B - ARMv7
```bash
------------------------ NEON
i32x4 top [0+ ->]:
  0,1,2,3,
i32x4 bottom [0+ ->]:
  4,5,6,7,
i16x8 sum [0+ ->]:
  0x0004,0x0005,0x0006,0x0007,0x0000,0x0001,0x0002,0x0003,
```

Raspberry Pi 4B - AArch64
```bash
------------------------ NEON
i32x4 top [0+ ->]:
  0,1,2,3,
i32x4 bottom [0+ ->]:
  4,5,6,7,
i16x8 sum [0+ ->]:
  0x0000,0x0001,0x0002,0x0003,0x0004,0x0005,0x0006,0x0007,
```

### DIFF
--- a/hwy/ops/arm_neon-inl.h
+++ b/hwy/ops/arm_neon-inl.h
@@ -4666,14 +4666,14 @@ HWY_API Vec128<int16_t> ReorderDemote2To(Full128<int16_t> d16,
   return Vec128<int16_t>(vqmovn_high_s32(a16.raw, b.raw));
 #else
   const Vec64<int16_t> b16(vqmovn_s32(b.raw));
-  return Combine(d16, a16, b16);
+  return Combine(d16, b16, a16);
 #endif
 }
 
 HWY_API Vec64<int16_t> ReorderDemote2To(Full64<int16_t> /*d16*/,
                                         Vec64<int32_t> a, Vec64<int32_t> b) {
   const Full128<int32_t> d32;
-  const Vec128<int32_t> ab = Combine(d32, a, b);
+  const Vec128<int32_t> ab = Combine(d32, b, a);
   return Vec64<int16_t>(vqmovn_s32(ab.raw));
 }
 


kleisauke:
Perhaps RISC-V needs something similar?
https://github.com/google/highway/blob/9cbed31509c9f324b276f14b1a91f1a11528ef97/hwy/ops/rvv-inl.h#L3120

https://github.com/google/highway/blob/9cbed31509c9f324b276f14b1a91f1a11528ef97/hwy/ops/rvv-inl.h#L3131

https://github.com/google/highway/blob/9cbed31509c9f324b276f14b1a91f1a11528ef97/hwy/ops/rvv-inl.h#L3179
(not sure about the this one)

Unfortunately, I don't currently have the necessary hardware to test that target.

kleisauke:
@jan-wassenberg Ah, I see. Sorry for the misunderstanding. `DemoteTo` isn't an option for me, as it only accepts a single vector. I'm looking for an equivalent of `packs_epi32` that allows me to pack two i32's into a single i16 vector with signed saturation while maintaining order.

Assuming that retaining the order doesn't cause a performance penalty on any target, perhaps the i16-based version of `ReorderDemote2To` could be renamed to `Demote2To` where the later operation retains order? ABI/API compatibility would still be preserved, since this operation is not yet included in any released version of Highway.

jan-wassenberg:
No worries :) Please do not hesitate to report issues, we're happy to have a look.

hm, I understand it would be nice to use packs_epi32. But what happens with AVX-512 - shall we conform to its per-128-bit-block ordering? That could be doable, but expensive on RISC-V and SVE. The reason we require the freedom to reorder lanes is that SVE2 defines demotion as writing to the upper or lower half of each lane, rather than compacting them within a vector as SSE4 or AVX2 do.

I'm curious why it's essential to pack two input vectors at a time - is it not instead possible to write two demoted half-vectors (each the result of DemoteTo)?

If it's essential for performance and you only care about NEON, perhaps target-specific code would be a non-ideal but pragmatic solution?

kleisauke:
@jan-wassenberg Thanks for the explanation! AVX-512, RISC-V and SVE are actually the only targets I haven't tested yet. For SVE, I'll see if I can test it on an AWS Graviton3-based instance. For RISC-V, I was planning to test it with QEMU.

> I'm curious why it's essential to pack two input vectors at a time

I'm using `ReorderDemote2To` here:
https://github.com/kleisauke/libvips/blob/e17e420e666c8289da01c3c97c9306b84d6baa5d/libvips/resample/reduce_hwy.cpp#L155-L156

Which is just a 1:1 translation of the previous SSE4.1/AVX2 intrinsics I experimented with.
https://github.com/kleisauke/libvips/blob/9f9104dab252afa885ce34b2f28bdd8d1df4615d/libvips/convolution/convi_sse41.c#L226-L227

I also probably incorrectly assumed that the i16-based versions of `ReorderWidenMulAccumulate` preserves order on all targets.
https://github.com/kleisauke/libvips/blob/e17e420e666c8289da01c3c97c9306b84d6baa5d/libvips/resample/reduce_hwy.cpp#L143-L148

> is it not instead possible to write two demoted half-vectors (each the result of DemoteTo)?

Good idea! Implemented with commit https://github.com/kleisauke/libvips/commit/898b483b12e681d9b68c8732ec359efa9f0a65c2, but I'm not sure if this causes a performance penalty on x86. Perhaps the compiler optimizes it either way.

> If it's essential for performance and you only care about NEON, perhaps target-specific code would be a non-ideal but pragmatic solution?

I think for libvips an abstraction layer that can be ported to different architectures is preferable. We are currently using [liborc](https://gitlab.freedesktop.org/gstreamer/orc) for that, but are considering using Highway instead.

jan-wassenberg:
You're welcome :)

> For SVE, I'll see if I can test it on an AWS Graviton3-based instance.

Yes, that works well for us. You may be interested in build instructions (specifically for vqsort, but it depends on Highway): https://github.com/google/highway/tree/master/hwy/contrib/sort#aws-graviton3

> For RISC-V, I was planning to test it with QEMU.

Makes sense, that is also what we use.

> Which is just a 1:1 translation of the previous SSE4.1/AVX2 intrinsics I experimented with.

I understand. In this case, differences between platforms have caused us to shape the wrapper op a bit differently. Often it is possible to match x86 semantics on other platforms, but here it was too costly.

> I also probably incorrectly assumed that the i16-based versions of ReorderWidenMulAccumulate preserves order on all targets.

Oh, I thought you were only using the sum of all lanes? `ReorderWidenMulAccumulate` does indeed change order, but that's usually fine for dot products.

> Good idea! Implemented with commit https://github.com/kleisauke/libvips/commit/898b483b12e681d9b68c8732ec359efa9f0a65c2, but I'm not sure if this causes a performance penalty on x86.

Looks good! I do not expect much if any penalty, because the remainder of the loop isn't using port5 (shuffle) nor storing, so those two extra instructions should overlap with others.

> I think for libvips an abstraction layer that can be ported to different architectures is preferable. We are currently using [liborc](https://gitlab.freedesktop.org/gstreamer/orc) for that, but are considering using Highway instead.

Sure. Sorry to be unclear, I meant that it is possible when using Highway to write the generic version (2x DemoteTo), but also to have an `#if HWY_TARGET == HWY_NEON` specialization for one particular target of the few intrinsics that cannot be expressed performance-portably.

kleisauke:
Thanks!

> Oh, I thought you were only using the sum of all lanes? `ReorderWidenMulAccumulate` does indeed change order, but that's usually fine for dot products.

Still need to verify this, I haven't encountered a issue in the current implementation, at least not on SSE4, AVX2, NEON (ARMv7 and v8), Wasm and EMU128.

I tried testing RISC-V with QEMU today and ran into these two issues:
- The CMake build didn't install `hwy/ops/rvv-inl.h`, so I had to install that file manually to fix these compile errors: https://gist.github.com/kleisauke/9815fbbbb05f4b5662e1c2453e3f6a56
- Using `ZipLower` / `ZipUpper` seems to fail on compile time, see for example: https://gist.github.com/kleisauke/c23d2bdc42c9c3f537c1db6d4c0352eb

Let me know if you prefer to track this in a separate issue. I'll close this PR, as it has been resolved with the above mentioned commit instead.

jan-wassenberg:
Oops, that file is indeed missing from CMake. Adding, thanks :)

For ZipUpper, I believe we have a small misunderstanding. The D arg, as per usual convention, describes the return type. Zip is defined so as to return a double-width type, so the D cannot be u8 (because then the input would have had to be u4, which doesn't exist).
If you just want to interleave u8 lanes, that op is called InterleaveUpper/Lower.

Will update the documentation to try to make this more clear :)

kleisauke:
Ah, indeed. According to https://reviews.llvm.org/D129013#3707301, the Linux kernel is missing support for vector save/restore, so all vector instructions currently cause a `SIGILL`. QEMU's user space emulation, however, works properly.
```bash
$ qemu-riscv64 -cpu rv64,v=true,vlen=128,vext_spec=v1.0 -L /usr/riscv64-linux-gnu <EXECUTABLE>

# or:
# $ export QEMU_CPU=rv64,v=true,vlen=128,vext_spec=v1.0
# $ export QEMU_LD_PREFIX=/usr/riscv64-linux-gnu
# $ qemu-riscv64 <EXECUTABLE>
```

kleisauke:
> > Oh, I thought you were only using the sum of all lanes? `ReorderWidenMulAccumulate` does indeed change order, but that's usually fine for dot products.
> 
> Still need to verify this, I haven't encountered a issue in the current implementation, at least not on SSE4, AVX2, NEON (ARMv7 and v8), Wasm and EMU128.

I confirm that this is not a issue with our current implementation on the RISC-V target. However, I had to apply these patches:
https://gist.github.com/kleisauke/495919f31a9451d08bba0c009c53a43e (as noticed in https://github.com/google/highway/pull/1019#issuecomment-1280791861)
https://gist.github.com/kleisauke/9112516b80e8ef2b8bd8328bf16cb07d (rebased version of [Debian's patch](https://salsa.debian.org/debian-phototools-team/highway/-/blob/a953469c08ea7f4f5d1e27ca8ed5bb555d6c719e/debian/patches/riscv.patch))

(+ I needed to revert commit https://github.com/kleisauke/libvips/commit/c03a272c30fc8b95cdf7355e908b8cf0b74378ad locally)

Please let me know if you are interested in the first mentioned patch. I'll take a look at testing the (sizeless) SVE[2] targets next.

Note that libvips is considering adopting Highway in libvips 8.15, due in June 2023, which gives us some time to benchmark and test this properly. I'm not sure if we are still going to support [liborc](https://gitlab.freedesktop.org/gstreamer/orc) paths, but if we are going to, we probably prefer the Highway implementation, [if available](https://repology.org/project/highway-simd-library/versions). :)

> Sure. Sorry to be unclear, I meant that it is possible when using Highway to write the generic version (2x DemoteTo), but also to have an `#if HWY_TARGET == HWY_NEON` specialization for one particular target of the few intrinsics that cannot be expressed performance-portably.

Note that for this particular case, `#if HWY_TARGET == HWY_NEON && HWY_ARCH_ARM_V7` can be used, since this is not a issue on AArch64.

Due to this, I avoided using `ReorderDemote2To` altogether with commit https://github.com/kleisauke/libvips/commit/b072bcaf4f317f463f6f17cb13a256eec52827db, but hopefully I can revisit this in the future by doing that only for `#if (HWY_TARGET == HWY_NEON && HWY_ARCH_ARM_V7) || HWY_HAVE_SCALABLE`.

jan-wassenberg:
It's nice to see your progress, congrats and thanks for sharing!

I still think that reliance on `Reorder*` ops not reordering elements is a bug. If we guarantee an ordering on a particular platform, you could do the `#if`, but that takes away the possibility of us later improving performance if it requires another order. Do you agree with that reasoning? And if so, perhaps the Highway 'bug' fixed by your first patch is actually helpful, in that it reminds users not to rely on the order?

Your patch that instead does 2x DemoteTo+Store looks good. I'm curious whether you noticed a performance reduction from that?

Happy to help with SVE in case you run into issues.

kleisauke:
No problem! You're right that I should not rely on the order of the `Reorder*` ops, as mentioned in the docs and the name of the operation.

I agree with the reasoning that not preserving the order may be beneficial for future performance. For `ReorderDemote2To` this can indeed by avoided by doing 2x DemoteTo+Store (afaik, the perfomance degradation is negligible in that case), which _may_ be `#ifdef`'d only for RVV/SVE and ARMv7. But how about `ReorderWidenMulAccumulate`? There's no `MulAddAdjacent` operation available that can be used as an alternative.

For reference, without having tested the SVE[2] targets yet, the reordering of the `Reorder*` operations can be circumvented with these patches:
https://gist.github.com/kleisauke/495919f31a9451d08bba0c009c53a43e (RVV - `ReorderWidenMulAccumulate`)
https://gist.github.com/kleisauke/9a85c9f1dc5aae037a8b80c83925d2f7 (RVV - `ReorderDemote2To`)
https://github.com/google/highway/pull/1019.patch (ARMv7 - `ReorderDemote2To`, i.e. this abandoned PR)

Since this _may_ only be a "issue" on the SVE[2] targets, how about providing inline non-reorder overloads on targets where this reordering does not happen? For example, the specific SSE2-`Reorder*` operations for can be implemented like this:
```cpp
// madd_epi16 maintains order
template <size_t N>
HWY_INLINE Vec128<int32_t, N> WidenMulAccumulate(
        Simd<int32_t, N, 0> d32, Vec128<int16_t, 2 * N> a,
        Vec128<int16_t, 2 * N> b, const Vec128<int32_t, N> sum0,
        Vec128<int32_t, N>& sum1) {
  return ReorderWidenMulAccumulate(d32, a, b, sum0, sum1);
}

// packs_epi32 maintains order
HWY_INLINE Vec128<int16_t> Demote2To(Full128<int16_t> d16,
                                     Vec128<int32_t> a, Vec128<int32_t> b) {
  return ReorderDemote2To(d16, a, b);
}
```
(it can also be implemented vice-versa, e.g. `ReorderDemote2To` calls `Demote2To`)

Or would that be a maintenance burden for future SIMD targets? I take the possible perfomance pentalty for granted on SVE[2] and other further targets, since that's always better than having to fall-back on scalar code. :)

jan-wassenberg:
Glad to hear 2x DemoteTo is acceptable for speed.

hm, I'm thinking about the implications of providing a non-reordering WidenMulAccumulate. Would the semantics be that adjacent pairs are added together like madd_epi16?
SVE2 svmlalt_s32 takes the odd or even lanes and multiplies. If we are forced to add the results of both together, we are reducing ILP and doubling latency.
For NEON it seems considerably worse - it can mul together the low or high halves of two regs, so we'd have to uzp1+2 to get the odds/evens before doing the two muls, right?

Highway does aim to avoid such performance cliffs, where non-reordering WidenMul could be a good bit worse than what native code would have done.

Is there any way to avoid the need for a known order of WidenMul output in your app code?

kleisauke:
> Would the semantics be that adjacent pairs are added together like madd_epi16?

No, I think it's fine to use two accumulators for non-x86 and Wasm targets. My current implementation works for that, so there's no need to change the function signature of `ReorderWidenMulAccumulate` when implementing a non-reorder variant (i.e. `WidenMulAccumulate`), IIUC.

> For NEON it seems considerably worse - it can mul together the low or high halves of two regs, so we'd have to uzp1+2 to get the odds/evens before doing the two muls, right?

Ah, you mean something like this?:
https://github.com/simd-everywhere/simde/blob/dd0b662fd8cf4b1617dbbb4d08aa053e512b08e4/simde/x86/sse2.h#L3696-L3705

Indeed, that's probably considerably worse, in terms of performance.

> Is there any way to avoid the need for a known order of WidenMul output in your app code?

I need to look into this, the convolution-based resampling paths probably does not have this requirement, at least not when I initially implemented the scalar implementation, see for example:
https://github.com/kleisauke/libvips/blob/2859d4722614397396acb89619de7f2bcfe95ca1/libvips/resample/reduce_hwy.cpp

After that, it was implemented this way:
https://github.com/kleisauke/libvips/blob/eefe35dcb72865093010b7cf111da069c36c5215/libvips/resample/reduce_hwy.cpp

Though, I'm not sure if the same applies to the [`vips_convi()` highway path](https://github.com/kleisauke/libvips/blob/eefe35dcb72865093010b7cf111da069c36c5215/libvips/convolution/convi_hwy.cpp).

jan-wassenberg:
Thanks for the link. Looks like I was overly pessimistic and we don't actually require uzp1. Only the latency issue remains, and that can go away if we allow two accumulators as you mention.

I just noticed ReorderWidenMulAccumulate is not documented. This is an interesting opportunity to perhaps guarantee more of an order :)
Let's recap what each arch can do:
1) x86: already adds pairs into one accumulator; the second accumulator does nothing
2) NEON, RVV: first acc does lower half of the whole register, second acc does upper half
3) SVE2: first acc does even lanes, second acc does odd lanes

For x86 and SVE2 we could say that sum0+sum1 has the sum of odd and even lanes. For NEON that would require vpaddq_s32 as SIMDe is doing (pairwise add), which is decent for A64 but expensive to emulate for v7. And on RVV it's even more expensive to do pairwise add.

The bare minimum guarantee is that the sum of all lanes of sum0+sum1 is the desired sum. Slightly better is that (sum0+sum1)[i] is either `product[2*i+1]+product[2*i]`, or `product[2*i+N]+product[2*i]`. Would that help?

kleisauke:
(I apologize for the following questions, SIMD in general is a relatively new area for me)

> Only the latency issue remains, and that can go away if we allow two accumulators as you mention.

Benchmarking for non-x86 targets is still on my TODO list. Although x86 uses one accumulator, we noticed a nice speed improvement on x86 with this implementation, see e.g.:
https://github.com/kleisauke/vips-microbench/blob/simd/results/simd.md
https://github.com/kleisauke/vips-microbench/blob/simd/results/simd-convi.md

> I just noticed ReorderWidenMulAccumulate is not documented.

It looks like it's documented in `g3doc/quick_reference.md`: https://github.com/google/highway/blob/master/g3doc/quick_reference.md#multiply

> This is an interesting opportunity to perhaps guarantee more of an order :)

Thanks for considering this. :)

> For x86 and SVE2 we could say that sum0+sum1 has the sum of odd and even lanes.

I assume this does not require changes to the existing `ReorderWidenMulAccumulate` function for x86?

> For NEON that would require vpaddq_s32 as SIMDe is doing (pairwise add), which is decent for A64

Does this mean that the semantics for AArch64 is then same as x86 (i.e. adjacent pairs are added together, second accumulator is no-op)? 

> but expensive to emulate for v7. And on RVV it's even more expensive to do pairwise add.

FWIW, you may have noticed that I was doing this (before the store):
```c
#if !(HWY_ARCH_X86 || HWY_ARCH_WASM) || HWY_TARGET == HWY_EMU128
    /* De-interleave all accumulators by pairs.
     */
    sum0 = ConcatEven(di32, /* hi */ sum1, /* lo */ sum0);
#endif
```
(IIUC, that would end-up calling `vuzp1_s32` on AArch64/ARMv7)

I'm fine with modifying that to:
```diff
@@ -1,4 +1,4 @@
-#if !(HWY_ARCH_X86 || HWY_ARCH_WASM) || HWY_TARGET == HWY_EMU128
+#if (HWY_TARGET == HWY_NEON && HWY_ARCH_ARM_V7) || HWY_TARGET == HWY_RVV
     /* De-interleave all accumulators by pairs.
      */
     sum0 = ConcatEven(di32, /* hi */ sum1, /* lo */ sum0);
```

And still using `ReorderWidenMulAccumulate` on ARMv7 and RVV. Note that I'm currently testing ARMv7 on a Raspberry Pi 3B with Windows 10 IoT and RVV with QEMU's user space emulation, so it's probably only useful to test its correctness (I don't think I could run any decent benchmarks on it).

> The bare minimum guarantee is that the sum of all lanes of sum0+sum1 is the desired sum. Slightly better is that (sum0+sum1)[i] is either `product[2*i+1]+product[2*i]`, or `product[2*i+N]+product[2*i]`. Would that help?

Would that correspond to the current description of `ReorderWidenMulAccumulate` in the docs?:
> [..] The only guarantee is that `SumOfLanes(d, Add(return_value, sum1))` is the sum of all `a[i] * b[i]`. 

If so, that would probably still work for us, as long as `ConcatEven(di32, /* hi */ sum1, /* lo */ return_value)` produces the same output on all supported targets that use two accumulators (i.e. no reordering) or `return_value` has its adjacent pairs added together when one accumulator is used (i.e. the current x86 and Wasm semantics).

jan-wassenberg:
> It looks like it's documented in g3doc/quick_reference.md:

Thank you for pointing that out. It was not showing up in my search because of a formatting issue I have just fixed (#1070).

> I assume this does not require changes to the existing ReorderWidenMulAccumulate function for x86?

That's right.

> Does this mean that the semantics for AArch64 is then same as x86 (i.e. adjacent pairs are added together, second accumulator is no-op)?

Yes, it would be - though the paddq would already be more expensive than we'd like.

> Would that correspond to the current description of ReorderWidenMulAccumulate in the docs?:

Yes, what I said is the bare minimum, is the current documented behavior.

I just had a new idea based on this discussion :) You call ReorderWidenMulAccumulate in the loop, and then fixup outside the loop.
My concern was about pulling paddq into ReorderWidenMulAccumulate, which would be expensive.
However, we could introduce a new op to do the fixup (either paddq, or ConcatEven - the latter is more expensive for RVV).

The op would already know what ReorderWidenMulAccumulate did, so it can establish your preferred order.
(BTW, I think ConcatEven of sum1 and sum0 is not 100% safe, or guaranteed by the current documentation.)
How about we guarantee the return value is `product[2*i+1]+product[2*i]`?
Any suggestions on what to call the new op? `RearrangeToOddPlusEven`?

kleisauke:
> You call ReorderWidenMulAccumulate in the loop, and then fixup outside the loop.
> My concern was about pulling paddq into ReorderWidenMulAccumulate, which would be expensive.

For AArch64, I wasn't sure how expensive calling `vpaddq_s32` inside the nested loop is, but indeed, it's probably more expensive than calling `vuzp1_s32` once outside that loop.

> I just had a new idea based on this discussion [...] we could introduce a new op to do the fixup (either paddq, or ConcatEven - the latter is more expensive for RVV).
> The op would already know what ReorderWidenMulAccumulate did, so it can establish your preferred order.

That would be great! For RVV, perhaps something along this could be used (as alternative to `ConcatEven`)?
https://github.com/google/highway/blob/f309fd8ef3a1eed3ed6f471e3f680248393c759a/hwy/contrib/bit_pack/bit_pack-inl.h#L1402-L1403
(though, I'm not sure if that only applies to 8-bit lanes)

> (BTW, I think ConcatEven of sum1 and sum0 is not 100% safe, or guaranteed by the current documentation.)

Indeed, since it's not guaranteed by the docs, it was not future-proof either. It was found during trial-and-error and _may_ only be safe for the targets that uses two accumulators. At least on NEON (ARMv7 and v8), RVV and EMU128 it did not cause any issues. However, as mentioned above, RVV required [this patch](https://gist.github.com/kleisauke/495919f31a9451d08bba0c009c53a43e), and I did not test the SVE[2] targets (yet).

For now, I only support targets with fixed-size vectors due to this (commit https://github.com/kleisauke/libvips/commit/c03a272c30fc8b95cdf7355e908b8cf0b74378ad, this is presumably reasonable, e.g. libjxl also does not support SVE/RVV by default unless you build with `-DJPEGXL_ENABLE_SIZELESS_VECTORS=true`).

> How about we guarantee the return value is `product[2*i+1]+product[2*i]`?

IIUC, within my current implementation, on AArch64, this would call `vpaddq_s32` and thus this new op should be placed inside the nested loop in favor of calling `vuzp1_s32` once outside that loop? Let me properly benchmark that first.

I assume this new op is also implemented for x86 and Wasm? If so, it could just return `sum0` in that case (though, that feels a bit odd).

Anyhow, I'll properly test the SVE[2] targets next week, perhaps that `ConcatEven` trick will work there too(?).

> Any suggestions on what to call the new op? `RearrangeToOddPlusEven`?

`RearrangeToOddPlusEven` sounds good to me. Or perhaps just `AddPairwise` (allowing it to be used outside the `ReorderWidenMulAccumulate` context)?

jan-wassenberg:
> For RVV, perhaps something along this could be used (as alternative to ConcatEven)

hm, this trick replaces ConcatEven, but only when a different ordering would be acceptable. It would interleave sum0 and sum1, which in this context means `product[2*i]` followed by `product[2*i+N]`, which is not what we want here. I think on RVV we will have to do a pairwise add, emulated either using the shift+AND trick you mention, or slide1up/down, followed by a DemoteTo.

> For now, I only support targets with fixed-size vectors due to this

Then let us solve the problem :)

> IIUC, within my current implementation, on AArch64, this would call vpaddq_s32 and thus this new op should be placed inside the nested loop in favor of calling vuzp1_s32 once outside that loop?

Clarification: I propose that you keep your code unchanged, except that your #if .. ConcatEven outside the loops is replaced with RearrangeToOddPlusEven.
That new op will (like all others) be implemented for all platforms, and will do the right thing on each, but it will only make sense to use it together with ReorderWidenMulAccumulate, because its behavior is linked to what ReorderWidenMulAccumulate did. (On x86 it will be a no-op.)

I will implement this next week.

jan-wassenberg:
@kleisauke done :) Please let me know whether this works for you.

kleisauke:
You made my day, thank you! :)

Implemented with commit https://github.com/kleisauke/libvips/commit/228c0b3d660bd32ace5663bc33892d123de5697f and https://github.com/kleisauke/libvips/commit/de418bf65438382d08f872e11c1d356ecfecf402, it seems to work on at least:
- x86 (as expected, it's no-op there);
- Neon (both AArch64 and Armv7);
- Wasm (only tested baseline, still need to test EMU256).

For RVV, it produces a different image when using `vlen=256` instead of `vlen=128`. 128-bit vectors fortunately still match its C output.
```console
$ sha256sum x-c.png x-rvv-128.png x-rvv-256.png
8a41a18698d1f03496538caa6262c862185473e2d84d286150cf8ef2555a2912  x-c.png
8a41a18698d1f03496538caa6262c862185473e2d84d286150cf8ef2555a2912  x-rvv-128.png
58a1ca9b213fa3b8090264b741d76e35498d750184b12f4a99cc77d387fd7cc9  x-rvv-256.png
$ vips avg x-c.png
96.826229
$ vips avg x-rvv-256.png
96.845063
```
However, this might be a bug in QEMU's user space emulation (tested on v7.0.0, as available on Fedora 37).

Note that Wasm is always doing `Add(sum0, sum1)`, which should not be necessary for the i16 variant (since that has the same semantics as x86). It could cause issues when `sum1` is a non-zero vector (see e.g. the CI result of commit https://github.com/kleisauke/wasm-vips/commit/b276ad41995238195def7a3ce236ae5e663c05cf and the subsequent commit https://github.com/kleisauke/wasm-vips/commit/ad323399ba357c6abef14f545345c9ce20f63d6a which fixed that). This was probably not noticed by the `RearrangeToOddPlusEven()` test, since that uses a zero vector for `sum1`:
https://github.com/google/highway/blob/6fd9ebd6693da37bc6797ca43db72a5c4156e95f/hwy/tests/mul_test.cc#L465

FWIW, while doing this I also noticed test failures on EMU128 (using `export VIPS_SIMD=-2305843009213693953`, i.e. `~HWY_EMU128`), but that seems to also fail prior to these commits, so it needs some investigation on my side.

> https://gist.github.com/kleisauke/9a85c9f1dc5aae037a8b80c83925d2f7 (RVV - `ReorderDemote2To`)
> https://github.com/google/highway/pull/1019.patch (ARMv7 - `ReorderDemote2To`, i.e. this abandoned PR)

While not necessarily a bug fix, please let me know if there is interest in these patches. I could polish the commit messages and open a new PR, after I tested the SVE[2] targets (to verify whether the reordering takes place there).

jan-wassenberg:
:) Glad to hear, thanks for confirming the results.
Good point about wasm, we'll fix that so you can hopefully avoid those `#if` for initializing.

Do let me know if you find any issue with EMU128 or RVV.

> While not necessarily a bug fix, please let me know if there is interest in these patches.

Yes please, let's get that in now that RearrangeToOddPlusEven is guaranteeing the order.

kleisauke:
Great! Simplified that initialization with commit https://github.com/kleisauke/libvips/commit/2b63cd16a630a534309bf9bb4ce017783b375593 which passes the wasm-vips CI when tested in combination with commit 712312b0324555850b9dcf9b890880aa554230e3.

I could resolve the RVV issue with commit https://github.com/kleisauke/libvips/commit/c0a2a207be81a882625f02bbe87915eb43c18f14, although I'm not sure why that does not affect `vlen=128`. This was tested in combination with the above `ReorderDemote2To` patch. I confirm that it's not reproducible on Highway's test suite, at least not on this script:
<details>
  <summary>Details</summary>

```bash
export QEMU_CPU=rv64,v=true,vlen=256,vext_spec=v1.0
export QEMU_LD_PREFIX=/usr/riscv64-linux-gnu

export CC=clang-15
export CXX=clang++-15

export CFLAGS="-march=rv64gcv1p0 -menable-experimental-extensions"
export CXXFLAGS="$CFLAGS"
export LDFLAGS="-fuse-ld=lld"

# Cross-build (assumes RHEL/Fedora)
export CFLAGS+=" --target=riscv64-redhat-linux -I/usr/riscv-linux-gnu/include --sysroot=$PWD/gcc_toolchain"
export CXXFLAGS="$CFLAGS"

# Assume that vector registers are at least 256-bit wide
# Note: this has also been tested without these flags
export CFLAGS+=" -mllvm --riscv-v-vector-bits-min=256"
export CXXFLAGS="$CFLAGS"
export LDFLAGS+=" -mllvm --riscv-v-vector-bits-min=256"

cd highway/
rm -rf build_riscv64
mkdir build_riscv64
cmake -G Ninja -S . -B build_riscv64 \
  -DCMAKE_BUILD_TYPE=Release \
  -DHWY_WARNINGS_ARE_ERRORS:BOOL=ON
cmake --build build_riscv64 -- -j$(nproc)
ctest --test-dir build_riscv64
```
</details>

I haven't yet been able to figure out what is causing those test failures on `EMU128`. It looks like some sort of precision error (perhaps incorrect codegen?). It's also not reproducible on Highway's test suite, using this test script:
<details>
  <summary>Details</summary>

```bash
export CC=clang-15
export CXX=clang++-15

export CFLAGS="-DHWY_COMPILE_ONLY_EMU128"
export CXXFLAGS="$CFLAGS"

rm -rf build_emu128
mkdir build_emu128
cmake -G Ninja -S . -B build_emu128 \
  -DCMAKE_BUILD_TYPE=Release \
  -DHWY_WARNINGS_ARE_ERRORS:BOOL=ON
cmake --build build_emu128 -- -j$(nproc)
ctest --test-dir build_emu128
```
</details>

(I could not test it with GCC since Fedora 37 provides version 12.2.1, so it will end up defining `HWY_BROKEN_EMU128` as `1`)

I'll look into writing a smoke test for this using randomly generated vectors, [similar to the tests in SIMDe](https://github.com/simd-everywhere/simde/blob/dd0b662fd8cf4b1617dbbb4d08aa053e512b08e4/test/x86/avx512/madd.c#L95-L109).

jan-wassenberg:
Congrats, it's nice to avoid the #if :)

Incorrect codegen does unfortunately happen with some regularity.
Thanks, great idea to add a more extensive test with random vectors, will be happy to review and integrate.

kleisauke:
Fortunately, it was not an incorrect codegen, as I was able to reproduce it with GCC 12.2.1 when compiling with `-DHWY_BROKEN_EMU128=0`. I just opened issue #1081 for this.

PR #1036, opened 2022-11-03
Make sure to link targets_test against GCC/atomic lib

malaterre:
This commit will fix the following compilation error on armel:

/usr/bin/ld: CMakeFiles/targets_test.dir/hwy/targets_test.cc.o: undefined reference to symbol '__atomic_store_8@@LIBATOMIC_1.0' /usr/bin/ld: /usr/lib/arm-linux-gnueabi/libatomic.so.1: error adding symbols: DSO missing from command line

Some minor CMake comments shuffling to make sections symetrics.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -294,8 +294,8 @@ set_target_properties(hwy PROPERTIES
   LINK_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/hwy/hwy.version)
 # For GCC __atomic_store_8, see #887
 target_link_libraries(hwy PRIVATE ${ATOMICS_LIBRARIES})
+# not supported by MSVC/Clang, safe to skip (we use DLLEXPORT annotations)
 if(UNIX AND NOT APPLE)
-  # not supported by MSVC/Clang, safe to skip (we use DLLEXPORT annotations)
   set_property(TARGET hwy APPEND_STRING PROPERTY
     LINK_FLAGS " -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/hwy/hwy.version")
 endif()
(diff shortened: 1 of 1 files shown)


malaterre:
Hopefully the last tweak for GCC/atomic support ...

ref:
* https://buildd.debian.org/status/fetch.php?pkg=highway&arch=armel&ver=1.0.2-1&stamp=1667405898&raw=0

jan-wassenberg (approved):
Ah, sorry I didn't see this one was missing. Thanks for adding it!

malaterre:
> Ah, sorry I didn't see this one was missing. Thanks for adding it!

I believe this is a side effect of me adding the cmake keyword `PRIVATE` in my previous pull request.

malaterre:
@jan-wassenberg please hold

jan-wassenberg:
Sure. BTW while we're at it, would it be a good idea to add HWY_ prefix to ATOMICS_LIBRARIES, just in case some other project also defines that and Highway is used as a subdirectory?

malaterre:
@jan-wassenberg buildds are ok today:

* https://buildd.debian.org/status/package.php?p=highway

please merge, thanks !

malaterre:
> good idea to add HWY_ prefix to ATOMICS_LIBRARIES, just in case some other project also defines that and Highway is used as a subdirectory?

I've never seen that done before for JPEG, PNG or others. Could you describe further your suggestion ?

jan-wassenberg:
Ah, sure. I am merely proposing to rename `ATOMICS_LIBRARIES` to `HWY_ATOMICS_LIBRARIES` in CMakeLists.txt. Adding such a prefix helps avoid naming collisions.

malaterre:
@jan-wassenberg sorry to insist, but in cmake variable are in function or directory scope:

* https://cmake.org/cmake/help/latest/command/set.html

ATOMICS_LIBRARIES cannot be propagated to parent directories.

jan-wassenberg:
I did not know that, thank you for clarifying and keeping the code uncluttered :) Merging now.

jan-wassenberg (approved):

PR #1042, opened 2022-11-06
Don't include asm/hwcap.h

Hello71:
sys/auxv.h already provides hwcap macros; on glibc, it includes asm/hwcap.h, and on musl, it includes bits/hwcap.h. manually adding asm/hwcap.h increases unnecessary dependencies and chance of conflicting definitions

### DIFF
--- a/hwy/targets.cc
+++ b/hwy/targets.cc
@@ -43,7 +43,6 @@
 #endif  // HWY_COMPILER_MSVC
 
 #elif HWY_ARCH_ARM && HWY_OS_LINUX
-#include <asm/hwcap.h>
 #include <sys/auxv.h>
 #endif  // HWY_ARCH_*
 


hwy/targets.cc
```diff
@@ -28,7 +28,7 @@
 
 #include "hwy/per_target.h"  // VectorBytes
 
-#if HWY_IS_ASAN || HWY_IS_MSAN || HWY_IS_TSAN
+#if HWY_IS_ASAN || HWY_IS_MSAN || HW62e29717ac4e15a66f0f8dd3db2d350fc1de5a68Y_IS_TSAN
```
jan-wassenberg, line 31: Looks like something got pasted here, please remove :)

jan-wassenberg:
Thanks, makes sense.

jan-wassenberg (approved):

PR #1071, opened 2022-11-25
Check for the presence of <sys/auxv.h>

jolivain:
Not all gcc versions are providing <sys/auxv.h>. Checking for HWY_ARCH_ARM && HWY_COMPILER_GCC_ACTUAL && HWY_OS_LINUX is not sufficient and fail to build in some situations (it was observed for some gcc armv7m toolchains).

This patch adds a check for <sys/auxv.h> and include it only if present.

Signed-off-by: Julien Olivain <ju.o@free.fr>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -258,6 +258,12 @@ else()
 
 endif()  # !MSVC
 
+include(CheckIncludeFile)
+check_include_file(sys/auxv.h  HAVE_SYS_AUXV_H)
+if (NOT HAVE_SYS_AUXV_H)
+  list(APPEND HWY_FLAGS -DTOOLCHAIN_MISS_SYS_AUXV_H=1)
+endif()
+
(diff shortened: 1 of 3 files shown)


jan-wassenberg:
Thank you @jolivain for this change :) We also support other build systems, and it seems like the non-presence of this header would be the exceptional case. Would you mind reversing the logic, excluding the include only if it is known (by CMake) not to be present?

jolivain:
Hi @jan-wassenberg,
I just updated the patch to include your suggestion.
Thanks for the review!

jan-wassenberg (approved):
Thank you @jolivain  for making the change, looks good. Optional, do you think we should add a HWY_ prefix to `TOOLCHAIN_MISS_SYS_AUXV_H` just in case some other project happens to also define that on the command line, which could cause conflicts?

jan-wassenberg:
(FYI our internal CI is currently stuck.)

PR #1101, opened 2023-01-02
Ensure `ReorderDemote2To` preserves order on Armv7/RVV

kleisauke:
This is a revisited patch of PR https://github.com/google/highway/pull/1019, now that I have confirmed that the [`simd-highway`](https://github.com/kleisauke/libvips/tree/simd-highway) branch of libvips also works on SVE (tested using [this script](https://gist.github.com/kleisauke/e085c4f3a546c8fe66604bdf484ff3dd)). This PR allows me to drop commit https://github.com/kleisauke/libvips/commit/0ca6a161a01f4c5129b90fcf1197887e05f8616f. 

Note that commit 36671ddd6c27b1619a2d9d81faefbbd00b6eb947 does not always ensure that the order is preserved on RVV, it could still cause a reorder with `vlen=256` (as noticed in https://github.com/google/highway/pull/1019#issuecomment-1336407603). Let me know if you want me to omit this commit from this PR (and retitle it accordingly).

As a possible follow-up, I could document that `ReorderDemote2To` does _not_ preverse the order only when `HWY_HAVE_SCALABLE == 1` (see e.g. commit https://github.com/kleisauke/libvips/commit/fa2355f578fce92586c6e1e1f27a337d84af6ed9).

### DIFF
--- a/hwy/ops/arm_neon-inl.h
+++ b/hwy/ops/arm_neon-inl.h
@@ -4731,14 +4731,14 @@ HWY_API Vec128<int16_t> ReorderDemote2To(Full128<int16_t> d16,
   return Vec128<int16_t>(vqmovn_high_s32(a16.raw, b.raw));
 #else
   const Vec64<int16_t> b16(vqmovn_s32(b.raw));
-  return Combine(d16, a16, b16);
+  return Combine(d16, b16, a16);
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


jan-wassenberg:
Sorry to have missed this one. Feel free to ping/mention when it's been more than two days :)
Glad to hear it works, and thanks for sending the patch!

Let's keep the RVV change in for simplicity.

jan-wassenberg (approved):

Issue #1109, opened 2023-01-12
Add the "Shift Inwards" option to "Strip-mining loops"


emilmelnikov:
Currently, there are 4 options listed in the [_Strip-mining loops_](https://github.com/google/highway#strip-mining-loops) README section for handling remaining elements if the total number of elements is not divisible by the current number of lanes:

1. Do nothing (requires padded inputs)
2. Use `Transform*` functions that select between options 3 and 4 automatically
3. Scalar loop (always works, but may be slow)
4. Masked load/store (cannot be used if `HWY_MEM_OPS_MIGHT_FAULT`)

There is another option: process an additional whole vector for elements from `count - N` to `count`, that is, shift the loop tail inwards and potentially repeat computation for some elements:
```c++
for (size_t i = 0; i < count; i += N) LoopBody<false>(d, HWY_MIN(i, count - N), 0);
```

In Halide, there is a similar [`ShiftInwards` tail strategy](https://halide-lang.org/docs/namespace_halide.html#a6c6557df562bd7850664e70fdb8fea0f) option.

Pros: simple, almost as fast as option 1, doesn't require masking.

Cons: `count` cannot be smaller than `N`, computation needs to be pure.

jan-wassenberg:
Hi @emilmelnikov, that makes a lot of sense. I thought we had also documented that one, but it seems not[1]. Would you like to send a pull request with some text? (BTW I think we can mention 'idempotent' as the requirement.)

[1] Ah, I found what I remembered - it was some [comments on HN](https://news.ycombinator.com/item?id=32847924).

PR #1110, opened 2023-01-12
Describe the "Shift Inwards" stip-mining strategy

emilmelnikov:
Closes https://github.com/google/highway/issues/1109


emilmelnikov:
This strategy is described as a "second preferred option", but I am not sure about that.

I also considered to modify `Transform*` helper functions to support this case, but, similarly to the first option, it seems simpler to write a loop directly.

jan-wassenberg (approved):
Nice, thanks for writing/adding this!
FYI contrib/dot/dot-inl.h also includes similar logic.

PR #1113, opened 2023-01-17
Remove explicit setting of noexecstack

malaterre:
This setting is set on a particular glibc shipped in current Debian release. This is intentionally done by glibc when targeting minimum kernel 4.8.0 or older with mips hardfloat:

https://github.com/bminor/glibc/blob/595c22ecd8e87a27fd19270ed30fdbae9ad25426/sysdeps/unix/sysv/linux/mips/configure.ac#L138-L143

This backwards compatibility should not be more important than the security hardening of a non-executable stack.

Full reference:
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1025436

This reverts commit 97e4b7b74b4ef4f3765e5b09572ce8c583f33c07.

Co-authored-by: Jessica Clarke <jrtc27@debian.org>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -306,21 +306,6 @@ if(UNIX AND NOT APPLE)
     LINK_FLAGS " -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/hwy/hwy.version")
 endif()
 
-if (CMAKE_SYSTEM_PROCESSOR MATCHES "unknown")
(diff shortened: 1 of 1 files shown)


jan-wassenberg:
@boxerab does this make sense to you, would you be ok with removing the noexecstack?

boxerab:
@jan-wassenberg thanks, that's fine with me. I can always add it back into my cmake files, if needed.

jan-wassenberg (approved):

PR #1116, opened 2023-01-17
Generate and install a HWYConfig.cmake file

malaterre:
This will be needed by cmake user consuming HWY.

Also remove FindHWY.cmake since not needed in a cmake-based build tool setup.

Fixes 1114


jan-wassenberg (approved):

jan-wassenberg (approved):

CMakeLists.txt
```diff
@@ -571,3 +571,9 @@ endforeach ()
 target_sources(skeleton_test PRIVATE hwy/examples/skeleton.cc)
 
 endif()  # BUILD_TESTING
+
+# write HWYConfig file:
+include(CMakePackageConfigHelpers)
+write_basic_package_version_file("${CMAKE_CURRENT_BINARY_DIR}/HWYConfigVersion.cmake" COMPATIBILITY SameMajorVersion)
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/HWYConfigVersion.cmake" DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/hwy")
+install(EXPORT hwy_targets FILE HWYConfig.cmake DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/hwy")
```
dg0yt, line 579: The export should add a namespace (eg. `HWY::`). This lets downstream CMake recognize targets in link library properties and provide proper diagnostics at CMake generation time. 
(Plain literals like `hwy` could be namespecs to be passed to the linker. When the user fails to properly do `find_package(HWY)`, the lack of `hwy` can only be noticed at compile time or link time.)

dg0yt:
No patch. The change is to insert `NAMESPACE HWY::` into this line. Cf.
https://cmake.org/cmake/help/v3.20/command/install.html#export
https://cmake.org/cmake/help/v3.20/manual/cmake-packages.7.html#creating-packages

No reproducer. Expected UX app. like this.
No namespace:
~~~cmake
# CMakeLists.txt
# forgotten or not found: find_package(HWY CONFIG)
target_link_libaries(main PRIVATE hwy)
~~~
~~~
cmake ..
  (success)
make
  (wait a while)
... compiler error: Can't find include
... or linker error: Can't find a lib for spec `hwy`
~~~

With namespace:
~~~cmake
# CMakeLists.txt
# forgotten or not found: find_package(HWY CONFIG)
target_link_libaries(main PRIVATE HWY::hwy)
~~~
~~~
cmake ..
... cmake error: 
  HWY::hwy names a target but it wasn't defined. 
  Did you forget ... `find_package(...)?
~~~

PR #1126, opened 2023-01-26
Change cmake COMPATIBILITY for major version

malaterre:
We want to make sure that `find_package(HWY 0.15.0)` actually resolve when using any of hwy 1.0.x versions.

Documentation:
* https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -574,6 +574,6 @@ endif()  # BUILD_TESTING
 
 # write HWYConfig file:
 include(CMakePackageConfigHelpers)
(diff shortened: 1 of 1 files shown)


eustas:
While it is nice for transition 0 -> 1, it might be unexpected when we would have 1 -> 2 transition.
@jan-wassenberg WDYT?

eustas:
Awesome. Then it is worth adding an issue, so that won't be forgotten =)

jan-wassenberg:
Thanks @malaterre , it's great that you have patched jxl to use hwy >=1.0.0.

Issue #1123, opened 2023-01-21
BitCast for HWY_SCALAR target copies too many bytes if sizeof(T) < sizeof(FromT)


johnplatts:
Here is the current implementation of BitCast for HWY_SCALAR target:
https://github.com/google/highway/blob/9ad490b6dc5b3252c525b30e430577a92830653f/hwy/ops/scalar-inl.h#L92-L98

The CopyBytes operation in the current implementation of BitCast for the HWY_SCALAR target has undefined behavior if sizeof(T) < sizeof(FromT) is true.

Here is the corrected implementation of BitCast for HWY_SCALAR target:
```
 template <typename T, typename FromT> 
 HWY_API Vec1<T> BitCast(Sisd<T> /* tag */, Vec1<FromT> v) { 
   static_assert(sizeof(T) <= sizeof(FromT), "Promoting is undefined"); 
   T to; 
   CopyBytes<sizeof(T)>(&v.raw, &to);  // not same size - ok to shrink 
   return Vec1<T>(to); 
 }
```

jan-wassenberg:
Thanks @johnplatts , I agree with your fix. Would you mind sending a pull request for this change?

jan-wassenberg:
This is merged, closing. Thanks again :)

PR #1127, opened 2023-01-27
Fixed HWY_SCALAR BitCast to copy only sizeof(T) bytes

johnplatts:
Corrected the BitCast function in hwy/ops/scalar-inl.h to copy only sizeof(T) bytes.

The fix in this pull request resolves issue https://github.com/google/highway/issues/1123.


jan-wassenberg:
Thank you @johnplatts :)

jan-wassenberg (approved):

jan-wassenberg:
(FYI I am traveling without access to the internal CI tooling, so we'll merge this when I am back on Wed.)

PR #1129, opened 2023-01-30
Add support for find_package + Config mode

malaterre:
Also prepend target names with hwy:: namespace for better CMake user experience.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,8 @@ if(POLICY CMP0111)
 endif()
 
 project(hwy VERSION 1.0.3)  # Keep in sync with highway.h version
+# `hwy` is lowercase to handle find_package() in Config mode:
+set(namespace "${PROJECT_NAME}::")
 
 # Directly define the ABI version from the cmake project() version values:
 set(LIBRARY_VERSION "${hwy_VERSION}")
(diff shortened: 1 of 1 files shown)


dg0yt:
... actually I came here from looking at libjxl (in vcpkg).

IIUC you are bound by the existing information in `FindHMY.cmake`. I didn't consider this case.
If you want to keep compatibility but still migrate to the namespace, you can continue to provide the target under the old name as *deprecated* form along with the namespaced form. But admittedly this takes more code.

IMO explicitly dealing with `target_include_directories` is rarely needed because include directories are passed as transitive usage requirement when linking to a target as `PUBLIC`. 
There are also generator expression to test for the presence of a target so a migration path can be included where it is used.

jan-wassenberg:
This looks reasonable to me but we do not want to break libjxl, which [either does](https://github.com/libjxl/libjxl/blob/main/third_party/CMakeLists.txt) find_package or add_subdirectory.

Any thoughts on what we should change here or in libjxl so that this works?

@dg0yt FYI we have recently removed FindHWY.cmake - does that change anything?

jan-wassenberg:
Ah yes, libjxl indeed has its own FindHWY and so does ssimulacra2.

We can consider patching both libjxl and hwy but ideally it will remain working at each step - for example first add an opt-in to hwy, then use it in jxl, then remove the old way in hwy. Is that possible for the INTERFACE_INCLUDE_DIRECTORIES? (Perhaps by adding a CMake option that Highway sets if it has the hwy:: change, and jxl tests before using)

dg0yt:
> `find_package`

Bascially, migrations benefit from `HWYConfig.cmake` providing what `FindHWY.cmake` used to provide. This also enables use of `CMAKE_FIND_PACKAGE_PREFER_CONFIG`. Variables probably don't have to change at all. For targets, old flat names should no longer be advertised in documentation.

> `add_subdirectory`

IIUC this case is normally handled by `ALIAS` targets which provide the official namespaced external name with in a single buildsystem.

> ` $<TARGET_PROPERTY:hwy,INTERFACE_INCLUDE_DIRECTORIES>`

Ideally there is little reason to read such properties at all... 
There are migration paths, but you may call them ugly:
~~~cmake
 $<TARGET_PROPERTY:$<IF:$<TARGET_EXISTS:HWY::hwy>:HWY::hwy,hwy>,INTERFACE_INCLUDE_DIRECTORIES>
~~~

jan-wassenberg:
Thanks @dg0yt, I like your `<IF` approach, this seems like a good way forward. @malaterre , would you mind updating your patch with that and sending it to libjxl? After that goes in, we can then make this change in hwy.

jan-wassenberg:
Super, thanks for the heads-up. I'll begin the merge.

jan-wassenberg (approved):

jan-wassenberg:
FYI @eustas has encountered a possible bug (only in older versions) of CMake:
` Error evaluating generator expression:
    $<TARGET_EXISTS:hwy::hwy>
  Expression did not evaluate to a known generator expression
`
We are investigating possible workarounds.

Issue #5, opened 2020-09-03
ppc implementation ?


devnull31:
Hello, interesting project ! In the instruction matrix you speak about ppc instructions coverage, but I couldn't find relative header file, like for neon and x86. What about this ? Thanks !

jan-wassenberg:
Hi, thanks for your interest. FYI a major API update is in the works.

We do not yet have a PPC implementation. Would you be interested in adding one?

devnull31:
I don't know much about SIMD with PPC, that's why I would be interested buy your impl ^^ and as you have this instruction matrix with ppc, I thought you had at least some work in progress.

pramodk:
Same here - interested in PPC support but could only help in testing/validation aspects.

jan-wassenberg:
Hi, I understand. Let's reopen this in case someone would like to build (part of) a ppc implementation?

wyatt8740:
I would love to try this, but for starters, PowerPC doesn't have just one "SIMD" implementation, at least not historically. Altivec/VMX is now a part of Power ISA 2.03, but there's also older cores which might not have any SIMD extension or may have some different one. The PowerPC 750CL has something called "[paired singles](https://fail0verflow.com/media/files/ppc_750cl.pdf)" instead; it is likely that this chip is derived from the Nintendo Wii's "Broadway" chip design, so it is likely that the Wii also uses this instead of Altivec.

Some don't have any special SIMD implementations at all; in this case, the work would have to be done without them.

There's also the problem of me not having much low level experience with SIMD in general, let alone on POWER. I do have a Powerbook G4 (which has an older version of Altivec/VMX on it), so I might try to learn it and see if I can do anything to help. As silly as it is, I'd love to be able to make an excuse to still have a G4.

jan-wassenberg:
Interesting, hadn't heard of paired singles, reminds a bit of 3DNow. I agree that Altivec would have the most impact. 

On x86 we also have various instruction sets so what we end up doing is clustering them into SSE4, AVX2, AVX3. On SSSE3 CPUs, for simplicity, we fall back to scalar (no SIMD) because the few extra SSE4.1 ops aren't supported.

For POWER it could also make sense to check for some subset, but I'm not sure where to draw the line - is VMX enough or is VSX important for being able to implement all the ops?

johnplatts:
[...]

PR #1142, opened 2023-02-18
Added support for VSX on PPC8/PPC9/PPC10 targets

johnplatts:
I have added support for the HWY_PPC8/HWY_PPC9/HWY_PPC10 targets, and all of the unit tests complete successfully on little-endian HWY_PPC8/HWY_PPC9/HWY_PPC10 targets.

There are some unit tests that fail for big-endian HWY_PPC8/HWY_PPC9/HWY_PPC10 targets, and HWY_BROKEN_TARGETS includes HWY_PPC8/HWY_PPC9/HWY_PPC10 on big-endian HWY_PPC8/HWY_PPC9/HWY_PPC10 targets.

This implementation resolves issue #5.

### DIFF
--- a/hwy/ops/ppc_vsx-inl.h
+++ b/hwy/ops/ppc_vsx-inl.h
@@ -1,0 +1,4168 @@
+// Copyright 2019 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// 128-bit vectors for VSX
+// External include guard in highway.h - see comment there.
+
+// Must come before HWY_DIAGNOSTICS and HWY_COMPILER_GCC_ACTUAL
+#include "hwy/base.h"
+
+#pragma push_macro("vector")
+#pragma push_macro("pixel")
+#pragma push_macro("bool")
+
+#undef vector
+#undef pixel
+#undef bool
+
+#include <altivec.h>
+
+#pragma pop_macro("vector")
+#pragma pop_macro("pixel")
+#pragma pop_macro("bool")
+
+#include <stdio.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <string.h>  // memcpy
+
+#include "hwy/ops/shared-inl.h"
+
+HWY_BEFORE_NAMESPACE();
+namespace hwy {
+namespace HWY_NAMESPACE {
+namespace detail {
+
+template <typename T>
+struct Raw128;
+
+// Each Raw128 specialization defines the following typedefs:
+// - type:
+//   the backing Altivec/VSX raw vector type of the Vec128<T, N> type
+// - RawBoolVectType:
+//   the backing Altivec/VSX raw __bool vector type of the Mask128<T, N> type
+// - RawVectLaneType:
+//   the lane type of Raw128<T>::type
+// - AlignedLoadStoreRawVectType:
+//   the GCC/Clang vector type that is used for an aligned load of a full
+//   16-byte vector
+// - AlignedLoadStoreRawVectType:
+//   the GCC/Clang vector type that is used for an unaligned load of a full
+//   16-bit vector
+#define HWY_VSX_RAW128_SPECIALIZE(LANE_TYPE, RAW_VECT_LANE_TYPE, \
+                                  RAW_BOOL_VECT_LANE_TYPE) \
+template<> \
+struct Raw128<LANE_TYPE> { \
+  using type = __vector RAW_VECT_LANE_TYPE; \
+  using RawBoolVectType = __vector __bool RAW_BOOL_VECT_LANE_TYPE; \
+  using RawVectLaneType = RAW_VECT_LANE_TYPE; \
+  typedef LANE_TYPE AlignedLoadStoreRawVectType __attribute__(( \
+    __vector_size__(16), __aligned__(16), __may_alias__)); \
+  typedef LANE_TYPE UnalignedLoadStoreRawVectType __attribute__(( \
+    __vector_size__(16), __aligned__(alignof(LANE_TYPE)), __may_alias__)); \
+};
+
+HWY_VSX_RAW128_SPECIALIZE(int8_t, signed char, char)
+HWY_VSX_RAW128_SPECIALIZE(uint8_t, unsigned char, char)
+HWY_VSX_RAW128_SPECIALIZE(int16_t, signed short, short)
+HWY_VSX_RAW128_SPECIALIZE(uint16_t, unsigned short, short)
+HWY_VSX_RAW128_SPECIALIZE(int32_t, signed int, int)
+HWY_VSX_RAW128_SPECIALIZE(uint32_t, unsigned int, int)
+HWY_VSX_RAW128_SPECIALIZE(int64_t, signed long long, long long)
+HWY_VSX_RAW128_SPECIALIZE(uint64_t, unsigned long long, long long)
+HWY_VSX_RAW128_SPECIALIZE(float, float, int)
+HWY_VSX_RAW128_SPECIALIZE(double, double, long long)
+
+template<>
+struct Raw128<bfloat16_t> : public Raw128<uint16_t> {};
+
+template<>
+struct Raw128<float16_t> : public Raw128<uint16_t> {};
+
+#undef HWY_ALTIVEC_RAW128_SPECIALIZE
+
+}  // namespace detail
+
+template <typename T, size_t N = 16 / sizeof(T)>
+class Vec128 {
+  using Raw = typename detail::Raw128<T>::type;
+
+ public:
+  using PrivateT = T;                     // only for DFromV
+  static constexpr size_t kPrivateN = N;  // only for DFromV
+
+  // Compound assignment. Only usable if there is a corresponding non-member
+  // binary operator overload. For example, only f32 and f64 support division.
+  HWY_INLINE Vec128& operator*=(const Vec128 other) {
+    return *this = (*this * other);
+  }
+  HWY_INLINE Vec128& operator/=(const Vec128 other) {
+    return *this = (*this / other);
+  }
+  HWY_INLINE Vec128& operator+=(const Vec128 other) {
+    return *this = (*this + other);
+  }
+  HWY_INLINE Vec128& operator-=(const Vec128 other) {
+    return *this = (*this - other);
+  }
+  HWY_INLINE Vec128& operator&=(const Vec128 other) {
+    return *this = (*this & other);
+  }
+  HWY_INLINE Vec128& operator|=(const Vec128 other) {
+    return *this = (*this | other);
+  }
+  HWY_INLINE Vec128& operator^=(const Vec128 other) {
+    return *this = (*this ^ other);
+  }
+
+  Raw raw;
+};
+
+template <typename T>
+using Vec64 = Vec128<T, 8 / sizeof(T)>;
+
+template <typename T>
+using Vec32 = Vec128<T, 4 / sizeof(T)>;
+
+template <typename T>
+using Vec16 = Vec128<T, 2 / sizeof(T)>;
+
+// FF..FF or 0.
+template <typename T, size_t N = 16 / sizeof(T)>
+struct Mask128 {
+  typename detail::Raw128<T>::RawBoolVectType raw;
+};
+
+template <class V>
+using DFromV = Simd<typename V::PrivateT, V::kPrivateN, 0>;
+
+template <class V>
+using TFromV = typename V::PrivateT;
+
+// ------------------------------ Zero
+
+// Returns an all-zero vector/part.
+template <class D>
+HWY_API Vec128<TFromD<D>, HWY_MAX_LANES_D(D)> Zero(D /* tag */) {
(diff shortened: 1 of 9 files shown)


hwy/ops/ppc_vsx-inl.h
```diff
@@ -0,0 +1,4168 @@
...
+// ------------------------------ Load
+
+template <class D, HWY_IF_V_SIZE_D(D, 16), typename T = TFromD<D>>
+HWY_API Vec128<T> Load(D /* tag */, const T* HWY_RESTRICT aligned) {
+  using LoadRawVectType =
+    typename detail::Raw128<T>::AlignedLoadStoreRawVectType;
+  using ResultRawVectType =
+    typename detail::Raw128<T>::type;
+  return Vec128<T>{(ResultRawVectType)
+    *reinterpret_cast<const LoadRawVectType*>(aligned)};
```
jan-wassenberg, line 337: Is the type punning safe also for clang? Should we instead use memcpy/CopyBytes?

hwy/ops/ppc_vsx-inl.h
```diff
@@ -0,0 +1,4168 @@
...
+  return Vec128<uint64_t>{mul128_result};
+#else
+  // Need to swap the two halves of mul128_result on big-endian targets as
+  // the upper 64 bits of the product are in lane 0 of mul128_result and
+  // the lower 64 bits of the product are in lane 1 of mul128_result
+  return Vec128<uint64_t>{vec_sld(mul128_result, mul128_result, 8)};
+#endif
+#else
+  alignas(16) uint64_t mul[2];
+  const Half<Full128<uint64_t>> d2;
```
jan-wassenberg, line 2105: This works, but we could also use `Full64<uint64_t>`?

hwy/ops/ppc_vsx-inl.h
```diff
@@ -0,0 +1,4168 @@
...
+  const VU32 odd = Set(du32, 0xFFFF0000u);
+  const VU32 ae = ShiftLeft<16>(BitCast(du32, a));
+  const VU32 ao = And(BitCast(du32, a), odd);
+  const VU32 be = ShiftLeft<16>(BitCast(du32, b));
+  const VU32 bo = And(BitCast(du32, b), odd);
+  sum1 = MulAdd(BitCast(df32, ao), BitCast(df32, bo), sum1);
+  return MulAdd(BitCast(df32, ae), BitCast(df32, be), sum0);
+}
+
+// Even if N=1, the input is always at least 2 lanes, hence madd_epi16 is safe.
```
jan-wassenberg, line 2133: Update to vec_msum?

hwy/ops/ppc_vsx-inl.h
```diff
@@ -0,0 +1,4168 @@
...
+                        !hwy::IsSpecialFloat<FromT>()>* = nullptr,
+          HWY_IF_T_SIZE(FromT, sizeof(TFromD<D>) * 2)>
+HWY_API VFromD<D> DemoteTo(D /* tag */,
+                           Vec128<FromT, Rebind<FromT, D>().MaxLanes()> v) {
+  return VFromD<D>{vec_packs(v.raw, v.raw)};
+}
+
+template <class D, HWY_IF_V_SIZE_LE_D(D, 4),
+          hwy::EnableIf<IsSame<TFromD<D>, uint8_t>() ||
+                        IsSame<TFromD<D>, int8_t>()>* = nullptr>
```
jan-wassenberg, line 2308: HWY_IF_T_SIZE_D(D, 1) instead?

hwy/ops/ppc_vsx-inl.h
```diff
@@ -0,0 +1,4168 @@
...
+HWY_API VFromD<D> TruncateTo(D d,
+                             Vec128<FromT, Rebind<FromT, D>().MaxLanes()> v) {
+  const Rebind<MakeNarrow<FromT>, decltype(d)> d2;
+  return TruncateTo(d, TruncateTo(d2, v));
+}
+
+// ------------------------------ Demotions (full -> part w/ narrow lanes)
+
+template <class D, class FromT, HWY_IF_V_SIZE_LE_D(D, 8),
+          HWY_IF_UNSIGNED_D(D), HWY_IF_T_SIZE_ONE_OF_D(D, 0x6),
```
jan-wassenberg, line 2282: This works but we've subsequently expanded the literal to (1 << 1) | (1 << 2) to make it a bit more obvious.
Also, we could add a HWY_IF_SIGNED_INT(FromT) that expands to this, to shorten the EnableIf here? Seems used often, too.

hwy/ops/ppc_vsx-inl.h
```diff
@@ -0,0 +1,4168 @@
...
+
+// Also works for N < 8 because the first 16 4-tuples only reference bytes 0-6.
+template <class D, HWY_IF_T_SIZE_D(D, 2)>
+HWY_INLINE VFromD<D> IndicesFromBits128(D d, uint64_t mask_bits) {
+  HWY_DASSERT(mask_bits < 256);
+  const Rebind<uint8_t, decltype(d)> d8;
+  const Twice<decltype(d8)> d8t;
+  const RebindToUnsigned<decltype(d)> du;
+
+  // compress_epi16 requires VBMI2 and there is no permutevar_epi16, so we need
```
jan-wassenberg, line 2944: Update or remove x86-specific comment?

hwy/ops/ppc_vsx-inl.h
```diff
@@ -0,0 +1,4168 @@
...
+}
+
+template <class D, HWY_IF_T_SIZE_D(D, 2)>
+HWY_INLINE VFromD<D> IndicesFromNotBits128(D d, uint64_t mask_bits) {
+  HWY_DASSERT(mask_bits < 256);
+  const Rebind<uint8_t, decltype(d)> d8;
+  const Twice<decltype(d8)> d8t;
+  const RebindToUnsigned<decltype(d)> du;
+
+  // compress_epi16 requires VBMI2 and there is no permutevar_epi16, so we need
```
jan-wassenberg, line 3101: Also here.

hwy/ops/set_macros-inl.h
```diff
@@ -221,7 +237,49 @@
...
+#define HWY_NATIVE_FMA 1
+#define HWY_CAP_GE256 0
+#define HWY_CAP_GE512 0
+
+#define HWY_NAMESPACE N_PPC9
+
+#define HWY_TARGET_STR HWY_TARGET_STR_PPC9
+
+//-----------------------------------------------------------------------------
+// PPC10
```
jan-wassenberg, line 264: It seems fine to have a separate ----- section for PPC8..10, or we could consider putting them in one section and adding `#if` only where they differ, i.e. HWY_NAMESPACE/TARGET_STR? (Similar example is HWY_SVE2/HWY_SVE)

jan-wassenberg (changes requested):
Hi @johnplatts, this is awesome work, thank you and tip of the hat, looks like a great implementation. Great to see you've also successfully implemented runtime dispatch.

Just a few small change suggestions before we merge :)

FYI we have https://github.com/google/highway/pull/1138 ready to go, which adds two ops. Do you have a reference to recommend for POWER intrinsics so I could attempt to implement them? (Last I saw, the official reference PDF was huge and very slow to browse)

hwy/ops/ppc_vsx-inl.h
```diff
@@ -0,0 +1,4168 @@
...
+#undef pixel
+#undef bool
+
+#include <altivec.h>
+
+#pragma pop_macro("vector")
+#pragma pop_macro("pixel")
+#pragma pop_macro("bool")
+
+#include <stdio.h>
```
jan-wassenberg, line 36: Can/should we remove stdio?

hwy/ops/ppc_vsx-inl.h
johnplatts: Yes

hwy/ops/ppc_vsx-inl.h
johnplatts: Yes

hwy/ops/ppc_vsx-inl.h
```diff
@@ -0,0 +1,4168 @@
...
+template <typename T, size_t N>
+HWY_API Vec128<T, N> InsertLane(Vec128<T, N> v, size_t i, T t) {
+  typename detail::Raw128<T>::type raw_result = v.raw;
+  raw_result[i] = t;
+  return Vec128<T, N>{raw_result};
+}
+
+// ------------------------------ CombineShiftRightBytes
+
+// NOTE: The ShiftRightLanes operation below moves the elements of lo to the
```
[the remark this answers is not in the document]
johnplatts: I meant CombineShiftRightBytes instead of ShiftRightLanes here.

hwy/ops/ppc_vsx-inl.h
johnplatts: The ```Half<Full128<uint64_t>>``` can be replaced by ```Full64<uint64_t>``` here

hwy/ops/ppc_vsx-inl.h
johnplatts: Yes, go ahead and remove the stdio.h include.

hwy/ops/ppc_vsx-inl.h
johnplatts: The ```vec_msum``` intrinsic is not available for the ```bfloat16_t``` to ```float``` ```ReorderWidenMulAccumulate``` operation.

hwy/ops/ppc_vsx-inl.h
johnplatts: Yes

hwy/ops/ppc_vsx-inl.h
johnplatts: Yes

hwy/ops/ppc_vsx-inl.h
```diff
@@ -0,0 +1,4168 @@
...
+HWY_API Vec128<T, N> ShiftLeft(Vec128<T, N> v) {
+  static_assert(0 <= kBits && kBits < sizeof(T) * 8, "Invalid shift");
+  return ShiftLeftSame(v, kBits);
+}
+
+// ------------------------------ ShiftRight
+
+template <int kBits, typename T, size_t N,
+          hwy::EnableIf<!hwy::IsFloat<T>() &&
+                        !hwy::IsSpecialFloat<T>()>* = nullptr>
```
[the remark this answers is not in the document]
johnplatts: > Hi @johnplatts, this is awesome work, thank you and tip of the hat, looks like a great implementation. Great to see you've also successfully implemented runtime dispatch.
> 
> Just a few small change suggestions before we merge :)
> 
> FYI we have #1138 ready to go, which adds two ops. Do you have a reference to recommend for POWER intrinsics so I could attempt to implement them? (Last I saw, the official reference PDF was huge and very slow to browse)

The Power Vector Intrinsic Programming Reference that describes the Altivec/VSX intrinsics can be found at  https://files.openpower.foundation/s/9nRDmJgfjM8MpR7.

The Altivec Technology Programming Environments Manual, which describes the Altivec instructions, can be found at https://www.nxp.com/docs/en/reference-manual/ALTIVECPEM.pdf.

The Power Instruction Set Architecture Version 3.1B specification which describes the VSX instructions can be found at https://files.openpower.foundation/s/dAYSdGzTfW4j2r2.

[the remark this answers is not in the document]
johnplatts: > This is happening often, want to add a HWY_IF variant that expands to this?

Yes

The reason why I have the ```!hwy::IsFloat<T>() && !hwy::IsSpecialFloat<T>()``` in multiple places is to avoid unnecessary repetition as the Altivec/VSX intrinsic functions are already overloaded for the different Altivec/VSX vector types (and the overloading also exists in C as a compiler extension).

The ```!hwy::IsSpecialFloat<T>()``` is there as we do not want some of the functions enabled for the ```float16_t``` and ```bfloat16_t``` types.

hwy/ops/set_macros-inl.h
johnplatts: Can combine the HWY_PPC8/HWY_PPC9/HWY_PPC10 into a single section and adding ```#if``` only where they differ

hwy/ops/ppc_vsx-inl.h
johnplatts: The reason why a type punning cast is used (which is safe in the code above on GCC/Clang as the pointer is being cast to a vector type that has the ```__may_alias__``` attribute) is that older versions of GCC don't optimize the memcpy of a full 16-byte vector load down to a single ```lvx``` instruction but optimize the above type-punning vector load down to a single ```lvx``` instruction.

Newer versions of GCC such as GCC 11 and GCC 12 will optimize CopyBytes down to a single ```lvx``` instruction if ```__builtin_assume_aligned(aligned, 16)``` is passed into ```__builtin_memcpy```.

Here is the alternative implementation of a full 16-byte vector load using CopyBytes and HWY_ASSUME_ALIGNED:
```
template <class D, HWY_IF_V_SIZE_D(D, 16), typename T = TFromD<D>>
HWY_API Vec128<T> Load(D /* tag */, const T* HWY_RESTRICT aligned) {
  typename detail::Raw128<T>::type result;
  CopyBytes<sizeof(detail::Raw128<T>::type)>(
    HWY_ASSUME_ALIGNED(aligned, 16), &result);
  return result;
}
```

Note that HWY_ASSUME_ALIGNED is needed above for GCC/Clang to treat the above load as an aligned load.

jan-wassenberg (approved):
I'll make the proposed changes :)

hwy/ops/ppc_vsx-inl.h
jan-wassenberg: Got it, thanks for explaining. Seems safer to keep the current type punning, then. But I'll add a comment to make clear why we are doing this.

hwy/ops/ppc_vsx-inl.h
jan-wassenberg: Sorry to be unclear, I meant updating madd_epi16 in the comment to vec_msum; done.

PR #1154, opened 2023-02-24
Remove simple warnings from compilation log

malaterre:
This remove the following warnings (treated as error):

hwy/ops/ppc_vsx-inl.h:1994:29: error: unused variable ‘d’ [-Werror=unused-variable]
 1994 |   const DFromV<decltype(a)> d;
      |                             ^
cc1plus: all warnings being treated as errors


jan-wassenberg:
Thanks, I've made sure these are being fixed in the current batch of warning fixes :) (Rather than remove the d, I had intended to use it instead of `Simd<T, N, 0>`.)

PR #1167, opened 2023-02-28
Update CMakeLists.txt to include hwy/ops/ppc_vsx-inl.h

johnplatts:
Added hwy/ops/ppc_vsx-inl.h to HWY_SOURCES in CMakeLists.txt file


jan-wassenberg:
I believe the platforms/projects which use BUILD/gni do not support ppc so I'm tempted to leave this file out of them to avoid potential complications (such as a "build each header standalone" verification check failing because it doesn't support PPC), would that work for you?

Issue #1162, opened 2023-02-27
Should the HWY_SSE2 target be added?


johnplatts:
While most x86 CPU's manufactured in the last 10 years have at least SSSE3/SSE4.1/SSE4.2 support, adding the HWY_SSE2 target would allow older x86 CPU's that have SSE2 support but don't have SSSE3/SSE4.1/SSE4.2 support to take advantage of the SSE/SSE2 operations that are implemented in hwy/ops/x86_128-inl.h.

The biggest issue that would be there with the HWY_SSE2 target is that SSE2 lacks an instruction for the TableLookupBytes operation that is there on most of the other SIMD instruction sets that support 128-bit or larger vectors.

Here is how TableLookupLanes could be implemented for SSE2:
```
template <typename T, size_t N, typename TI, size_t NI>
HWY_API Vec128<TI, NI> TableLookupBytes(const Vec128<T, N> bytes,
                                        const Vec128<TI, NI> from) {
#if HWY_COMPILER_GCC_ACTUAL && HWY_HAS_BUILTIN(__builtin_shuffle)
  typedef uint8_t GccU8RawVect __attribute__((__vector_size__(16)));
  return Vec128<TI, NI>{(detail::Raw128<TI>::type)
    __builtin_shuffle((GccU8RawVect)bytes.raw, (GccU8RawVect)from.raw)};
#else
  const DFromV<decltype(from)> d;
  const Repartition<uint8_t, decltype(d)> du8;
  const Full128<uint8_t> du8_full;

  const DFromV<decltype(bytes)> d_bytes;
  const Repartition<uint8_t, decltype(d_bytes)> du8_bytes;

  alignas(16) uint8_t result_bytes[16];
  alignas(16) uint8_t u8_bytes[16];
  alignas(16) uint8_t from_bytes[16];
  
  Store(Vec128<uint8_t>{BitCast(du8_bytes, bytes).raw}, du8_full, u8_bytes);
  Store(Vec128<uint8_t>{BitCast(du8, from).raw}, du8_full, from_bytes);

  for(int i = 0; i < 16; i++)
    result_bytes[i] = u8_bytes[from_bytes[i] & 15];

  return BitCast(d, Load(du8, result_bytes));
#endif
}
```

The addition of the HWY_SSE2 target would also allow all of the 64-bit x86 CPU's to be targeted without having to enable the HWY_SCALAR or HWY_EMU128 targets as all of the 64-bit x86 CPU's have at least SSE2 support.

[...]

PR #1168, opened 2023-02-28
Added HWY_SSE2 target

johnplatts:
Added the HWY_SSE2 target, including support for the HWY_SSE2 target in hwy/ops/x86_128-inl.h.

Resolves issue #1162


hwy/targets.cc
```diff
@@ -174,6 +174,9 @@ HWY_INLINE constexpr uint64_t Bit(FeatureIndex index) {
   return 1ull << static_cast<size_t>(index);
 }
 
+constexpr uint64_t kGroupSSE2 =
+    Bit(FeatureIndex::kSSE) | Bit(FeatureIndex::kSSE2);
+
 constexpr uint64_t kGroupSSSE3 =
     Bit(FeatureIndex::kSSE) | Bit(FeatureIndex::kSSE2) |
```
jan-wassenberg, line 181: Let's reuse kGroupSSE2 in kGroupSSSE3, replacing its SSE and SSE2?

hwy/targets.cc
```diff
@@ -292,24 +299,25 @@ int64_t DetectTargets() {
...
     const int64_t min_avx3 = HWY_AVX3 | HWY_AVX3_DL;
     const int64_t min_avx2 = HWY_AVX2 | min_avx3;
-    // XMM
-    if (!IsBitSet(xcr0, 1)) {
-      bits &= ~(HWY_SSSE3 | HWY_SSE4 | min_avx2);
-    }
-    // YMM
-    if (!IsBitSet(xcr0, 2)) {
+    // XMM, YMM
+    if((xcr0 & 0x06) != 0x06) {
```
jan-wassenberg, line 316: Shouldn't we still be disabling SSE2/SSSE3/SSE4 if the XMM regs aren't being saved by the OS?

hwy/ops/x86_128-inl.h
```diff
@@ -32,7 +32,7 @@ HWY_DIAGNOSTICS_OFF(disable : 4703 6001 26494, ignored "-Wmaybe-uninitialized")
 #include <stdio.h>
 #if HWY_TARGET == HWY_SSSE3
```
jan-wassenberg, line 33: I think we should include emmintrin.h for SSE2, right?

hwy/ops/x86_128-inl.h
```diff
@@ -5473,21 +5730,36 @@ HWY_API VFromD<DTo> TruncateTo(DTo /* tag */, Vec128<From, 1> v) {
...
 template <class D, HWY_IF_U16_D(D)>
-HWY_API Vec32<uint16_t> TruncateTo(D /* tag */, Vec128<uint64_t> v) {
+HWY_API Vec32<uint16_t> TruncateTo(D d, Vec128<uint64_t> v) {
+#if HWY_TARGET == HWY_SSE2
+  const Vec128<uint16_t, 1> lo{v.raw};
+  const Vec128<uint16_t, 1> hi{_mm_unpackhi_epi64(v.raw, v.raw)};
+  return Combine(d, hi, lo);
+#else
+  (void)d;
+  const Repartition<uint8_t, DFromV<decltype(v)>> d8;
```
jan-wassenberg, line 5756: d8 is unused, please remove.

hwy/ops/x86_128-inl.h
```diff
@@ -1229,14 +1210,50 @@ HWY_API Vec128<int8_t, N> ShiftRight(const Vec128<int8_t, N> v) {
 template <typename T, size_t N, typename TI, size_t NI>
```
jan-wassenberg, line 1210: This function will likely have to be moved after the definition of Load/Store to address the build issue:

```
In file included from /home/runner/work/highway/highway/hwy/highway.h:383:
/home/runner/work/highway/highway/hwy/ops/x86_128-inl.h:1230:3: error: use of undeclared identifier 'Store'; did you mean 'N_SSSE3::Store'?
  Store(Vec128<uint8_t>{BitCast(du8_bytes, bytes).raw}, du8_full, u8_bytes);
```

jan-wassenberg (changes requested):
Wonderful, implementation looks good. Thank you @johnplatts for adding this, and it's great you also found native workarounds for the PSHUFB usages :D

Some changes will be required before we can merge this because at least older clang currently fails to build (see [status](https://github.com/google/highway/actions/runs/4289730427/jobs/7473058211)). Would you mind updating the code, pushing again to git and updating the pull request? The Github actions linked from the conversation tab of this pull request (where it says "Some checks were not successful") will re-run those tests afterwards.

hwy/ops/x86_128-inl.h
johnplatts: Yes

hwy/targets.cc
johnplatts: The lower 128 bits of the XMM registers are guaranteed to be saved on x86_64, even if bit 1 of XCR0 is not set, as the operating system will save the lower 128 bits of XMM0-XMM15 on x86_64 through another mechanism if bit 1 of XCR0 is not set.

The AMD64 ABI also guarantees that at least the lower 128 bits the XMM0-XMM15 registers and the SSE2 instructions are always available to programs, regardless of the value of the XCR0 register.

jan-wassenberg (approved):
Thanks for making the changes! Looks like the build is now passing.

hwy/targets.cc
jan-wassenberg: I see, was not aware of that, thanks for adding the comment.

hwy/ops/x86_128-inl.h
jan-wassenberg: Ah, my question was - shouldn't we be including emmintrin explicitly? Currently we are assuming that some other header is pulling that in for the HWY_SSE2 target. But no worries, I can make that change separately.

johnplatts: > Ah, my question was - shouldn't we be including emmintrin explicitly? Currently we are assuming that some other header is pulling that in for the HWY_SSE2 target. But no worries, I can make that change separately.

There are no includes that would already have been pulling in emmintrin.h for the HWY_SSE2 target with GCC/Clang compilers, but the intrin.h include in the hwy/base.h header will automatically include emmintrin.h with the MSVC compiler on x86 if -arch:SSE2 is enabled.

The emmintrin.h include is already there in line 31 of hwy/ops/x86_128-inl.h and needs to remain to avoid compiler errors. The emmintrin.h was also already being included explicitly for the SSSE3/SSE4/AVX2/AVX3 targets prior to the addition of the SSE2 target.

jan-wassenberg: Oops, indeed. That wasn't visible in the diff and I forgot it was already there. Thanks for pointing that out :)

jan-wassenberg:
Our CI has noticed an issue: we also want to update foreach_target.h so that SSE2 code is generated for dynamic dispatch. I'm happy to take care of that, and will mark the (new) pull request as being basically this pull request.

Alternatively, if you'd prefer to update this pull request with the change, please let me know soon (within a few hours if possible).

The code to add (before the HWY_SSSE3 section) would be:
```
#if (HWY_TARGETS & HWY_SSE2) && (HWY_STATIC_TARGET != HWY_SSE2)
#undef HWY_TARGET
#define HWY_TARGET HWY_SSE2
#include HWY_TARGET_INCLUDE
#ifdef HWY_TARGET_TOGGLE
#undef HWY_TARGET_TOGGLE
#else
#define HWY_TARGET_TOGGLE
#endif
#endif
```

PR #1172, opened 2023-02-28
Added prefetch prior to store for Stream operation on AArch64

johnplatts:
AArch64 has support for a non-temporal prefetch for store via the ```prfm PSTL1STRM, memory_location``` instruction, and GCC/Clang/MSVC all have intrinsics for the non-temporal prefetch for store operation.

```__builtin_prefetch(aligned, 1, 0)``` is also already there before the store in the PPC8/PPC9/PPC10 implementation of the Stream operation, which compiles to a ```dcbtst``` instruction on PPC8/PPC9/PPC10.


jan-wassenberg (approved):
Nice, thanks :)

PR #1173, opened 2023-03-01
Initial support for RISCV64/GCC-13

malaterre:


jan-wassenberg (changes requested):
Thanks! Looks like we want to enable gcv also for gcc. Let's update the comment to say that gcc(13) and recent clang both support V, but not yet runtime dispatch, so we add the gcv compiler flag, which then requires the CPU (now when using either compiler) to support V.

Issue #1174, opened 2023-03-01
compile with gcc on x86_64-pc-linux-gnu is broken with: cc1plus: error: attribute ‘avxvnni’ argument ‘target’ is unknown


stefson:
hey, I just compiled latest highway code from git and had this compile error: 

```
/usr/bin/x86_64-pc-linux-gnu-g++ -DHWY_SHARED_DEFINE -Dhwy_EXPORTS -I/var/tmp/portage/dev-cpp/highway-9999/work/highway-9999  -O2 -march=znver1 -pipe -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -Wno-builtin-macro-redefined -D__DATE__=\"redacted\" -D__TIMESTAMP__=\"redacted\" -D__TIME__=\"redacted\" -fmerge-all-constants -Wall -Wextra -Wconversion -Wsign-conversion -Wvla -Wnon-virtual-dtor -fmath-errno -fno-exceptions -MD -MT CMakeFiles/hwy.dir/hwy/per_target.cc.o -MF CMakeFiles/hwy.dir/hwy/per_target.cc.o.d -o CMakeFiles/hwy.dir/hwy/per_target.cc.o -c /var/tmp/portage/dev-cpp/highway-9999/work/highway-9999/hwy/per_target.cc
cc1plus: error: attribute ‘avxvnni’ argument ‘target’ is unknown
cc1plus: error: attribute ‘avxvnni’ argument ‘target’ is unknown
cc1plus: error: attribute ‘avxvnni’ argument ‘target’ is unknown
cc1plus: error: attribute ‘avxvnni’ argument ‘target’ is unknown
In file included from /var/tmp/portage/dev-cpp/highway-9999/work/highway-9999/hwy/highway.h:379,
                 from /var/tmp/portage/dev-cpp/highway-9999/work/highway-9999/hwy/per_target.cc:21,
                 from /var/tmp/portage/dev-cpp/highway-9999/work/highway-9999/hwy/foreach_target.h:202,
                 from /var/tmp/portage/dev-cpp/highway-9999/work/highway-9999/hwy/per_target.cc:20:
/var/tmp/portage/dev-cpp/highway-9999/work/highway-9999/hwy/ops/x86_512-inl.h: In function ‘__m512i hwy::N_AVX3_ZEN4::detail::BitCastToInteger(__m512i)’:
[...]

PR #1176, opened 2023-03-01
Changed AVXVNNI to AVX512VNNI

johnplatts:
Changed defined(__AVXVNNI__) with defined(__AVX512VNNI__) in hwy/detect_targets.h, and changed avxvnni with avx512vnni in hwy/detect_targets.h.

There are also some x86 CPU's with AVX512 support that support AVX512_VNNI but not the new VEX-encoded AVX-VNNI instructions available on Alder Lake, Sapphire Rapids, and newer Intel CPU's.

Resolves issue #1174

### DIFF
--- a/hwy/detect_targets.h
+++ b/hwy/detect_targets.h
@@ -381,7 +381,7 @@
 #endif
 
 // TODO(janwas): not yet known whether these will be set by MSVC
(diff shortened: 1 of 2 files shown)


jan-wassenberg (approved):
Thank you!

PR #1177, opened 2023-03-01
Added hwy/ops/ppc_vsx-inl.h to BUILD and hwy.gni files

johnplatts:


jan-wassenberg:
I am a bit concerned about adding this to those files because two projects that use them AFAIK do not support PPC, and adding the header might now or in future run into trouble with the build step that compiles each header standalone.

Is it important to you that we also add the file to these other build systems? If so, one thing we can do is have Copybara insert them only into the open-source version.

jan-wassenberg:
Let's close this for now, please feel free to re-open if you or anyone else require this file in the Bazel/GN build systems :)

Issue #1169, opened 2023-02-28
Multiple test failures on PPC8/big-endian


malaterre:
Currently HWY_PPC8 fails on big-endian arches with:

```
The following tests FAILED:
        149 - HwyCompressTestGroup/HwyCompressTest.TestAllCompress/PPC8  # GetParam() = 562949953421312 (SIGTRAP)
        165 - HwyConvertTestGroup/HwyConvertTest.TestAllTruncate/PPC8  # GetParam() = 562949953421312 (SIGTRAP)
        177 - HwyCryptoTestGroup/HwyCryptoTest.TestAllCLMul/PPC8  # GetParam() = 562949953421312 (SIGTRAP)
        185 - HwyDemoteTestGroup/HwyDemoteTest.TestAllReorderDemote2To/PPC8  # GetParam() = 562949953421312 (SIGTRAP)
        221 - HwyInterleavedTestGroup/HwyInterleavedTest.TestAllLoadStoreInterleaved3/PPC8  # GetParam() = 562949953421312 (SIGTRAP)
        255 - HwyMaskTestGroup/HwyMaskTest.TestAllStoreMaskBits/PPC8  # GetParam() = 562949953421312 (SIGTRAP)
        271 - HwyMulTestGroup/HwyMulTest.TestAllMul/PPC8  # GetParam() = 562949953421312 (SIGTRAP)
        281 - HwyMulTestGroup/HwyMulTest.TestAllReorderWidenMulAccumulate/PPC8  # GetParam() = 562949953421312 (SIGTRAP)
        283 - HwyMulTestGroup/HwyMulTest.TestAllRearrangeToOddPlusEven/PPC8  # GetParam() = 562949953421312 (SIGTRAP)
        287 - HwyReductionTestGroup/HwyReductionTest.TestAllMinMaxOfLanes/PPC8  # GetParam() = 562949953421312 (SIGTRAP)
```

jan-wassenberg:
@johnplatts would you mind looking into these? Or do you think it is worth supporting big-endian SIMD? One option could be to simply mark HWY_PPC* as broken in big endian, so HWY_EMU128 will instead be used.

jan-wassenberg:
> They are... I am simply investigating why.

Right. It is very easy to have big-endian specific bugs. The BitCast you mention is by itself portable, but generally requires endian-specific indices before/afterwards.

[...]

PR #1178, opened 2023-03-02
Fixed issue with LoadInterleaved3 implementation in generic_ops-inl.h on big-endian targets

johnplatts:
Modified the LoadInterleaved3 implementation for 8 16-bit lanes in generic_ops-inl.h to use uint8_t arrays for the indices instead of uint16_t indices.

This change to LoadInterleaved3 addresses one of the issues that is there in issue https://github.com/google/highway/issues/1169.


jan-wassenberg (approved):
Ah, this was needlessly little-endian-specific, thanks for fixing it.

jan-wassenberg:
> odd that this did not show up earlier on s390x/ppc64 or powerpc test suite

I think it's because PPC* are disabled/marked broken on big endian, manually re-enabling them is so far the first and only set of vector intrinsics running in big endian. By contrast, current s390x builds are using the software fallback emu128 which provides its own LoadInterleaved rather than using generic_ops.

PR #1183, opened 2023-03-02
Fixed issues on big-endian PPC8/PPC9/PPC10 targets

johnplatts:
Made changes to hwy/ops/ppc_vsx-inl.h to fix issues that are there on big-endian PPC8/PPC9/PPC10 targets.

Fixes issue #1169

### DIFF
--- a/hwy/ops/ppc_vsx-inl.h
+++ b/hwy/ops/ppc_vsx-inl.h
@@ -329,15 +329,39 @@ HWY_API Vec128<T, N> Abs(Vec128<T, N> v) {
 
 // ------------------------------ CopySign
 
-template <typename T, size_t N>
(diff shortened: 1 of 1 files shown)


jan-wassenberg:
Outstanding :D I will review tomorrow. In the meantime, would you like to remove the HWY_BROKEN_PPC in detect_targets.h?

johnplatts:
> Outstanding :D I will review tomorrow. In the meantime, would you like to remove the HWY_BROKEN_PPC in detect_targets.h?

Yes

jan-wassenberg (approved):
Nice. A request: clang apparently doesn't like C-style casts of PPC vectors, they cause warnings. I will continue to make passes over the file, replacing them with reinterpret_cast<> or BitCast (the latter is more compact but requires extra `Simd<>` tags). Let's avoid adding more C-style casts :)

PR #1203, opened 2023-03-08
Fix compilation on PPC

malaterre:
Fixes compilation error:

highway/hwy/targets.cc:49:10: fatal error: asm/hwcap.h: No such file or directory
   49 | #include <asm/hwcap.h>
      |          ^~~~~~~~~~~~~
compilation terminated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,6 +263,10 @@ endif()  # !MSVC
 
 include(CheckIncludeFile)
 check_include_file(sys/auxv.h  HAVE_SYS_AUXV_H)
+check_include_file(asm/hwcap.h HAVE_ASM_HWCAP_H)
+if (NOT HAVE_ASM_HWCAP_H)
+  list(APPEND HWY_FLAGS -DTOOLCHAIN_MISS_ASM_HWCAP_H)
+endif()
 
 # By default prefer STATIC build (legacy behavior)
 option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
--- a/hwy/targets.cc
+++ b/hwy/targets.cc
@@ -42,12 +42,15 @@
 #include <cpuid.h>
 #endif  // HWY_COMPILER_MSVC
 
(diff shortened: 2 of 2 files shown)


jan-wassenberg:
Thanks, this seems reasonable to me. Or should we add another CMake variable like TOOLCHAIN_MISS_SYS_AUXV_H to check for this header?

jan-wassenberg:
Indeed we only add that to the compiler flags: `list(APPEND HWY_FLAGS -DTOOLCHAIN_MISS_SYS_AUXV_H=1)`
What's required - adding an actual CMake variable? Would you like to add that?

jan-wassenberg (changes requested):
Here also I think we need to turn the CMake variable into an actual compiler flag by prepending `-D`. So `target_compile_definitions(hwy PRIVATE -DHAVE_ASM_HWCAP_H)`?

jan-wassenberg (approved):
I see, thanks for explaining. Looks good, then :)

jan-wassenberg (approved):
Oops, just realized: as in #1205, there is one thing we still want to change.
Let's also invert the logic here and include the header `#ifndef MISSING_ASM_HWCAP`, so that the Bazel-based build is unaffected and does not require an additional -D flag.

jan-wassenberg (changes requested):
(Changed to 'request changes', see above.)

jan-wassenberg:
PR #1205 changing CMake from private flag to public looks good.
This one is also fine, but do you want to switch it to the same check_include_file + target_compile_options + `#ifndef MISSING_ASM_HWCAP` as in #1205? I think that would be more robust.

For background on Bazel: we would like the current code to continue working as-is. Introducing a new HAVE_ flag to the code, and requiring CMake to set it, would also require us to add a flag to the Bazel build system, which is undesirable.
Hence the idea is to allow CMake to set a MISSING_ flag if it wants, and Bazel is free to ignore/not set that flag. FYI our Bazel build anyway does not support PPC.

jan-wassenberg (approved):
Thanks for making the change!

PR #1204, opened 2023-03-08
Allow 32bits compilation on ppc64el

malaterre:


jan-wassenberg (approved):

jan-wassenberg:
We can differentiate if/when there is something that must be done differently between 32/64-bit. On x86 that is required because some intrinsics are missing.

PR #1205, opened 2023-03-08
cmake: Make sure to export public definition

malaterre:
HAVE_SYS_AUXV_H was being set in the PRIVATE definition list

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,9 +263,6 @@ endif()  # !MSVC
 
 include(CheckIncludeFile)
 check_include_file(sys/auxv.h  HAVE_SYS_AUXV_H)
-if (NOT HAVE_SYS_AUXV_H)
-  list(APPEND HWY_FLAGS -DTOOLCHAIN_MISS_SYS_AUXV_H=1)
-endif()
 
 # By default prefer STATIC build (legacy behavior)
 option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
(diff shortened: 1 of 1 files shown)


jan-wassenberg (changes requested):
I think we're missing a step here. The new HAVE_SYS_AUXV is only a cmake variable. Projects that use this pattern then have a configure_file containing `#cmakedefine` for each variable, which CMake then converts into a C++ macro of the same name which we can test via `#if` in the code.

This seems unnecessarily complex to me (requires including another config header from code, and adding a config.cmake.in file). How about we retain the existing pattern of `if cmake_variable: list(APPEND HWY_PUBLIC_FLAGS -D..)` and then HWY_PUBLIC_FLAGS (or some better name) can be non-private target_compile_options?

jan-wassenberg:
As in #1203, I think we need `-D`, so `target_compile_options(hwy PUBLIC -DHAVE_SYS_AUXV_H)`?

jan-wassenberg (changes requested):
Oh, there is one thing we still want to change. As mentioned in https://github.com/google/highway/commit/1cab22047a6ef750ef2bc38ad47d6e765d6d376a, we do not always/only build with CMake. Let's invert the logic here to only opt-out of the include if `!HAVE`, so that the Bazel-based build is unaffected and does not require an additional `-D` flag.

mo271:
Could it be that this broke our MSYS2 build?
 https://github.com/libjxl/libjxl/actions/runs/4373916712/jobs/7652699512

mo271:
Thanks for the superquick fix, @malaterre!

PR #1211, opened 2023-03-09
Add fake SSE2 target on ppc64el arch

malaterre:


jan-wassenberg:
hm, I'm curious about the use case. Typically apps do not have to care what target they are running. Is your intention to get some kind of vector support even where we don't have PPC8 available?

jan-wassenberg:
OK, how do you mean perform: speed or correctness? The Highway tests generally only check for correctness. We could perhaps use bench_sort to compare the throughput with several targets, but there is no infra in place for that yet.
Is there any particular reason you are concerned about the PPC8 speed/correctness?

jan-wassenberg:
Ah, but aren't those dependent on the performance of the emulator? I'm not sure "faster in emulator" translates into "faster on real CPU".

johnplatts (commented):
```HWY_TARGET_STR_SSE2``` needs to be defined as ```"sse2"``` if compiling for real x86 targets, and there will be compiler errors if Google Highway is compiled on 32-bit x86 with GCC or Clang without the ```-msse2``` compiler option if ```HWY_TARGET_STR_SSE2``` is blank.

Here is how the HWY_TARGET_STR_SSE2 macro should be defined in hwy/ops/set_macros-inl.h:
```
#if HWY_ARCH_X86
#define HWY_TARGET_STR_SSE2 "sse2"
#else
#define HWY_TARGET_STR_SSE2 ""
#endif
```

jan-wassenberg:
> Why are you talking about emulator here ?

We do test/run on an emulator, so I'm concerned about (automated) usage of performance numbers. If we're occasionally manually looking at results, especially when you are running on native, then that's fine.

> unit test are no benchmark tools, or

Yes, that is correct. Not that many of our tests are useful for benchmarking; bench_sort is, but it only covers a tiny subset of the ops.

jan-wassenberg:
Nice find :)

PR #1241, opened 2023-03-20
Fix compilation of hwy_benchmark on ppc32

malaterre:
Fixes compilation error:

/usr/bin/ld: CMakeFiles/hwy_benchmark.dir/hwy/examples/benchmark.cc.o: undefined reference to symbol '__atomic_load_8@@LIBATOMIC_1.0'
/usr/bin/ld: /usr/lib/powerpc-linux-gnu/libatomic.so.1: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status


jan-wassenberg (approved):
Thanks! We can consider putting both hwy and ATOMICS_LIBRARIES on a single line, or keep this if you prefer.

PR #1243, opened 2023-03-20
Fixed FlagsFromCPUID to only clear the AVX2/AVX3 bits on x86_64 if bit 0 of XCR0 is not set

johnplatts:
The previous version of FlagsFromCPUID failed to clear the HWY_AVX2, HWY_AVX3, HWY_AVX3_DL, and HWY_AVX3_ZEN4 bits on x86_64 if bit 1 of XCR0 is not set. This updated version will clear the HWY_AVX2, HWY_AVX3, HWY_AVX3_DL, and HWY_AVX3_ZEN4 bits on x86_64 if bit 1 of XCR0 is not set.


jan-wassenberg (approved):
Thanks, I didn't see this while reviewing :)

Issue #1124, opened 2023-01-24
Fixes and enhancements to DemoteTo and ReorderDemote2To methods


johnplatts:
<b>Update:</b> The _mm_packus_epi16 (packuswb) operation converts signed 16-bit integers to unsigned 8-bit integers using unsigned saturation according to the Intel 64 and IA-32 Architectures Software Developer's Manual. The _mm_packus_epi32 (packusdw) operation converts signed 32-bit integers to unsigned 16-bit integers using unsigned saturation according to the Intel 64 and IA-32 Architectures Software Developer's manual.

The current implementation of DemoteTo for i32->u16, i32->u8, and i16->u8 in hwy/ops/x86_128-inl.h are correct.

Here is how the partial vector ReorderDemote2To implementations can be improved for HWY_SSSE3/HWY_SSE4/HWY_AVX2/HWY_AVX3 targets:
```
// Specializations for partial vectors because packs_epi32 sets lanes above 2*N.
HWY_API Vec128<int16_t, 2> ReorderDemote2To(Simd<int16_t, 2, 0> dn,
                                            Vec128<int32_t, 1> a,
                                            Vec128<int32_t, 1> b) {
  const Twice<DFromV<decltype(a)>> d2;
  return DemoteTo(dn, Combine(d2, b, a));
}
HWY_API Vec128<int16_t, 4> ReorderDemote2To(Simd<int16_t, 4, 0> dn,
                                            Vec128<int32_t, 2> a,
                                            Vec128<int32_t, 2> b) {
  return Vec128<int16_t>{_mm_shuffle_epi32(_mm_packs_epi32(a.raw, b.raw), 0xD8)};
}
```

Overloads of DemoteTo should be added for u16->u8, u32->u8, and u32->u16 as the DemoteTo operations for u16->u8, u32->u8, and u32->u16 can be implemented using Min and TruncateTo if necessary.

Here is how the DemoteTo operations can be implemented for u16->u8, u32->u8, and u32->u16 for HWY_SSSE3/HWY_SSE4/HWY_AVX2/HWY_AVX3:
```
template <size_t N>
HWY_API Vec128<uint16_t, N> DemoteTo(Simd<uint16_t, N, 0> /* tag */,
                                     const Vec128<uint32_t, N> v) {
#if HWY_TARGET <= HWY_AVX3
  // NOTE: The _mm_cvtusepi32_epi16 operation is a saturated conversion
  // of unsigned 32-bit integers to unsigned 16-bit integers.
[...]

PR #1247, opened 2023-03-21
Added more demotion operations

johnplatts:
Added the following new demotion operations:
- DemoteTo from unsigned integer types to signed integer types
- DemoteTo from unsigned integer types to unsigned integer types
- ReorderDemote2To for {i,u}16->{i,u}8, i32->u16, u32->{u,i}16, and {i,u}64->{i,u}32
- OrderedDemote2To for {i,u}16->{i,u}8, {i,u}32->{u,i}16, {i,u}64->{i,u}32, and f32->bf16

Resolves issue #1124

### DIFF
--- a/hwy/ops/rvv-inl.h
+++ b/hwy/ops/rvv-inl.h
@@ -945,6 +945,15 @@ HWY_RVV_FOREACH_I(HWY_RVV_SHIFT_II, Shr, sra, _ALL)
 
 // ------------------------------ Min
 
+namespace detail {
+
+HWY_RVV_FOREACH_U(HWY_RVV_RETV_ARGVS, MinS, minu_vx, _ALL)
+HWY_RVV_FOREACH_I(HWY_RVV_RETV_ARGVS, MinS, min_vx, _ALL)
+HWY_RVV_FOREACH_F(HWY_RVV_RETV_ARGVS, MinS, fmin_vf, _ALL)
+
+}  // namespace detail
+
+
 HWY_RVV_FOREACH_U(HWY_RVV_RETV_ARGVV, Min, minu, _ALL)
 HWY_RVV_FOREACH_I(HWY_RVV_RETV_ARGVV, Min, min, _ALL)
 HWY_RVV_FOREACH_F(HWY_RVV_RETV_ARGVV, Min, fmin, _ALL)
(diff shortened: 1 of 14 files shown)


jan-wassenberg:
Wow, nice :) I am reviewing the code..

hwy/tests/demote_test.cc
```diff
@@ -54,42 +54,106 @@ struct TestDemoteTo {
...
+  const ForDemoteVectors<TestDemoteTo<uint8_t>, 2> from_i32_to_u8;
+  from_i32_to_u8(int32_t());
+  from_i32_to_u8(uint32_t());
+
+  const ForDemoteVectors<TestDemoteTo<int8_t>, 2> from_i32_to_i8;
+  from_i32_to_i8(int32_t());
+  from_i32_to_i8(uint32_t());
+
+#if HWY_HAVE_INTEGER64
+#if HWY_HAVE_SCALABLE
```
jan-wassenberg, line 107: I'm curious why we want to check HWY_HAVE_SCALABLE here? It's intended to be legit to set kPow2=3 in both fixed-size and scalable targets.

hwy/tests/demote_test.cc
```diff
@@ -257,8 +389,249 @@ class TestReorderDemote2To {
...
+    (void)d;
+#endif
+  }
+};
+
 HWY_NOINLINE void TestAllReorderDemote2To() {
   ForShrinkableVectors<TestReorderDemote2To>()(float());
+  ForShrinkableVectors<TestIntegerReorderDemote2To>()(int16_t());
+  ForShrinkableVectors<TestIntegerReorderDemote2To>()(uint16_t());
+  ForShrinkableVectors<TestIntegerReorderDemote2To>()(int32_t());
```
jan-wassenberg, line 485: Consider using ForUI163264 to replace these 6 calls?

hwy/tests/demote_test.cc
```diff
@@ -257,8 +389,249 @@ class TestReorderDemote2To {
...
+    DoIntegerOrderedDemote2ToTest(dn_i, T(), d);
+    DoIntegerOrderedDemote2ToTest(dn_u, T(), d);
+#else
+    (void)d;
+#endif
+  }
+};
+
+HWY_NOINLINE void TestAllOrderedDemote2To() {
+  ForShrinkableVectors<TestIntegerOrderedDemote2To>()(int16_t());
```
jan-wassenberg, line 626: Also here can use ForUI163264.

jan-wassenberg (approved):
Very nice! Some minor comments, please let me know if you'd like to change the code, I'm also happy to make these changes after the pull request is merged.

hwy/tests/demote_test.cc
johnplatts: The reason why HWY_HAVE_SCALABLE is here is because the code was failing to compile on the HWY_RVV target when I originally wrote the unit tests.

hwy/ops/rvv-inl.h
```diff
@@ -2054,6 +2121,20 @@ HWY_API vint32m4_t DemoteTo(Simd<int32_t, N, 2> d, const vfloat64m8_t v) {
   return __riscv_vfncvt_rtz_x_f_w_i32m4(v, Lanes(d));
 }
 
+// SEW is for the source so we can use _DEMOTE_TO_SHR_16.
```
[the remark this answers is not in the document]
johnplatts: The reason why I had _DEMOTE_TO_SHR_16 in the comment above is that the comment was copied from a comment that had just _DEMOTE instead of _DEMOTE_VIRT.

hwy/tests/demote_test.cc
jan-wassenberg: Interesting. I will investigate after this is merged.

hwy/tests/demote_test.cc
jan-wassenberg: OK, I can make that change after this is merged.

hwy/ops/rvv-inl.h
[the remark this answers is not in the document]
jan-wassenberg: Thanks for the heads-up, I am updating the other comments as well.

_DEMOTE_VIRT was added after those comments were written. The comment refers to the choice of SEW vs SEWD (double) or SEWH (half) arguments.

Issue #885, opened 2022-08-01
Extend support of Ge/Le comparison for integer type


hellishfire:
Currently highway only supports Ge/Le comparison for float type. I suppose it's not difficult to extend this support for integer type. For hardware where no native instruction is available, simply implement them on top of existing Gt/Lt and invert the result?

jan-wassenberg:
Hi @hellishfire, sounds good, that would indeed be useful. Would you like to send a pull request for this?

jan-wassenberg:
Thank you @johnplatts for implementing these!

PR #1249, opened 2023-03-21
Added Le and Ge operations for integer vectors

johnplatts:
Resolves issue #885


jan-wassenberg (approved):
Nice, thank you for adding these!

PR #1262, opened 2023-03-28
Made enhancements to a few operations in x86_128-inl.h

johnplatts:
Made enhancements to a few of the shuffle operations on the HWY_SSE2 target.

Also updated the 32-bit signed right shift operation in x86_128-inl.h to use the _mm_srav_epi32 intrinsic on the HWY_AVX2 target as the _mm_srav_epi32 intrinsic is supported on AVX2 according to the Intel Intrinsics guide at https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html.

### DIFF
--- a/hwy/ops/x86_128-inl.h
+++ b/hwy/ops/x86_128-inl.h
@@ -1405,8 +1405,9 @@ HWY_API Vec128<TI, NI> TableLookupBytes(const Vec128<T, N> bytes,
 #if HWY_TARGET == HWY_SSE2
 #if HWY_COMPILER_GCC_ACTUAL && HWY_HAS_BUILTIN(__builtin_shuffle)
   typedef uint8_t GccU8RawVectType __attribute__((__vector_size__(16)));
-  return Vec128<TI, NI>{(typename detail::Raw128<TI>::type)
-    __builtin_shuffle((GccU8RawVectType)bytes.raw, (GccU8RawVectType)from.raw)};
+  return Vec128<TI, NI>{reinterpret_cast<typename detail::Raw128<TI>::type>(
+    __builtin_shuffle(reinterpret_cast<GccU8RawVectType>(bytes.raw),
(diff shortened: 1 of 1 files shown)


hwy/ops/x86_128-inl.h
```diff
@@ -4497,9 +4505,15 @@ HWY_API VFromD<D> Reverse2(D d, VFromD<D> v) {
...
 template <class D, typename T = TFromD<D>, HWY_IF_T_SIZE(T, 2)>
 HWY_API VFromD<D> Reverse2(D d, VFromD<D> v) {
-#if HWY_TARGET <= HWY_AVX3 || HWY_TARGET == HWY_SSE2
+#if HWY_TARGET <= HWY_AVX3
   const Repartition<uint32_t, decltype(d)> du32;
   return BitCast(d, RotateRight<16>(BitCast(du32, v)));
+#elif HWY_TARGET == HWY_SSE2
+  constexpr size_t kN = MaxLanes(d);
+  __m128i shuf_result = _mm_shufflelo_epi16(v.raw, _MM_SHUFFLE(2, 3, 0, 1));
+  if(kN > 4)
```
jan-wassenberg, line 4514: A request for future code: we use clang-format with the Google style. Please generally write braces for if-conditions unless it's something trivial such as a return or break that fits in one line.

So:
```
if (kN > 4) {
  shuf_result = ..;
}
```
No worries about this one, will update it the next time I edit this file.

jan-wassenberg (approved):
Nice, thanks! Good catch on the srav.

FYI we have made some changes to the security settings of this repo, hopefully the merge still works.

hwy/ops/x86_128-inl.h
jan-wassenberg: hm, looks like that mainly validates/checks the code - or does it also send pull requests? If not, we are mostly covered already because our IDE does display lint warnings, and also auto-formats when saving that file.

PR #1274, opened 2023-03-30
Directly use the specific macro __riscv_v_intrinsic

malaterre:


jan-wassenberg:
Good idea 👍 How about we implement this in the actual check for RVV? We can replace
`#if HWY_ARCH_RVV && defined(__riscv_vector)`
With your `#if HWY_ARCH_RVV && defined(__riscv_v_intrinsic) && __riscv_v_intrinsic >= 11000`.
Then remove the HWY_BROKEN_RVV - what do you think?

jan-wassenberg (approved):
Thanks for updating the patch, looks good!

jan-wassenberg:
FYI our internal CI is temporarily stuck, I am working on it.

PR #1276, opened 2023-04-03
Added LeadingZeroCount and a few more new operations plus bug fixes

johnplatts:
The following operations were added:
- LeadingZeroCount - returns the number of leading zeros for each lane (similar to the AVX3 VPLZCNTD/VPLZCNTQ instruction or the GCC/Clang __builtin_clz intrinsic)
- TrailingZeroCount - returns the number of trailing zeros for each lane (similar to the x86 BSF/TZCNT instruction, the GCC/Clang __builtin_ctz intrinsic, or the MSVC _BitScanForward intrinsic)
- HighestSetBitIndex - returns the index of the highest set bit for each lane (similar to the x86 BSR instruction or the MSVC _BitScanReverse intrinsic)
- OrderedTruncate2To - similar to OrderedDemote2To, but truncates values instead of saturating values
- ReverseLaneBytes - swaps the bytes of each lane, more efficient than TableLookupBytes on NEON/SVE/SVE2/SSE2/PPC9/PPC10, used to implement ReverseBits for I16/U16/I32/U32/I64/U64 on some platforms
- ReverseBits - reverses the bits of each lane, used to implement TrailingZeroCount on NEON/SVE

Also fixed bugs with AVX3_DL TruncateTo, SVE ConcatEven for partial vectors, and SVE ConcatOdd for partial vectors.

All of the Highway tests now pass successfully on the HWY_AVX3_DL target with the TruncateTo bug fixes.

Also fixed compilation errors with the FirstN implementation in x86_512-inl.h on 32-bit x86. All of the HWY_AVX3 tests also now pass on 32-bit x86 when compiled with GCC 12.

Some of the Highway source files have also been reformatted.

### DIFF
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -45,6 +45,16 @@ using Vec = decltype(Zero(D()));
 template <class D>
 using Mask = decltype(MaskFromVec(Zero(D())));
 
+namespace detail {
+#if HWY_TARGET == HWY_RVV
+template <class T>
+using LargestVecScalableTag = ScalableTag<T, HWY_MAX_POW2>;
+#else
+template <class T>
+using LargestVecScalableTag = ScalableTag<T>;
+#endif
+}  // namespace detail
+
 // Returns the closest value to v within [lo, hi].
 template <class V>
 HWY_API V Clamp(const V v, const V lo, const V hi) {
@@ -1022,8 +1032,8 @@ HWY_API VFromD<DN> DemoteTo(DN dn, V v) {
   // using an unsigned Min operation.
   const auto max_signed_val = Set(dn, hwy::HighestValue<TFromD<DN>>());
 
-  return BitCast(dn, Min(BitCast(dn_u, i2i_demote_result),
-                         BitCast(dn_u, max_signed_val)));
+  return BitCast(
+      dn, Min(BitCast(dn_u, i2i_demote_result), BitCast(dn_u, max_signed_val)));
 }
 
 #if HWY_TARGET != HWY_SCALAR || HWY_IDE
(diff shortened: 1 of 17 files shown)


jan-wassenberg:
Wow, very interesting set of new ops, thanks for adding them! I will review today. For faster review, would you mind creating  multiple pull requests for such large changes in future? (I understand that ReverseBits is used to implement ctz, so it makes sense to combine them.)

hwy/ops/generic_ops-inl.h
```diff
@@ -2227,6 +2535,117 @@ HWY_API VFromD<D> LoadExpand(MFromD<D> mask, D d,
...
+#else
+  const RebindToUnsigned<decltype(d)> d_shift;
+#endif
+
+  const auto v_to_shift = BitCast(d_shift, v);
+  const auto shl_result = BitCast(d, ShiftLeft<kShiftAmt>(v_to_shift));
+  const auto shr_result = BitCast(d, ShiftRight<kShiftAmt>(v_to_shift));
+  const auto shr_result_mask =
+      BitCast(d, Set(du, static_cast<uint8_t>(kShrResultMask)));
+  return Or(And(shr_result, shr_result_mask),
```
jan-wassenberg, line 2607: If you wanted, you could use IfVecThenElse instead of or+and+andnot, that's faster on AVX3+.

hwy/tests/logical_test.cc
```diff
@@ -226,6 +226,236 @@ HWY_NOINLINE void TestAllPopulationCount() {
...
+    const RebindToUnsigned<decltype(d)> du;
+    size_t N = Lanes(d);
+    auto data = AllocateAligned<T>(N);
+    auto lzcnt = AllocateAligned<T>(N);
+
+    constexpr T kNumOfBitsInT{sizeof(T) * 8};
+    for (size_t j = 0; j < N; j++) {
+      lzcnt[j] = kNumOfBitsInT;
+    }
+    HWY_ASSERT_VEC_EQ(d, lzcnt.get(), LeadingZeroCount(Zero(d)));
```
jan-wassenberg, line 263: No need to change this but just FYI it is possible to `HWY_ASSERT_VEC_EQ(d, Set(d, kNumOfBitsInT), LeadingZeroCount())`.

jan-wassenberg (changes requested):
Wow. I respect your clever implementation including GFNI and appreciate your attention to detail in fixing LANE_SIZE -> T_SIZE + the nested loop counter i. Thanks also for fixing ConvertTo + Concat*.
Some suggestions - would you like to make these changes?

hwy/ops/generic_ops-inl.h
```diff
@@ -2227,6 +2535,117 @@ HWY_API VFromD<D> LoadExpand(MFromD<D> mask, D d,
...
+#ifdef HWY_NATIVE_REVERSE_BITS_UI8
+#undef HWY_NATIVE_REVERSE_BITS_UI8
+#else
+#define HWY_NATIVE_REVERSE_BITS_UI8
+#endif
+
+namespace detail {
+
+template <int kShiftAmt, int kShrResultMask, class V,
+          HWY_IF_V_SIZE_GT_D(DFromV<V>, ((HWY_TARGET >= HWY_AVX3 &&
```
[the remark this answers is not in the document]
johnplatts: The real reason why a different threshold is used for the UI8ReverseBitsStep on WASM/SSE2/SSSE3/SSE4/AVX2/AVX3 vs. other targets is that 8-bit shifts are emulated using 16-bit shifts on WASM/SSE2/SSSE3/SSE4/AVX2/AVX3, and we need at least 2 int8_t/uint8_t lanes to be able to BitCast to an uint16_t vector and back to an uint8_t vector.

We want to also avoid the masking out that is done by the 8-bit ShiftLeft/ShiftRight operations on WASM/SSE2/SSSE3/SSE4/AVX2/AVX3 as the And and AndNot operations in the UI8ReverseBitsStep implementation do the required masking out of the ShiftLeft/ShiftRight result.

The UI8ReverseBitsStep operation needs to be special-cased for 1-lane int8_t/uint8_t vectors on WASM/SSE2/SSSE3/SSE4/AVX2/AVX3 as we need to wrap ```v.raw``` in a vector type that has at least 2 lanes on WASM/SSE2/SSSE3/SSE4/AVX2/AVX3 to be able to BitCast to a uint16_t vector and back to a uint16_t vector.

The AVX3_DL/AVX3_ZEN4 targets have support for GFNI and use a GFNI-based implementation of ReverseBits for int8_t/uint8_t vectors that is implemented in x86_128-inl.h, x86_256-inl.h, and x86_512-inl.h.

hwy/ops/generic_ops-inl.h
johnplatts: The ```IfVecThenElse(V mask, V yes, V no)``` implementation operation actually does a ```_mm_blendv_epi8``` operation on SSE4/AVX2, which is not what we want here as kShrResultMask is not a mask where each lane is either zero or 0xFF.

Replacing the Or+And+AndNot with IfVecThenElse would also produce incorrect results on SSE4/AVX2.

GCC 12 and later and Clang 10 and later will also optimize ```Or(And(shr_result, shr_result_mask), AndNot(shr_result_mask, shl_result))``` to a single VPTERNLOGQ instruction on AVX3 when optimizations are enabled.

hwy/ops/generic_ops-inl.h
```diff
@@ -1040,17 +1050,315 @@ HWY_API VFromD<DN> ReorderDemote2To(DN dn, V a, V b) {
...
+#else
+#define HWY_NATIVE_LEADING_ZERO_COUNT
+#endif
+
+namespace detail {
+
+template <class D, HWY_IF_U32_D(D)>
+HWY_INLINE VFromD<D> UIntToF32BiasedExp(D d, VFromD<D> v) {
+  const RebindToFloat<decltype(d)> df;
+#if HWY_TARGET > HWY_AVX3 && HWY_TARGET <= HWY_SSE2
```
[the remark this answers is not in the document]
jan-wassenberg: Wow, that is nifty. Thank you for explaining and adding the comment!

hwy/ops/generic_ops-inl.h
jan-wassenberg: I see, thanks. Great to hear that recent compilers already get the ternlog (which is also the goal of our AVX3 IfVecThenElse).

jan-wassenberg (approved):
Nice, thanks for making the changes!

hwy/ops/generic_ops-inl.h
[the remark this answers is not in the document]
jan-wassenberg: Got it, that's a great comment. I will add this to the code after we get this merged.

jan-wassenberg:
@johnplatts the internal CI reports a memory sanitizer issue, it claims uninitialized memory is used. Here's the cleaned up log:
```
MemorySanitizer: use-of-uninitialized-value hwy/ops/x86_512-inl.h:3429:29 in DemoteTo<Simd<unsigned char, 16UL, 0>
    #0 in DemoteTo<Simd<unsigned char, 16UL, 0>> hwy/ops/x86_512-inl.h:3429:29
    #1 in Lzcnt32ForU8OrU16<Vec128<unsigned char, 16UL>> hwy/ops/x86_512-inl.h:5128:10
    #2 in LeadingZeroCount<Vec128<signed char, 16UL>> hwy/ops/x86_512-inl.h:5182:26
    #3 in void TestLeadingZeroCount::operator()<signed char, Simd<signed char, 16ul, 0> hwy/tests/logical_test.cc:298:7
```
Unfortunately we did not run in track-origins mode so we do not know where the memory came from. I am fixing this for next time.

From a quick look, we're on the 128-bit int8 path, which promotes to 32-bit and then demotes. Seems straightforward and I do not see any issue.

Are you able to reproduce the issue with msan? (-fsanitize=memory compiler flag)

johnplatts:
> Are you able to reproduce the issue with msan? (-fsanitize=memory compiler flag)

I compiled the code with Clang 15.0.6 with a MSan-instrumented version of libc++ 15.0.7 on Ubuntu 22.04, and all of the tests passed with no MSan errors when compiled with Clang 15.0.6 a MSan-instrumented version of libc++ 15.0.7 on Ubuntu 22.04.

jan-wassenberg:
Thanks @johnplatts for running the test! It does sometimes happen that msan has false positives with certain compiler versions.
I am updating our CI so it will hopefully be possible to skip that test. It is uncertain whether that will succeed though, especially before I am on Easter break tomorrow. Would you mind adding HWY_ATTR_NO_MSAN before the test function (and its return type) that is failing, and then push again? I think that would be TestLeadingZeroCount's operator() on line 251.

`HWY_ATTR_NO_MSAN HWY_NOINLINE void operator() ..`.

Hopefully this would be enough to merge this pull request for now. I'm happy to experiment with our internal compiler when back next week and see if there is another workaround.

jan-wassenberg (approved):
Thanks for adding the NOMSAN :)

PR #1284, opened 2023-04-06
Add github pages support for highway

ax7e:
Add github pages support for hwy. 
@jan-wassenberg :
You need to change the following settings:
1. 
截屏2023-04-06 08 34 28
2.
截屏2023-04-06 08 42 11


.DS_Store
jan-wassenberg, line 1: Let's remove this file? (Hidden, created by OS X) Also for docs/.

docs/conf.py
```diff
@@ -0,0 +1,160 @@
...
+html_context['downloads'] = list()
+ 
+html_context['downloads'].append( ('epub', '/' +REPO_NAME+ '/' +current_language+ '/' +current_version+ '/' +REPO_NAME+ '_' +current_language+ '_' +current_version+ '.epub') )
+ 
+##########################
+# "EDIT ON GITHUB" LINKS #
+##########################
+ 
+html_context['display_github'] = True
+html_context['github_user'] = 'eric-lyun'
```
jan-wassenberg, line 157: What user should we have here?

jan-wassenberg (changes requested):
Nice, thanks for adding this! Let's make some small changes before merging:

.DS_Store
ax7e: Removed

docs/conf.py
ax7e: This corresponds to the "Edit on GitHub" buttons on the website. I am in the process of fixing it.

docs/conf.py
```diff
@@ -0,0 +1,160 @@
...
+import os
+import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'HIGHWAY'
+copyright = '2020, BSD-3-Clause License'
+author = 'Massimo DiPierro'
```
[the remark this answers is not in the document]
ax7e: I simply deleted it, it refers to the script only

PR #1285, opened 2023-04-06
Fixes for Mul compilation errors on PPC8/PPC9/PPC10

johnplatts:


jan-wassenberg (approved):
Thank you :) Unfortunately our ppc test setup currently requires the open-source version and we only run that before releases.
You still prefer to use the latest git version instead of releases, right?

johnplatts:
> Thank you :) Unfortunately our ppc test setup currently requires the open-source version and we only run that before releases. You still prefer to use the latest git version instead of releases, right?

Yes

jan-wassenberg:
Got it. When back next week I'll see if there is some way we can include a ppc test in the local/workstation test infra, which at least runs more frequently than the releases.

PR #1287, opened 2023-04-08
Improvements to ConvertTo routines on SSE2/SSSE3/SSE4/AVX2

johnplatts:
Made the following changes to the f64->i64 ConvertTo on SSE2/SSSE3/SSE4/AVX2:
- Used biased_exp instead of exp to compute shift_mnt and shift_int in the f64->i64 conversion routine used on SSE2/SSSE3/SSE4/AVX2 as using biased_exp allows the use of the more efficient uint16_t SaturatedSub operation
- Used the uint16_t SaturatedSub operation to compute shift_mnt and shift_int in the f64->i64 conversion routine used on SSE2/SSSE3/SSE4/AVX2
- Used DupEven followed by an int32_t comparison to compute the in_range mask on SSE2/SSSE3
- Computed shift_mnt and shift_int based on the biased exponent of 2^52
- Replaced int64_t right shifts with uint64_t right shifts as uint64_t right shifts are more efficient than int64_t right shifts on SSE2/SSSE3/SSE4/AVX2
- Replaced int52 (which is the mantissa with the implicit 1-bit as a 52-bit or smaller quantity) with int53 (which is the mantissa with the implicit 1-bit as a 53-bit or smaller quantity), which removes the need for the restored step

Moved the uint64_to_double128_fast lambda in the u64->f64 conversion routine used on SSE2/SSSE3/SSE4/AVX2 to a separate detail::U64ToF64VecFast function.

Also removed the duplicate u32->f32, i64->f64, u64->f64, and f64->i64 implementations for AVX2 from the x86_256-inl.h header and used the implementations of the u32->f32, i64->f64, u64->f64, and f64->i64 conversions in x86_128-inl.h for 256-vector types on AVX2.

Also separated out the AVX3 u32->f32, i64->f64, u64->f64, and f64->i64 conversion routines from the SSE2/SSSE3/SSE4/AVX2 u32->f32, i64->f64, u64->f64, and f64->i64 routines to allow the u32->f32, i64->f64, u64->f64, and f64->i64 ConvertTo routines in x86_128-inl.h to be used for 256-bit vectors on AVX2.

### DIFF
--- a/hwy/ops/x86_128-inl.h
+++ b/hwy/ops/x86_128-inl.h
@@ -6485,12 +6485,24 @@ HWY_API VFromD<D> ConvertTo(D /* tag */, VFromD<Rebind<int32_t, D>> v) {
   return VFromD<D>{_mm_cvtepi32_ps(v.raw)};
 }
 
-template <class D, HWY_IF_V_SIZE_LE_D(D, 16), HWY_IF_F32_D(D)>
-HWY_API VFromD<D> ConvertTo(D df, VFromD<Rebind<uint32_t, D>> v) {
 #if HWY_TARGET <= HWY_AVX3
-  (void)df;
+template <class D, HWY_IF_V_SIZE_LE_D(D, 16), HWY_IF_F32_D(D)>
+HWY_API VFromD<D> ConvertTo(D /*df*/, VFromD<Rebind<uint32_t, D>> v) {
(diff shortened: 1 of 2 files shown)


jan-wassenberg:
Nice improvement, thank you! It's great we are sharing code between targets.

Issue #751, opened 2022-06-10
IndicesFromVec, SetTableIndices and TableLookupLanes {u,i}{16} support 


michaeljclark:
IndicesFromVec, SetTableIndices and TableLookupLanes current only support {u,i}{32,64} for integers. I'm trying to figure out what is required to support for {u,i}{16} on x86. I'm bacoming familiar with the user side for highway but haven't implemented anything inside of highway itself yet so it might be a bit out of my depth until I become familiar with the highway template system.

It seems it should be possible to support 16-bit table lookups on x86 with:
- `PSHUFB` (SSSE3)
- `VPSHUFB` (AVX,AVX2)
- `VPERMW` (AVX2, AVX-512)

Although one would need to translate tables for use with `PSHUFB`. `VPERMW` already uses 16-bit tables. is my assumption correct that the difficulty is translating word indices to bytes indices so that they can be used with `PSHUFB` on 128-bit x86? e.g.

- { `0000 0001 0002 0003 0004 0005 0006 0007` } to { `0001 0203 0405 0607 0809 0A0B 0C0D 0E0F` }

For `IndicesFromVec` one needs code to replicate odd-to-even lanes and add 1 to odd lanes. is there anything required for SetTableIndices? can any of this be done as constant expressions so that transformed constants are stored in the text?

Tangentially :- for other architectures, could the word to byte index translation be implemented generically? I don't know enough about table lookups on the other vector architectures but I know they all have `TableLookupBytes` and that approach might work for some of them as an intermediate fix? although that won't work for AVX-512 where `TableLookupBytes` indices are 4-bit quadrant local. so I think AVX2 and AVX-512 really need to use `VPERMW` to get indices that can access all lanes.

michaeljclark:
noting that I made a couple of factual mistakes about shuffle and permute on x86. `VPERMW` does not exist for AVX2 or AVX512B rather it is part of AVX512VL. also`VPSHUFB` would also have the 4-bit lane limitation for 256-bit vectors...

jan-wassenberg:
Hi @michaeljclark, great to hear you're interested in looking into this. It would indeed be nice to extend those to 16-bit lanes and I'm happy to help.
Have you seen https://github.com/google/highway/blob/master/g3doc/impl_details.md ? That has some pointers on how to add new ops (in this case, the functions already exist and it's more a case of adding extra overloads for the 16-bit types).

[...]

Issue #982, opened 2022-09-14
Add support for TwoTablesLookupBytes


johnplatts:
Some SIMD instruction sets such as NEON, Altivec, and AVX-512VBMI have 2-table lookup instructions.

TwoTablesLookupBytes is more efficient than doing two TableLookupBytes operations, computing a mask, and combining the results using an IfThenElse operation on SIMD targets with 2-table lookup instructions such as NEON, Altivec, and AVX-512VBMI.

Here is how TwoTablesLookupBytes can be implemented for full 128-bit vectors for the tables on x86 targets with SSSE3 support:
```
template <typename T, typename TI, size_t NI>
HWY_API Vec128<TI, NI> TwoTablesLookupBytes(const Vec128<T> bytes1,
                                            const Vec128<T> bytes2,
                                            const Vec128<TI, NI> from) {
#if HWY_TARGET <= HWY_AVX3_DL
  return Vec128<TI, NI>{_mm_permutex2var_epi8(bytes1.raw, from.raw, bytes2.raw)};
#else
  const DFromV<decltype(from)> d;
  const Repartition<int8_t, decltype(d)> d8;
  const Repartition<uint16_t, decltype(d)> du16;
  using V8 = VFromD<decltype(d8)>;
  
  auto selMaskVect = BitCast(d8, ShiftLeft<3>(BitCast(du16, from)));
  #if HWY_TARGET == HWY_SSSE3
  selMaskVect = BroadcastSignBit(selMaskVect);
  #endif
  
  const auto selMask = MaskFromVec(selMaskVect);
  const V8 resultVect1{_mm_shuffle_epi8(bytes1.raw, from.raw)};

  #if HWY_TARGET == HWY_AVX3
  return Vec128<TI, NI>{_mm_mask_shuffle_epi8(resultVect1.raw, selMask.raw, bytes2.raw, from.raw)};
  #else
  const V8 resultVect2{_mm_shuffle_epi8(bytes2.raw, from.raw)};
  return BitCast(d, IfThenElse(selMask, resultVect2, resultVect1));
  #endif
#endif
}
```

Here is how TwoTablesLookupBytes can be implemented for full 128-bit vectors for the tables on ARM targets with NEON support:
```
template <typename T, typename TI>
HWY_API Vec128<TI> TwoTablesLookupBytes(const Vec128<T> bytes1,
                                        const Vec128<T> bytes2,
                                        const Vec128<TI> from) {
  const Full128<TI> d;
[...]

PR #1303, opened 2023-04-15
Added TwoTablesLookupLanes and I8/U8/I16/U16 TableLookupLanes

johnplatts:
Added the TwoTablesLookupLanes operation, which is a 2-table lookup from two tables.

Overloads of TwoTablesLookupLanes with the D parameter were also added since it is possible for ```Lanes(D())``` to be less than ```Lanes(DFromV<VFromD<D>>())``` on RVV/SVE.

Modified debug-time checks in IndicesFromVec to allow for indices up to ```2 * Lanes(d) - 1``` to allow indices greater than ```Lanes(d)``` to be used in the TwoTablesLookupLanes operation.

Also added TableLookupLanes, IndicesFromVec, and SetTableIndices for I8/U8/I16/U16 vectors.

Resolves issues #751 and #982.


hwy/ops/rvv-inl.h
```diff
@@ -2862,6 +2866,40 @@ HWY_API VI TableLookupBytesOr0(const VT vt, const VI idx) {
   return BitCast(di, IfThenZeroElse(detail::LtS(idx8, 0), lookup));
 }
 
+// ------------------------------ TwoTablesLookupLanes
+
+template <class D, HWY_IF_POW2_LE_D(D, 2)>
```
jan-wassenberg, line 2871: I think we will have to special-case 8-bit lanes because their (2-table) indices are likely to exceed 255: even with LMUL=4 that will happen for 512-bit vectors, and I know of actual silicon with 2048 bits.
The 8-bit special case can use MaskedTableLookupLanes16 (or a variant without masking).

hwy/ops/rvv-inl.h
```diff
@@ -2862,6 +2866,40 @@ HWY_API VI TableLookupBytesOr0(const VT vt, const VI idx) {
...
+  const RebindToUnsigned<decltype(d)> du;
+  using TU = TFromD<decltype(du)>;
+
+  const size_t num_of_lanes = Lanes(d);
+  const auto idx_mod = detail::AndS(idx, static_cast<TU>(num_of_lanes - 1));
+  const auto sel_a_mask = Eq(idx, idx_mod);
+
+  const auto a_lookup_result = TableLookupLanes(a, idx_mod);
+  const auto b_lookup_result = TableLookupLanes(b, idx_mod);
+  return IfThenElse(sel_a_mask, a_lookup_result, b_lookup_result);
```
jan-wassenberg, line 2893: Here we could also fuse the IfThenElse into the second lookup via MaskedTableLookupLanes.

hwy/ops/x86_256-inl.h
```diff
@@ -3130,14 +3130,60 @@ struct Indices256 {
...
+  (void)di;
+  return Indices256<T>{vec.raw};
+#else
+  const Repartition<uint8_t, decltype(di)> d8;
+  using V8 = VFromD<decltype(d8)>;
+  alignas(32) static constexpr uint8_t kByteOffsets[32] = {
+      0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1,
+      0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1};
+
+  // Broadcast each lane index to all 4 bytes of T
```
jan-wassenberg, line 3165: Minor correction: 2 bytes?

hwy/ops/x86_256-inl.h
```diff
@@ -3173,6 +3219,43 @@ HWY_API Indices256<T> SetTableIndices(D d, const TI* idx) {
...
+#if HWY_TARGET <= HWY_AVX3_DL
+  return Vec256<T>{_mm256_permutexvar_epi8(idx.raw, v.raw)};
+#else
+  const Vec256<T> idx_vec{idx.raw};
+  const DFromV<decltype(v)> d;
+  const Repartition<uint16_t, decltype(d)> du16;
+  const auto sel_hi_mask =
+      MaskFromVec(BitCast(d, ShiftLeft<3>(BitCast(du16, idx_vec))));
+
+  const Vec256<T> a{_mm256_permute2x128_si256(v.raw, v.raw, 0x00)};
```
jan-wassenberg, line 3233: Consider using ConcatLowerLower here - it's faster on Zen1.

jan-wassenberg (commented):
Very nice, great to extend the previous TableLookupLanes to all types. And thanks for adding the TwoTables :)

hwy/ops/arm_sve-inl.h
```diff
@@ -2220,6 +2224,28 @@ HWY_API VFromD<RebindToUnsigned<D>> SetTableIndices(D d, const TI* idx) {
...
+  const RebindToUnsigned<decltype(d)> du;
+  using TU = TFromD<decltype(du)>;
+
+  const size_t num_of_lanes = Lanes(d);
+  const auto idx_mod = detail::AndN(idx, static_cast<TU>(num_of_lanes - 1));
+  const auto sel_a_mask = Eq(idx, idx_mod);
+
+  const auto a_lookup_result = TableLookupLanes(a, idx_mod);
+  const auto b_lookup_result = TableLookupLanes(b, idx_mod);
+  return IfThenElse(sel_a_mask, a_lookup_result, b_lookup_result);
```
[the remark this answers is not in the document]
johnplatts: I am aware that SVE2 has svtbl2, but using svtbl2 is not correct for the case where `Lanes(d) < detail::AllHardwareLanes<TFromD<decltype(d)>>()`.

hwy/ops/x86_256-inl.h
johnplatts: Yes

hwy/ops/arm_sve-inl.h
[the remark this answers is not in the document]
jan-wassenberg: Agreed. This happens elsewhere as well. If we check `HWY_SVE_IS_POW2` (now defaults to 1, not sure 0 is still functional) and detail::IsFull, it should be fine to use. I'm happy to take care of this after merging the PR.

jan-wassenberg (approved):

jan-wassenberg:
@johnplatts CI reports a build issue (only in debug builds):
```
hwy/ops/arm_neon-inl.h:4224: error: use of undeclared identifier 'N'
  HWY_DASSERT(AllTrue(du, Lt(vec, Set(du, static_cast<TU>(N * 2)))));
                                                          ^
hwy/ops/arm_neon-inl.h:4240: error: use of undeclared identifier 'N'
  HWY_DASSERT(AllTrue(du, Lt(vec, Set(du, static_cast<TU>(N * 2)))));
```
Would you mind fixing it? Otherwise I can also patch in your pull request and update it.

I'm thinking how we can help with the testing - would it be useful to open-source our internal python script that can run any/all configs? Note that that wouldn't have caught this either unless we explicitly ask for --dbg mode.

jan-wassenberg (approved):
Thanks for updating, looks good!

PR #1304, opened 2023-04-15
Added endian detection macros

johnplatts:
HWY_IS_BIG_ENDIAN and HWY_IS_LITTLE_ENDIAN macros are added to facilitate writing portable code as the HWY_SCALAR/HWY_EMU128/HWY_PPC8/HWY_PPC9/HWY_PPC10 targets support both big-endian and little-endian targets.

The `__BYTE_ORDER__`, `__ORDER_LITTLE_ENDIAN__`, and `__ORDER_BIG_ENDIAN__` macros have also been around since GCC 4.6 and later and Clang 3.2 and later, and the current trunk of Google Highway will fail to compile with GCC 4.7 or earlier or Clang 3.8 or earlier (even with C++11 or later mode enabled).

Detecting byte order is simpler with MSVC as the XBox 360 is the only big-endian target that was supported by MSVC (with the cl compiler included in the XBox 360 SDK) whereas all other targets supported by MSVC (including x86/x86_64/ARM/AArch64) are little-endian.

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -1477,6 +1477,12 @@ The above were previously known as `HWY_CAP_INTEGER64`, `HWY_CAP_FLOAT16`, and
     implemented as `Add(Mul(f, m), a)`. Checking this can be useful for
     increasing the tolerance of expected results (around 1E-5 or 1E-6).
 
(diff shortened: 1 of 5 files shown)


jan-wassenberg (approved):
A nice simplification, thanks! 👍

PR #1308, opened 2023-04-17
Added FindLastTrue and FindKnownLastTrue operations

johnplatts:
Added the FindLastTrue and FindKnownLastTrue operations.

Also fixed the Reverse operation for large int8_t/uint8_t vectors on HWY_RVV as the HWY_RVV implementation of the FindLastTrue and FindKnownLastTrue operations uses the Reverse operation.


jan-wassenberg (dismissed):
Very nice, thanks for adding!

PR #1319, opened 2023-04-20
Compilation fixes for NEON

ag01:


jan-wassenberg (approved):
Thank you! Out of curiosity, which compiler required this?
If it's GCC, we currently only test that one before releases, but I might be able to add that to our internal manual test script if it would be helpful.

ag01:
> Thank you! Out of curiosity, which compiler required this? If it's GCC, we currently only test that one before releases, but I might be able to add that to our internal manual test script if it would be helpful.

This was tested with aarch64-linux-gnu-gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0

I think it would be very useful for us automating testing of NEON with GCC :)

jan-wassenberg:
Got it, thanks, added to TODO :)

PR #1333, opened 2023-04-24
 Added ResizeBitCast and ZeroExtendResizeBitCast operations

johnplatts:
Added the ResizeBitCast and ZeroExtendResizeBitCast operations, which allows a vector to be bit-casted to a smaller or larger vector type.

The ResizeBitCast operation is also usually more efficient than doing a BitCast followed by 2 or more LowerHalf operations in the case where `Lanes(D()) * sizeof(TFromD<D>) <= (Lanes(DFromV<V>()) * sizeof(TFromV<V>)) / 4`.

The ResizeBitCast operation is also usually more efficient than a BitCast operation followed by a ZeroExtendVector operation in the case where `Lanes(D()) * sizeof(TFromD<D>) > Lanes(DFromV<V>()) * sizeof(TFromV<V>)`.

In the case of a ResizeBitCast from a smaller vector type to a larger vector type, the contents of any bytes of the result vector past the past the first `Lanes(DFromV<V>()) * sizeof(TFromV<V>)` bytes of the result vector is unspecified, which avoids unnecessary overhead in the case where you do not care about the contents of any bytes past the first `Lanes(DFromV<V>()) * sizeof(TFromV<V>)`  bytes.

The ZeroExtendResizeBitCast operation is equivalent to a ResizeBitCast operation followed by a zero-out of the upper lanes of the result in the case where  `Lanes(D()) * sizeof(TFromD<D>) > Lanes(DFromV<V>()) * sizeof(TFromV<V>)`.

A ZeroExtendResizeBitCast(DTo, DFrom, VFromD<DFrom>) overload is also added since it is possible for `Lanes(DFromV<VFromD<DFrom>>())` to be greater than `Lanes(DFrom())` on RVV/SVE.

The ResizeBitCast operation also avoids unnecessary zero-out on many targets, including SSE2/SSSE3/SSE4/AVX2/AVX3/NEON/SVE/PPC8/PPC9/PPC10/WASM.

There are also use cases for ResizeBitCast in the implementations of StoreInterleaved3, StoreInterleaved4, and UI8ReverseBitsStep in hwy/ops/generic_ops.h.

Code such as `const VFromD<decltype(d_full)> v0{part0.raw};` can be replaced with `const auto v0 = ResizeBitCast(d_full, part0);` with the addition of the ResizeBitCast operation.

### DIFF
--- a/hwy/ops/emu128-inl.h
+++ b/hwy/ops/emu128-inl.h
@@ -109,6 +109,37 @@ HWY_API VFromD<D> BitCast(D /* tag */, VFrom v) {
   return to;
 }
 
+// ------------------------------ ResizeBitCast
+
+template <class D, class VFrom>
+HWY_API VFromD<D> ResizeBitCast(D d, VFrom v) {
+  using DFrom = DFromV<VFrom>;
+  using TFrom = TFromD<DFrom>;
+  using TTo = TFromD<D>;
+
+  constexpr size_t kFromByteLen = sizeof(TFrom) * HWY_MAX_LANES_D(DFrom);
+  constexpr size_t kToByteLen = sizeof(TTo) * HWY_MAX_LANES_D(D);
+  constexpr size_t kCopyByteLen = HWY_MIN(kFromByteLen, kToByteLen);
+
+  VFromD<D> to = Zero(d);
+  CopyBytes<kCopyByteLen>(&v, &to);
+  return to;
+}
+
+namespace detail {
+
+// ResizeBitCast on the HWY_EMU128 target has zero-extending semantics if
+// VFromD<DTo> is a larger vector than FromV
+template <class FromSizeTag, class ToSizeTag, class DTo, class DFrom>
+HWY_INLINE VFromD<DTo> ZeroExtendResizeBitCast(FromSizeTag /* from_size_tag */,
+                                               ToSizeTag /* to_size_tag */,
+                                               DTo d_to, DFrom /* d_from */,
+                                               VFromD<DFrom> v) {
+  return ResizeBitCast(d_to, v);
+}
+
+}  // namespace detail
+
 // ------------------------------ Set
 template <class D, typename T2>
 HWY_API VFromD<D> Set(D d, const T2 t) {
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -91,6 +91,86 @@ HWY_API Vec<D> Inf(D d) {
   return BitCast(d, Set(du, max_x2 >> 1));
 }
 
+// ------------------------------ ZeroExtendResizeBitCast
+
+// The implementation of detail::ZeroExtendResizeBitCast for the HWY_EMU128
+// target is in emu128-inl.h, and the implementation of
+// detail::ZeroExtendResizeBitCast for the HWY_SCALAR target is in scalar-inl.h
+#if HWY_TARGET != HWY_EMU128 && HWY_TARGET != HWY_SCALAR
+namespace detail {
+
+#if HWY_HAVE_SCALABLE
+template <size_t kFromVectSize, size_t kToVectSize, class DTo, class DFrom>
+HWY_INLINE VFromD<DTo> ZeroExtendResizeBitCast(
+    hwy::SizeTag<kFromVectSize> /* from_size_tag */,
+    hwy::SizeTag<kToVectSize> /* to_size_tag */, DTo d_to, DFrom d_from,
+    VFromD<DFrom> v) {
+  using TFrom = TFromD<DFrom>;
+  using TTo = TFromD<DTo>;
+  using TResize = UnsignedFromSize<HWY_MIN(sizeof(TFrom), sizeof(TTo))>;
+
+  const Repartition<TResize, decltype(d_from)> d_resize_from;
+  const Repartition<TResize, decltype(d_to)> d_resize_to;
+  return BitCast(d_to, IfThenElseZero(FirstN(d_resize_to, Lanes(d_resize_from)),
+                                      ResizeBitCast(d_resize_to, v)));
+}
+#else   // target that uses fixed-size vectors
+// Truncating or same-size resizing cast: same as ResizeBitCast
+template <size_t kFromVectSize, size_t kToVectSize, class DTo, class DFrom,
+          HWY_IF_LANES_LE(kToVectSize, kFromVectSize)>
+HWY_INLINE VFromD<DTo> ZeroExtendResizeBitCast(
+    hwy::SizeTag<kFromVectSize> /* from_size_tag */,
+    hwy::SizeTag<kToVectSize> /* to_size_tag */, DTo d_to, DFrom /*d_from*/,
+    VFromD<DFrom> v) {
+  return ResizeBitCast(d_to, v);
+}
+
+// Resizing cast to vector that has twice the number of lanes of the source
+// vector
+template <size_t kFromVectSize, size_t kToVectSize, class DTo, class DFrom,
+          HWY_IF_LANES(kToVectSize, kFromVectSize * 2)>
+HWY_INLINE VFromD<DTo> ZeroExtendResizeBitCast(
+    hwy::SizeTag<kFromVectSize> /* from_size_tag */,
+    hwy::SizeTag<kToVectSize> /* to_size_tag */, DTo d_to, DFrom d_from,
+    VFromD<DFrom> v) {
+  const Twice<decltype(d_from)> dt_from;
+  return BitCast(d_to, ZeroExtendVector(dt_from, v));
+}
+
+// Resizing cast to vector that has more than twice the number of lanes of the
+// source vector
+template <size_t kFromVectSize, size_t kToVectSize, class DTo, class DFrom,
+          HWY_IF_LANES_GT(kToVectSize, kFromVectSize * 2)>
+HWY_INLINE VFromD<DTo> ZeroExtendResizeBitCast(
+    hwy::SizeTag<kFromVectSize> /* from_size_tag */,
+    hwy::SizeTag<kToVectSize> /* to_size_tag */, DTo d_to, DFrom /*d_from*/,
+    VFromD<DFrom> v) {
+  using TFrom = TFromD<DFrom>;
+  constexpr size_t kNumOfFromLanes = kFromVectSize / sizeof(TFrom);
+  const Repartition<TFrom, decltype(d_to)> d_resize_to;
+  return BitCast(d_to, IfThenElseZero(FirstN(d_resize_to, kNumOfFromLanes),
+                                      ResizeBitCast(d_resize_to, v)));
+}
+#endif  // HWY_HAVE_SCALABLE
+
+}  // namespace detail
+#endif  // HWY_TARGET != HWY_EMU128 && HWY_TARGET != HWY_SCALAR
+
+template <class DTo, class DFrom>
+HWY_API VFromD<DTo> ZeroExtendResizeBitCast(DTo d_to, DFrom d_from,
+                                            VFromD<DFrom> v) {
+  using TFrom = TFromD<DFrom>;
+  using TTo = TFromD<DTo>;
+  return detail::ZeroExtendResizeBitCast(
+      hwy::SizeTag<HWY_MAX_LANES_D(DFrom) * sizeof(TFrom)>(),
+      hwy::SizeTag<HWY_MAX_LANES_D(DTo) * sizeof(TTo)>(), d_to, d_from, v);
+}
+
+template <class DTo, class FromV>
+HWY_API VFromD<DTo> ZeroExtendResizeBitCast(DTo d_to, FromV v) {
+  return ZeroExtendResizeBitCast(d_to, DFromV<decltype(v)>(), v);
+}
+
 // ------------------------------ SafeFillN
 
 template <class D, typename T = TFromD<D>>
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -159,16 +159,9 @@ HWY_INLINE VFromD<DTo> ZeroExtendResizeBitCast(
 template <class DTo, class DFrom>
 HWY_API VFromD<DTo> ZeroExtendResizeBitCast(DTo d_to, DFrom d_from,
                                             VFromD<DFrom> v) {
-  using TFrom = TFromD<DFrom>;
-  using TTo = TFromD<DTo>;
-  return detail::ZeroExtendResizeBitCast(
-      hwy::SizeTag<HWY_MAX_LANES_D(DFrom) * sizeof(TFrom)>(),
-      hwy::SizeTag<HWY_MAX_LANES_D(DTo) * sizeof(TTo)>(), d_to, d_from, v);
-}
-
-template <class DTo, class FromV>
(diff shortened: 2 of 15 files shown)


hwy/ops/emu128-inl.h
```diff
@@ -109,6 +109,37 @@ HWY_API VFromD<D> BitCast(D /* tag */, VFrom v) {
...
 
+// ------------------------------ ResizeBitCast
+
+template <class D, class VFrom>
+HWY_API VFromD<D> ResizeBitCast(D d, VFrom v) {
+  using DFrom = DFromV<VFrom>;
+  using TFrom = TFromD<DFrom>;
+  using TTo = TFromD<D>;
+
+  constexpr size_t kFromByteLen = sizeof(TFrom) * HWY_MAX_LANES_D(DFrom);
```
jan-wassenberg, line 120: Note that Simd<> has a MaxBytes() member function which could simplify these. Do you want to update the code? (It's also fine as-is, just more verbose.)

hwy/ops/generic_ops-inl.h
```diff
@@ -91,6 +91,86 @@ HWY_API Vec<D> Inf(D d) {
   return BitCast(d, Set(du, max_x2 >> 1));
 }
 
+// ------------------------------ ZeroExtendResizeBitCast
+
+// The implementation of detail::ZeroExtendResizeBitCast for the HWY_EMU128
```
jan-wassenberg, line 96: Did we consider using the same 'include guard' technique like HWY_NATIVE_LOAD_STORE_INTERLEAVED here so that we aren't hard-coding EMU128/SCALAR? Or do you think it very unlikely that any other target would also want to implement this op themselves?

hwy/ops/generic_ops-inl.h
```diff
@@ -91,6 +91,86 @@ HWY_API Vec<D> Inf(D d) {
...
+}  // namespace detail
+#endif  // HWY_TARGET != HWY_EMU128 && HWY_TARGET != HWY_SCALAR
+
+template <class DTo, class DFrom>
+HWY_API VFromD<DTo> ZeroExtendResizeBitCast(DTo d_to, DFrom d_from,
+                                            VFromD<DFrom> v) {
+  using TFrom = TFromD<DFrom>;
+  using TTo = TFromD<DTo>;
+  return detail::ZeroExtendResizeBitCast(
+      hwy::SizeTag<HWY_MAX_LANES_D(DFrom) * sizeof(TFrom)>(),
```
jan-wassenberg, line 165: Here also we could use DFrom().MaxBytes()? (It is constexpr.)

hwy/ops/x86_256-inl.h
```diff
@@ -275,6 +275,42 @@ HWY_API Vec256<double> Undefined(D /* tag */) {
...
+          HWY_IF_V_SIZE_D(D, HWY_MAX_LANES_V(FromV) * sizeof(TFromV<FromV>))>
+HWY_API VFromD<D> ResizeBitCast(D d, FromV v) {
+  return BitCast(d, v);
+}
+
+// 32-byte vector to 16-byte vector (or 64-byte vector to 32-byte vector on
+// AVX3)
+template <class D, class FromV, HWY_IF_V_SIZE_GT_V(FromV, 16),
+          HWY_IF_V_SIZE_D(D,
+                          (HWY_MAX_LANES_V(FromV) * sizeof(TFromV<FromV>)) / 2)>
```
jan-wassenberg, line 292: Here and above we could also use DFrom<FromV>().MaxBytes() (/2)?

hwy/tests/combine_test.cc
```diff
@@ -284,6 +284,170 @@ HWY_NOINLINE void TestAllConcatOddEven() {
...
+    const auto v_full = Iota(d, 1);
+    const VFromD<decltype(dh)> expected_full_to_half = LowerHalf(dh, v_full);
+    HWY_ASSERT_VEC_EQ(dh, expected_full_to_half, ResizeBitCast(dh, v_full));
+    HWY_ASSERT_VEC_EQ(dh, expected_full_to_half,
+                      ZeroExtendResizeBitCast(dh, v_full));
+    HWY_ASSERT_VEC_EQ(dh, expected_full_to_half,
+                      ZeroExtendResizeBitCast(dh, d, v_full));
+
+    constexpr size_t kMaxLanes = MaxLanes(d);
+#if HWY_TARGET == HWY_RVV
```
jan-wassenberg, line 321: Can/should we use this same code on other targets, with just a target-dependent max actual pow2 (3 on RVV, 0 otherwise)?

hwy/tests/convert_test.cc
```diff
@@ -176,6 +176,168 @@ HWY_NOINLINE void TestAllBitCast() {
...
+
+    const FixedTag<TTo, 1> d_to;
+    TTo expected_bits = 0;
+    memcpy(&expected_bits, from_lanes.get(), num_of_bytes_to_copy);
+
+    const auto expected = Set(d_to, expected_bits);
+    const auto v_active_bits_mask = Set(d_to, active_bits_int_mask);
+
+    const auto actual_1 = And(v_active_bits_mask, ResizeBitCast(d_to, v));
+#if HWY_HAVE_SCALABLE || HWY_TARGET == HWY_SVE_256 || HWY_TARGET == HWY_SVE2_128
```
jan-wassenberg, line 211: Can we avoid this (and the previous) `#if` somehow?

jan-wassenberg (commented):
Great idea! This looks useful for avoiding the Vec128<T>{partial.raw} workaround. Some minor suggestions:

hwy/tests/convert_test.cc
johnplatts: The reason for the #if statement here is that the `ZeroExtendResizeBitCast(d_to, v)` overload treats `v` as a full vector that has `Lanes(DFromV<decltype(v)>())` lanes on RVV/SVE targets and as it is possible for `Lanes(d_from)` to be less than `Lanes(DFromV<decltype(v)>())` on RVV/SVE targets.

The `ZeroExtendResizeBitCast(d_to, d_from, v)` overload always treats `v` as a vector that has `Lanes(d_from)` lanes, even in the case where `Lanes(DFromV<decltype(v)>()) > Lanes(d_from)` is true (which is possible on RVV/SVE).

hwy/ops/generic_ops-inl.h
johnplatts: > Did we consider using the same 'include guard' technique like HWY_NATIVE_LOAD_STORE_INTERLEAVED here so that we aren't hard-coding EMU128/SCALAR? Or do you think it very unlikely that any other target would also want to implement this op themselves?

It is very likely that targets other than HWY_EMU128/HWY_SCALAR will use the implementations of detail::ZeroExtendResizeBitCast in hwy/ops/generic_ops-inl.h for at least some of the ZeroExtendResizeBitCast cases.

The detail::ZeroExtendResizeBitCast function has hwy::SizeTag parameters for the from vector length (in bytes) and the to vector length (in bytes) to allow the detail::ZeroExtendResizeBitCast to be specialized for a few specific cases on the HWY_AVX2/HWY_AVX3/HWY_WASM_EMU256 targets.

While it is possible to use the detail::ZeroExtendResizeBitCast implementation in hwy/ops/generic_ops-inl.h on the HWY_EMU128 target, the implementation of detail::ZeroExtendResizeBitCast in hwy/ops/emu128-inl.h is used for all of the ZeroExtendResizeBitCast cases on the HWY_EMU128 target to avoid the unnecessary zero-out of the upper bytes as the ResizeBitCast implementation in hwy/ops/emu128_inl.h already does the required zero-out of the upper bytes of the result vector.

The detail::ZeroExtendResizeBitCast implementation in hwy/ops/generic_ops-inl.h cannot be used on the HWY_SCALAR target as all of the vector types on the HWY_SCALAR target have only 1 lane. The detail::ZeroExtendResizeBitCast implementation on the HWY_SCALAR target is in hwy/ops/scalar-inl.h, and the implementation of detail::ZeroExtendResizeBitCast in hwy/ops/scalar-inl.h can handle the case of a ZeroExtendResizeBitCast from a smaller lane type to a bigger lane type.

Here is the implementation of ZeroExtendResizeBitCast in hwy/ops/generic_ops-inl.h, shared across all targets, which delegates to detail::ZeroExtendResizeBitCast:
https://github.com/google/highway/blob/04aea95474744df7cb09fc209886b590136846d7/hwy/ops/generic_ops-inl.h#L159-L172

Here are the specialized cases of detail::ZeroExtendResizeBitCast that are implemented for the HWY_AVX2/HWY_AVX3/HWY_WASM_EMU256 targets (with the remaining cases of detail::ZeroExtendResizeBitCast on the HWY_AVX2/HWY_AVX3/HWY_WASM_EMU256 targets using the implementations of detail::ZeroExtendResizeBitCast in hwy/ops/generic_ops-inl.h):
https://github.com/google/highway/blob/04aea95474744df7cb09fc209886b590136846d7/hwy/ops/x86_256-inl.h#L2938-L2949
https://github.com/google/highway/blob/04aea95474744df7cb09fc209886b590136846d7/hwy/ops/x86_512-inl.h#L2451-L2501
https://github.com/google/highway/blob/04aea95474744df7cb09fc209886b590136846d7/hwy/ops/wasm_256-inl.h#L1147-L1159

Note that the detail::ZeroExtendResizeBitCast implementations in hwy/ops/x86_256-inl.h, hwy/ops/x86_512-inl.h, and hwy/ops/wasm_256-inl.h have at least one hwy::SizeTag parameter with a specific size so that the implementations of detail::ZeroExtendResizeBitCast in hwy/ops/x86_256-inl.h, hwy/ops/x86_512-inl.h, and hwy/ops/wasm_256-inl.h will be treated as more specialized than the detail::ZeroExtendResizeBitCast implementations in hwy/ops/generic_ops-inl.h.

It is also unlikely that targets other than HWY_EMU128/HWY_SCALAR would completely replace the implementations of detail::ZeroExtendResizeBitCast in hwy/ops/generic_ops-inl.h with target-specific implementations of detail::ZeroExtendResizeBitCast that cover all of the possible ZeroExtendResizeBitCast cases.

If a new target other than HWY_EMU128/HWY_SCALAR/HWY_AVX2/HWY_AVX3/HWY_WASM_EMU256 needs to specialize the detail::ZeroExtendResizeBitCast function, such new targets would likely specialize only certain detail::ZeroExtendResizeBitCast cases with target-specific implementations that have at least one hwy::SizeTag parameter with a specific size and delegate to the default detail::ZeroExtendResizeBitCast implementations in hwy/ops/generic_ops-inl.h for the remaining cases.

jan-wassenberg: Thanks for explaining. I understand that those two targets are special and thus we can hard-code them in the if, and that we already have an extension point for other targets to use.

hwy/tests/convert_test.cc
jan-wassenberg: I see, thanks. We have to mask out the other lanes on SVE/RVV, but not other platforms.

Then I wonder: for user clarity, and avoiding surprising changes in behaviour across platforms, should we only provide the form with 3 arguments d_to, d_from, v_from?
The downside is only providing an extra argument, right? But this often happens, for example Reverse2() has a d-arg which is mostly unused on x86 or fixed-size vector, but the op still requires that argument for consistency across platforms.

hwy/ops/x86_256-inl.h
johnplatts: I was actually getting compilation errors on MSVC when I tried to replace `(HWY_MAX_LANES_V(FromV) * sizeof(TFromV<FromV>)) / 2` with `DFromV<FromV>().MaxBytes() / 2`.

Need to use the `(HWY_MAX_LANES_V(FromV) * sizeof(TFromV<FromV>)) / 2` here as x86_256-inl.h needs to be compilable with MSVC.

hwy/tests/convert_test.cc
johnplatts: > I see, thanks. We have to mask out the other lanes on SVE/RVV, but not other platforms.
> 
> Then I wonder: for user clarity, and avoiding surprising changes in behaviour across platforms, should we only provide the form with 3 arguments d_to, d_from, v_from? The downside is only providing an extra argument, right? But this often happens, for example Reverse2() has a d-arg which is mostly unused on x86 or fixed-size vector, but the op still requires that argument for consistency across platforms.

I have removed the 2-argument version of the ZeroExtendResizeBitCast function and only kept the 3-argument version of the ZeroExtendResizeBitCast function for user clarity to avoid surprising changes in behavior across platforms.

johnplatts:
The reason why the build / test of the pull request failing is due to errors in the hwy/contrib/sort/algo-inl.h, hwy/contrib/sort/bench_parallel.cc, hwy/contrib/sort/sort_test.cc, hwy/contrib/sort/vqsort.cc, and hwy/contrib/sort/vqsort.h files, and the fixes in pull request https://github.com/google/highway/pull/1345 will fix the compilation errors with pull request #1333.

hwy/ops/x86_256-inl.h
jan-wassenberg: I see, thanks for trying it :/ MSVC does indeed struggle with SFINAE.

jan-wassenberg:
Apologies for the build errors. Thank you for sending the PR for that! I found out why the one in bench_parallel got through: we are not building it in CI. Fix is under way.

jan-wassenberg:
FYI your fixes are waiting for our internal review. Thanks for removing the 2-argument form, that's great and I hope we can merge this PR tomorrow.

jan-wassenberg:
The fix is in! Would you mind rebasing so we can see the result of Github actions on this change?

johnplatts:
> The fix is in! Would you mind rebasing so we can see the result of Github actions on this change?

While the rebased code does successfully compile with Clang 6, Clang 10, and Clang 12, the Bazel build now fails with the following error message:
```
ERROR: /home/runner/work/highway/highway/BUILD:397:16: no such package 'base': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - /home/runner/work/highway/highway/base and referenced by '//:nanobenchmark_test'
ERROR: Analysis of target '//:nanobenchmark_test' failed; build aborted: Analysis failed
```

jan-wassenberg (approved):
Thanks for rebasing :)

hwy/tests/combine_test.cc
```diff
@@ -284,6 +284,151 @@ HWY_NOINLINE void TestAllConcatOddEven() {
...
+    const VFromD<DTo> actual_1 = ResizeBitCast(d_to, v);
+    HWY_ASSERT_VEC_EQ(d_to, expected, actual_1);
+
+    const VFromD<DTo> actual_2 = ZeroExtendResizeBitCast(d_to, d_from, v);
+    HWY_ASSERT_VEC_EQ(d_to, expected, actual_2);
+  }
+#endif  // HWY_TARGET != HWY_SCALAR
+ public:
+  template <class T, class D>
+  HWY_NOINLINE void operator()(T /*unused*/, D d) {
```
jan-wassenberg, line 304: @johnplatts our CI is failing because `d` is unused #if HWY_TARGET == HWY_SCALAR. Would you mind adding (for example) a `(void)d` here and in line 392?

jan-wassenberg (changes requested):
One minor change required for CI to pass :) (unused d arg, combine_test lines 304/392)

jan-wassenberg:
This one got stuck somehow, sorry about that. Restarting import..

PR #1340, opened 2023-04-26
Added missing I8/U8 Reverse operations for PPC8/PPC9/PPC10

johnplatts:
Fixes compilation errors on PPC8/PPC9/PPC10 by adding the missing I8/U8 Reverse operations for PPC8/PPC9/PPC10


jan-wassenberg (approved):
Thanks for fixing this!

PR #1341, opened 2023-04-26
Added AESKeyGenAssist operation

johnplatts:
Added the AESKeyGenAssist operation, which can be used to accelerate the AES Key Expansion algorithm, as the AESKeyGenAssist operation for a 128-bit uint8_t vector can be carried out in a single instruction on SSE4/AVX2/AVX3/AVX3_DL targets using the x86 AES-NI AESKEYGENASSIST instruction.

There are two different approaches that can be used to implement the AESKeyGenAssist operation for the other targets.

The first approach, which is used on NEON Crypto/SVE2 Crypto (and for 256-bit/512-bit vectors on AVX3_DL) is to do a U32 DupOdd operation, followed by an AESLastRound operation, followed by a shuffle using the TableLookupBytes operation.

The second approach, which is used by EMU128/PPC8/PPC9/PPC10/NEON_WITHOUT_AES/SSE2/SSSE3/RVV is to do an AES SubBytes operation, followed by a TableLookupBytes operation, followed by a Xor operation. The second approach also avoids the extra shuffle operation that the first approach would require on the EMU128/NEON_WITHOUT_AES/SSE2/SSSE3/RVV targets.

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -1471,6 +1471,24 @@ Ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
     step of the AES Equivalent Inverse Cipher algorithm. The latency is
     independent of the input values.
 
+*   `V`: `u8` \
+    <code>V **AESKeyGenAssist**&lt;uint8_t&gt;(V v)</code>: AES key generation
+    assist operation
+
+    The AESKeyGenAssist operation is equivalent to doing the following, which
+    matches the behavior of the x86 AES-NI AESKEYGENASSIST instruction:
+    *  Applying the AES SubBytes operation to each byte of `v`.
+    *  Doing a TableLookupBytes operation on each 128-bit block of the
+       result of the `SubBytes(v)` operation with the following indices
+       (which is broadcast to each 128-bit block in the case of vectors with 32
+       or more lanes):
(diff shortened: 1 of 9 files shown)


g3doc/quick_reference.md
```diff
@@ -1471,6 +1471,24 @@ Ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
     step of the AES Equivalent Inverse Cipher algorithm. The latency is
     independent of the input values.
 
+*   `V`: `u8` \
+    <code>V **AESKeyGenAssist**&lt;uint8_t&gt;(V v)</code>: AES key generation
```
jan-wassenberg, line 1475: Let's add kRcon as the name of the uint8_t template argument?

hwy/ops/generic_ops-inl.h
```diff
@@ -1615,6 +1615,19 @@ HWY_API V AESLastRoundInv(V state, const V round_key) {
   return state;
 }
 
+template <uint8_t kRoundConstant, class V>  // u8
```
jan-wassenberg, line 1618: Consider adding `HWY_IF_U8_D(DFromV<V>)` to ensure this doesn't compile (with clear explanation) if the type is not u8?

jan-wassenberg (commented):
Nice, thanks for adding this! Two minor suggestions:

jan-wassenberg (dismissed):
Thanks for making the changes :)

jan-wassenberg (approved):

PR #1354, opened 2023-05-03
Made enhancements to x86 InsertLane operation

johnplatts:
Re-implemented InsertLane on SSE2/SSSE3/SSE4/AVX2/AVX3 to do a Set operation followed by an IfThenElse operation.


jan-wassenberg (approved):
Nice idea, thanks :) Great to avoid the store-to-load forwarding stall that the previous implementation had.

Issue #948, opened 2022-08-25
Add support for SaturatedAdd and SaturatedSub for int32_t, uint32_t, int64_t, and uint64_t vector types


johnplatts:
Some SIMD instruction sets such as NEON, SVE, and RVV have instructions for saturated addition/subtraction of int32_t, uint32_t, int64_t, and uint64_t SIMD vectors.

AltiVec also has instructions for saturated addition/subtraction of int32_t and uint32_t SIMD vectors.

Here is how int32_t, uint32_t, int64_t, and uint64_t SaturatedAdd and SaturatedSub can be implemented for x86, WASM, and HWY_EMU128 targets:
```
template <size_t N>
HWY_API Vec128<uint32_t, N> SaturatedAdd(const Vec128<uint32_t, N> a,
                                         const Vec128<uint32_t, N> b) {
  return Add(a, Min(b, Not(a)));
}
template <size_t N>
HWY_API Vec128<uint64_t, N> SaturatedAdd(const Vec128<uint64_t, N> a,
                                         const Vec128<uint64_t, N> b) {
  return Add(a, Min(b, Not(a)));
}
template <size_t N>
HWY_API Vec128<uint32_t, N> SaturatedSub(const Vec128<uint32_t, N> a,
                                         const Vec128<uint32_t, N> b) {
  return Sub(Max(a, b), b);
}
template <size_t N>
HWY_API Vec128<uint64_t, N> SaturatedSub(const Vec128<uint64_t, N> a,
                                         const Vec128<uint64_t, N> b) {
  return Sub(Max(a, b), b);
}

template <size_t N>
HWY_API Vec128<int32_t, N> SaturatedAdd(const Vec128<int32_t, N> a,
                                        const Vec128<int32_t, N> b) {
  const DFromV<decltype(a)> d;
  const RebindToUnsigned<decltype(d)> du;
  
  const Vec128<int32_t, N> result = Add(a, b);
  const Vec128<int32_t, N> overflowMaskVect =
    BroadcastSignBit(And(Xor(a, result), Xor(b, result)));
  const Mask128<int32_t, N> overflowMask =
    MaskFromVec(overflowMaskVect);
  const Vec128<int32_t, N> overflowResult = BitCast(d, Add(
    ShiftRight<1>(BitCast(du, overflowMaskVect)),
    ShiftRight<31>(BitCast(du, a))));
  
  return IfThenElse(overflowMask, overflowResult, result);
}

template <size_t N>
[...]

PR #1357, opened 2023-05-03
Added I32/U32/I64/U64 SaturatedAdd/SaturatedSub

johnplatts:
I32/U32/I64/U64 SaturatedAdd/SaturatedSub operations were added as NEON/SVE/RVV have saturated addition/subtraction instructions for I32/U32/I64/U64 vectors and as PPC8/PPC9/PPC10 have saturated addition/subtraction instructions for I32/U32 vectors.

Resolves issue #948


hwy/tests/arithmetic_test.cc
```diff
@@ -143,16 +143,24 @@ HWY_NOINLINE void TestAllSaturatingArithmetic() {
   const ForPartialVectors<TestUnsignedSaturatingArithmetic> test_unsigned;
   test_unsigned(uint8_t());
   test_unsigned(uint16_t());
+  test_unsigned(uint32_t());
+  test_unsigned(uint64_t());
```
jan-wassenberg, line 147: Now that all types are supported, we could simplify this: remove the test_* variables and replace with ForUnsignedTypes(ForPartialVectors<TestUnsignedSaturatingArithmetic>());

And for U+I below it would be ForIntegerTypes. Would you like to make this change? If not, I can also do it after merge. (Bonus: this would safely handle HWY_HAVE_INTEGER64==0, though there is currently no target that lacks u/i64.)

jan-wassenberg (commented):
Nice, thanks for adding this :) Clever trick with the masked ternarylogic.

jan-wassenberg (approved):
👍

PR #1369, opened 2023-05-05
Added BitwiseIfThenElse operation

johnplatts:
While the BitwiseIfThenElse operation is similar to the IfVecThenElse operation, the BitwiseIfThenElse operation was added to provide guarantees that `BitwiseIfThenElse(mask, yes, no)` produces the same result as `Or(And(mask, yes), AndNot(mask, no))` in the case where `mask[i]` is neither zero nor all bits set.

There are some targets where `IfVecThenElse(mask, yes, no)` does not produce the same result as `BitwiseIfThenElse(mask, yes, no)` for the case where `mask[i]` is neither zero nor all bits set, including SSE4/AVX2.

Here is an example of code where BitwiseIfThenElse and IfVecThenElse generate different results on SSE4/AVX2:
```
alignas(16) static constexpr uint8_t kVectA[16] =
  { 0xB4, 0x9C, 0xC3, 0x97, 0xBA, 0x44, 0xA3, 0xB3, 0xDF, 0x29, 0x7A, 0xA3, 0x14, 0x4D, 0xE2, 0xF6 };
alignas(16) static constexpr uint8_t kVectB[16] =
  { 0x1B, 0x7B, 0x81, 0xF9, 0xBE, 0xF4, 0x42, 0x67, 0x6E, 0xD1, 0xA9, 0xB3, 0x5E, 0x1C, 0x0E, 0x69 };
alignas(16) static constexpr uint8_t kVectC[16] =
  { 0x79, 0x37, 0xE4, 0xAD, 0x01, 0x3E, 0x5C, 0x0E, 0xCE, 0x1E, 0xF4, 0x5B, 0x77, 0x8A, 0xDA, 0x59 };

const FixedTag<uint8_t, 16> d;
const auto a = Load(d, kVectA);
const auto b = Load(d, kVectB);
const auto c = Load(d, kVectC);

const auto r1 = BitwiseIfThenElse(a, b, c);
const auto r2 = IfVecThenElse(a, b, c);
```

Here is the result of `r1` in the example code above (on all targets):
`{0x59, 0x3B, 0xA5, 0xB9, 0xBB, 0x7E, 0x5E, 0x2F, 0x4E, 0x17, 0xAC, 0xFB, 0x77, 0x8E, 0x1A, 0x69}`

Here is the result of `r2` in the example code above on the SSE4/AVX2 targets (where IfVecThenElse is implemented using a `_mm_blendv_epi8` operation):
`{0x1B, 0x7B, 0x81, 0xF9, 0xBE, 0x3E, 0x42, 0x67, 0x6E, 0x1E, 0xF4, 0xB3, 0x77, 0x8A, 0x0E, 0x69}`

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -655,6 +655,12 @@ them from 2-argument functions:
 *   <code>V **Or3**(V o1, V o2, V o3)</code>: returns `o1[i] | o2[i] | o3[i]`.
     This is less efficient than `Xor3` on some targets; use that where possible.
 *   <code>V **OrAnd**(V o, V a1, V a2)</code>: returns `o[i] | (a1[i] & a2[i])`.
+*   <code>V **BitwiseIfThenElse**(V mask, V yes, V no)</code>: returns
+    `((mask[i] & yes[i]) | (~mask[i] & no[i]))`. `BitwiseIfThenElse` is more
+    efficient than `Or(And(mask, yes), AndNot(mask, no))` on some targets.
+
+    `BitwiseIfThenElse(mask, yes, no)` is guaranteed to return the same result
+    as `Or(And(mask, yes), AndNot(mask, no))` on all targets
 
 Special functions for signed types:
 
--- a/hwy/ops/arm_neon-inl.h
+++ b/hwy/ops/arm_neon-inl.h
@@ -2131,6 +2131,19 @@ HWY_API Vec128<T, N> IfVecThenElse(Vec128<T, N> mask, Vec128<T, N> yes,
   return IfThenElse(MaskFromVec(mask), yes, no);
 }
 
+// ------------------------------ BitwiseIfThenElse
+
+#ifdef HWY_NATIVE_BITWISE_IF_THEN_ELSE
+#undef HWY_NATIVE_BITWISE_IF_THEN_ELSE
+#else
+#define HWY_NATIVE_BITWISE_IF_THEN_ELSE
+#endif
+
+template <class V>
+HWY_API V BitwiseIfThenElse(V mask, V yes, V no) {
+  return IfVecThenElse(mask, yes, no);
+}
+
 // ------------------------------ Operator overloads (internal-only if float)
 
 template <typename T, size_t N>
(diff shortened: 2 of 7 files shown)


jan-wassenberg (approved):
Nice, seems useful. I am curious about your use-case of this vbsl instruction (as Arm calls it)?

johnplatts:
> Nice, seems useful. I am curious about your use-case of this vbsl instruction (as Arm calls it)?

One use case for the BitwiseIfThenElse operation is in the detail::UI8ReverseBitsStep implementation in hwy/ops/generic_ops-inl.h.

Another use case for BitwiseIfThenElse is where you want to do a `mask ? a : b` operation on unsigned integer vectors that are treated as bit sets (similar to C++ std::bitset or Java java.util.BitSet).

`Or(And(shr_result, shr_result_mask), AndNot(shr_result_mask, shl_result));` in detail::UI8ReverseBitsStep in hwy/ops/generic_ops-inl.h can be replaced by `BitwiseIfThenElse(shr_result_mask, shr_result, shl_result)` with the addition of the BitwiseIfThenElse method.

Note that it would be incorrect to use `IfVecTheElse(shr_result_mask, shr_result, shl_result)` in the detail::UI8ReverseBitsStep implementation as `IfVecTheElse(shr_result_mask, shr_result, shl_result)` will generate incorrect results on some targets, including SSE4/AVX2.

There are also a few other targets other than NEON/SVE that can carry out the BitwiseIfThenElse operation in a single instruction, including PPC8/PPC9/PPC10/AVX3.

Recent versions of GCC/Clang will also optimize the `Or(And(shr_result, shr_result_mask), AndNot(shr_result_mask, shl_result))` expression to a single instruction on NEON/AVX3 when optimizations are enabled.

GCC/Clang (at least up to GCC 12.2.0 and Clang 16.0.0) fail to optimize `Or(And(shr_result, shr_result_mask), AndNot(shr_result_mask, shl_result))` down to a single vbsl instruction on SVE.

GCC (at least up to GCC 12.2.0) fails to optimize `Or(And(shr_result, shr_result_mask), AndNot(shr_result_mask, shl_result))` down to a single xxsel or vsel instruction on PPC8/PPC9/PPC10.

jan-wassenberg:
Thanks, makes sense to add this op given that compilers are not automatically generating it :)

PR #1384, opened 2023-05-10
Fix compiler warnings in hwy/contrib/sort/vqsort-inl.h

johnplatts:
Fix compiler warnings in hwy/contrib/sort/vqsort-inl.h


jan-wassenberg:
Thank you! These will also be fixed by https://github.com/google/highway/pull/1382.
Except for the loop unroll, thanks for pointing that out. We recently added HWY_UNROLL and I should have used that. I will change that PR.

And the conversion warning is not yet fixed. Let's see how quickly 1382 goes in. How about we wait a bit for that, rebase, and then merge your ptrdiff_t fix?

jan-wassenberg:
Looks like 1382 won't go in today. Let's merge your change already :)

jan-wassenberg (approved):

PR #1386, opened 2023-05-10
Fix clang compilation error with PPC9/PPC10 TrailingZeroCount

johnplatts:
Replaced vec_vctz with vec_cnttz in the PPC9/PPC10 TrailingZeroCount implementation as vec_cnttz is available on GCC 7 and later and Clang 4.0 and later.


jan-wassenberg (approved):
👍

Issue #915, opened 2022-08-09
Missing data type promotion


hellishfire:
According to the manual of PromoteTo:

V,D: (u8,u16), (u16,u32), (u8,u32), (u32,u64), (u8,i16),
(u8,i32), (u16,i32), (i8,i16), (i8,i32), (i16,i32), (i32,i64) Vec<D> PromoteTo(D, V part): returns part[i] zero- or sign-extended to the integer type MakeWide<T>.

Why are some data type promotion missing? e.g, int8->int64, int16->int64

jan-wassenberg:
The initial design philosophy was to only provide ops which are efficient on all platforms. Since then, we have evolved to a more inclusive "any anything useful which doesn't have major performance cliffs, and applications cannot do any better themselves using other ops".
The short answer is: no one has asked for them yet :) Do you have a use case that requires them?

hellishfire:
I'm writing codes which involve calculating the sum of an input array (of InputType), and return the sum (of OutputType, where InputType is no wider than OutputType) . Something like the following snippet:

template<typename InputType, typename OutputType>
OutputType sum(const InputType* input, int num) {
  static_assert(std::is_integral_v<InputType> && std::is_integral_v<OutputType>);
  static_assert(sizeof(OutputType) >= sizeof(InputType));
  static_assert(std::is_signed_v<InputType> == std::is_signed_v<OutputType>);

  const ScalableTag<OutputType> d_output;
  const Rebind<InputType, decltype(d_output)> d_input;
  const size_t N = Lanes(d_output);
  auto vec_sum = Zero(d_output);
  int i = 0;
  for (; i + N <= num; i += N) {
    auto vec = LoadU(d_input, input);
    vec_sum = Add(vec_sum, PromoteTo(d_output, vec));
    input += N;
  }
  vec_sum = SumOfLanes(d_output, vec_sum);
  OutputType ret = GetLane(vec_sum);
  // ... handle trailing elements
  return ret;
}

Both InputType and OutputType range from int8 to int64. SumOfLanes only works with int32/64, but I can do without it. However, PromoteTo doesn't handle every type promotion combinations and fails to compile.

hellishfire:
Thanks for taking note of this issue. Unfortunately I've been on a tight schedule recently:(
Can't say I'm familiar with simd instructions, but I did skim through the x86 implementation of PromoteTo. Shouldn't be too hard to expand it, I guess, though some workaround is needed for sse3.

[...]

PR #1387, opened 2023-05-10
Added U8/U16->U64, U8/U16/U32->I64, and I8/I16->I64 PromoteTo operations

johnplatts:
U8/U16->U64, U8/U16/U32->I64, and I8/I16->I64 PromoteTo operations were added for all targets to ensure consistency across all targets.

SSE4/AVX2/AVX3/RVV can also do I8->I64, I16->I64, U8->U64, and U16->U64 PromoteTo operations using a single instruction.

U8->U64 PromoteTo was also already implemented on the SVE/SCALAR/EMU128 targets.

Also fixed ForPromoteVectors implementation in hwy/tests/test_util-inl.h.


jan-wassenberg (approved):
Very nice, thanks for adding the remaining types (#915)! I will update the documentation after this goes in.

PR #1389, opened 2023-05-11
Made enhancements to PPC10 operations

johnplatts:
Added the HWY_COMPILER_CLANG_PATCHLEVEL macro, which is set to `__clang_patchlevel__` for Clang and to `0` for non-Clang based compilers (or Clang versions detected using `__has_warning` or `__has_attribute` checks).

Updated HWY_BROKEN_PPC10 to mark HWY_PPC10 as non-broken on big-endian PPC with Clang 16.0.1 or later, GCC 11.x, and GCC 12.4 and later (excluding GCC 13.0 and 13.1).

All of the Highway unit tests pass on big-endian HWY_PPC10 when compiled with GCC 11.3 or Clang 16.0.4.

The __builtin_constant_p check for full vectors on PPC8/PPC9/PPC10 has also been refactored out to a new detail::IsConstantRawAltivecVect function (which is used to facilitate GCC/Clang constant folding optimizations for the Xor3, Or3, OrAnd, SumOfLanes, and SumsOf8 operations on PPC8/PPC9/PPC10).

detail::TernaryLogic (which is a wrapper for the PPC10 vec_ternarylogic intrinsic/xxeval instruction) is also now used to implement the Xor3, Or3, OrAnd, int64_t SaturatedAdd, and int64_t SaturatedSub operations on PPC10.

Also re-implemented detail::LoadMaskBits128 on PPC10 using the vec_genbm/vec_genhm/vec_genwm/vec_gendm intrinsics, which convert an mask back to a vector (similar to the _mm_movm_epi8/_mm_movm_epi16/_mm_movm_epi32/_mm_movm_epi64 operations on AVX512VL).

### DIFF
--- a/hwy/detect_compiler_arch.h
+++ b/hwy/detect_compiler_arch.h
@@ -102,14 +102,14 @@
 #else  // Anything older than 7.0 is not recommended for Highway.
 #define HWY_COMPILER_CLANG 600
 #endif  // __has_warning chain
-#define HWY_COMPILER_CLANG_PATCHLEVEL 0
+#define HWY_COMPILER3_CLANG (HWY_COMPILER_CLANG * 100)
 #else   // use normal version
 #define HWY_COMPILER_CLANG (__clang_major__ * 100 + __clang_minor__)
-#define HWY_COMPILER_CLANG_PATCHLEVEL __clang_patchlevel__
+#define HWY_COMPILER3_CLANG (__clang_major__ * 10000 + __clang_minor__ * 100 + __clang_patchlevel__)
 #endif
 #else  // Not clang
 #define HWY_COMPILER_CLANG 0
-#define HWY_COMPILER_CLANG_PATCHLEVEL 0
+#define HWY_COMPILER3_CLANG 0
 #endif
 
 #if HWY_COMPILER_GCC && !HWY_COMPILER_CLANG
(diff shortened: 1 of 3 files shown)


jan-wassenberg (approved):
Nice improvement/refactor 👍 Looks like it would have been a better design if I had initially defined the compiler version as MMmmPP (Major minor patch), oh well.

Do you think we will want patchlevel also for GCC? One thing we could consider is defining another compiler macro as the current one * 100 + patchlevel. Then we can check HWY_COMPILER3_CLANG < 160001. Let me know if you'd like to switch to something like that or prefer to keep patchlevel; I am OK with either.

johnplatts:
> Nice improvement/refactor 👍 Looks like it would have been a better design if I had initially defined the compiler version as MMmmPP (Major minor patch), oh well.
> 
> Do you think we will want patchlevel also for GCC? One thing we could consider is defining another compiler macro as the current one * 100 + patchlevel. Then we can check HWY_COMPILER3_CLANG < 160001. Let me know if you'd like to switch to something like that or prefer to keep patchlevel; I am OK with either.

I went ahead and replaced HWY_COMPILER_CLANG_PATCHLEVEL with HWY_COMPILER3_CLANG, which allows the Clang version check for Clang 16.0.1 or later to be simplified in hwy/detect_targets.h.

PR #1390, opened 2023-05-12
Fix GCC compilation error with BaseCase in vqsort-inl.h

johnplatts:
Removed HWY_RESTRICT from the keys_end and buf parameters in vqsort-inl.h to fix a compiler error when vqsort-inl.h is compiled with GCC.


jan-wassenberg (approved):
Interesting. I think this is a false positive - keys_end is only used for comparison and never de-referenced, and buf points to the same array, but is disjoint. But I'm fine to remove restrict here.

PR #1393, opened 2023-05-12
Added fixes for LoadDup128 and detail::LanesPerBlock on RVV/SVE

johnplatts:
Updated RVV LoadDup128 implementation to load no more than 16 bytes from `p`.

Also updated RVV LoadDup128 implementation to avoid an unnecessary TableLookupLanes operation in the case where `d.MaxBytes() <= 16` is true.

Updated HWY_SVE2_128 implementation of LoadDup128 to simply do a Load as the size of a vector cannot exceed 16 bytes on the HWY_SVE2_128 target.

Also updated the HWY_SVE/HWY_SVE2/HWY_SVE_256 LoadDup128 implementations to simply do a Load in the case where `d.MaxBytes() <= 16` is true.

The HWY_SVE/HWY_SVE2/HWY_SVE_256 LoadDup128 implementation will still do a svld1rq operation (wrapped by the detail::LoadDupFull128 function) in the case where `d.MaxBytes() > 16` is true.

Updated the implementation of detail::LanesPerBlock in hwy/ops/rvv-inl.h and hwy/ops/arm_sve-inl.h to use `MaxLanes(d)` as it is now possible for `MaxLanes(d) != detail::ScaleByPower(N, kPow2)` to be true in the case where `kPow2 < 0 && (detail::ScaleByPower(N, kPow2) >> (-pow2)) == 0`.

Updated the implementation of detail::LanesPerBlock in hwy/ops/rvv-inl.h to avoid unnecessary calls to `Lanes(d)` for fractional LMUL in the case where `kMaxLpb <= kMinVecLanes` is true as `kMaxLpb <= Lanes(d)` will be true if `kMaxLpb <= kMinVecLanes` is true since `Lanes(d) >= kMinVecLanes` is always true.

### DIFF
--- a/hwy/ops/arm_sve-inl.h
+++ b/hwy/ops/arm_sve-inl.h
@@ -1132,10 +1132,15 @@ HWY_API svbool_t IsFinite(const V v) {
 
 HWY_SVE_FOREACH(HWY_SVE_LOAD, Load, ld1)
 HWY_SVE_FOREACH(HWY_SVE_MASKED_LOAD, MaskedLoad, ld1)
-HWY_SVE_FOREACH(HWY_SVE_LOAD_DUP128, LoadDup128, ld1rq)
 HWY_SVE_FOREACH(HWY_SVE_STORE, Store, st1)
 HWY_SVE_FOREACH(HWY_SVE_STORE, Stream, stnt1)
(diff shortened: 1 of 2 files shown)


jan-wassenberg (approved):
Nice improvements 👍 Thank you!

PR #1399, opened 2023-05-15
Updated Iota implementations for NEON/PPC/x86

johnplatts:
Updated Iota implementation on NEON/PPC/x86 targets to use constant Iota0 vectors for each lane size, which is more optimal on GCC 5 through 11 and MSVC.

On MSVC, the NEON/x86 Iota implementation also zeroes out any lanes past the first `MaxLanes(d)` lanes to work around a MSVC compiler bug. This workaround is not needed on GCC or Clang as GCC and Clang will correctly ignore any lanes past `MaxLanes(d)` for partial vectors.

The `detail::Iota0` implementation for 128-bit or smaller vectors on SSE2/SSSE3/SSE4/AVX2/AVX3 also now uses `_mm_set_epi8`, `_mm_set_epi16`, `_mm_set_epi32`, `_mm_set_epi64x`, `_mm_set_ps`, and `_mm_set_pd` to allow more optimal code to be generated when compiled with GCC or MSVC.

GCC 5 and later, MSVC 19.20 and later, and Clang 3.4 and later will optimize the updated Iota operation down to a load of a single constant vector if `first` is constant when optimizations are enabled.

Also updated the SSE2/SSSE3/SSE4/AVX2/NEON implementations of FirstN to use `detail::Iota0` to avoid an unnecessary addition and to ensure that any lanes past the first `MaxLanes(d)` lanes are zeroed out for partial vectors if `N <= MaxLanes(d)` is true.

A link to Compiler Explorer snippets demonstrating the differences between two different Iota implementations at x86 can be found at https://godbolt.org/z/9jn33c67T.

### DIFF
--- a/hwy/ops/arm_neon-inl.h
+++ b/hwy/ops/arm_neon-inl.h
@@ -2800,7 +2800,7 @@ template <class D>
 HWY_API MFromD<D> FirstN(D d, size_t num) {
   const RebindToSigned<decltype(d)> di;  // Signed comparisons are cheaper.
   using TI = TFromD<decltype(di)>;
-  return RebindMask(d, Iota(di, 0) < Set(di, static_cast<TI>(num)));
+  return RebindMask(d, detail::Iota0(di) < Set(di, static_cast<TI>(num)));
 }
 
 // ------------------------------ TestBit (Eq)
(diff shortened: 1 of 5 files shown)


jan-wassenberg (approved):
👍 Thanks for improving these.

PR #1412, opened 2023-05-19
Updated AVX2 implementation of detail::BitsFromMask for I16/U16 masks

johnplatts:
The detail::BitsFromMask implementation for Mask256<int16_t> and Mask256<uint16_t> has been updated to use _pext_u32 instead of _pext_u64 as vectors have 32 or fewer lanes on the HWY_AVX2 target.

Also added the HWY_DISABLE_PEXT_ON_AVX2 option to disable the use of _pext_u32/_pext_u64 on the HWY_AVX2 target as Bulldozer/Zen 1/Zen 2 have AVX2 and BMI2 support but slow pext instructions.


jan-wassenberg (approved):
👍

PR #1426, opened 2023-05-26
WidenMulPairwiseAdd compilation fixes and reformatting

johnplatts:
Fixed compilation errors with WidenMulPairwiseAdd on PPC and WASM_EMU256 targets.

Also reformatted WidenMulPairwiseAdd on x86 and EMU128.


jan-wassenberg (approved):
Oops, I did not run the manual wasm256 test last time, sorry about that and thanks for fixing!

PR #1427, opened 2023-05-26
Added SetBeforeFirst, SetAtOrBeforeFirst, SetOnlyFirst, and SetAtOrAfterFirst mask ops

johnplatts:
The SetBeforeFirst, SetAtOrBeforeFirst, SetOnlyFirst, and SetAtOrAfterFirst mask operations were implemented as:
- `SetOnlyFirst(m)` is more efficient than `RebindMask(d, Eq(Iota(du, 0), static_cast<TU>(FindFirstTrue(d, m))))` on some targets
- `SetBeforeFirst(m)` is more efficient than `FirstN(d, HWY_MIN(static_cast<size_t>(FindFirstTrue(d, m)), MaxLanes(d)))` on some targets
- `SetAtOrBeforeFirst(m)` is usually more efficient than `Or(SetBeforeFirst(m), SetOnlyFirst(m))`
- RVV has instructions for the SetBeforeFirst, SetAtOrBeforeFirst, and SetOnlyFirst operations
- The implementations of SetBeforeFirst, SetAtOrBeforeFirst, SetOnlyFirst, and SetAtOrAfterFirst in hwy/ops/rvv-inl.h can correctly deal with masks that have more than 128 lanes
- SVE can efficiently carry out the SetBeforeFirst, SetAtOrBeforeFirst, and SetOnlyFirst operations using the svbrka_b_z and svbrkb_b_z intrinsics
- SVE can carry out the SetOnlyFirst operation using a single BRKA instruction
- The SetBeforeFirst, SetAtOrBeforeFirst, SetOnlyFirst, and SetAtOrAfterFirst operations can be efficiently implemented on AVX3 using the BMI1 _blsi_u32, _blsi_u64, _blsmsk_i32, and _blsmsk_i64 intrinsics
- PPC8/PPC9/PPC10 has a psubuqm instruction that can do a 128-bit negation, which allows SetBeforeFirst, SetAtOrBeforeFirst, SetOnlyFirst, and SetAtOrAfterFirst to be more efficiently implemented for 128-bit masks on PPC8/PPC9/PPC10

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -857,21 +857,22 @@ false is zero, true has all bits set:
     these operations. This op is for situations where the inputs are known to be
     mutually exclusive.
 
-*   <code>M **SetOnlyFirst**(M m)</code>: returns mask that has only lane `k`
-    set to true if any lanes are set in `m` and all other lanes set to false,
-    where `k` is the index of the first (i.e. lowest index) `m[i]` that is
-    true or the number of lanes in `m` if none of `m[i]` are true.
+*   <code>M **SetOnlyFirst**(M m)</code>: If none of `m[i]` are true, returns
+    all-false. Otherwise, only lane `k` is true, where `k` is equal to
+    `FindKnownFirstTrue(m)`. In other words, sets to false any lanes with index
+    greater than the first true lane, if it exists.
 
-*   <code>M **SetBeforeFirst**(M m)</code>: returns mask with the first `N`
-    lanes set to true and any remaining lanes set to false, where `N` is equal
-    to the index of the first `m[i]` that is true or the number of lanes in `m`
-    if none of `m[i]` are true.
+*   <code>M **SetBeforeFirst**(M m)</code>: If none of `m[i]` are true, returns
+    all-true. Otherwise, returns mask with the first `k` lanes true and all
+    remaining lanes false, where `k` is equal to `FindKnownFirstTrue(m)`. In
+    other words, if at least one of `m[i]` is true, sets to true any lanes with
+    index less than the first true lane and all remaining lanes to false.
 
-**  <code>M **SetAtOrBeforeFirst**(M m)</code>: equivalent to
+*   <code>M **SetAtOrBeforeFirst**(M m)</code>: equivalent to
     `Or(SetBeforeFirst(m), SetOnlyFirst(m))`, but `SetAtOrBeforeFirst(m)` is
     usually more efficient than `Or(SetBeforeFirst(m), SetOnlyFirst(m))`.
 
-**  <code>M **SetAtOrAfterFirst**(M m)</code>: equivalent to
+*   <code>M **SetAtOrAfterFirst**(M m)</code>: equivalent to
     `Not(SetBeforeFirst(m))`.
 
 #### Compress
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -857,6 +857,23 @@ false is zero, true has all bits set:
     these operations. This op is for situations where the inputs are known to be
     mutually exclusive.
 
+*   <code>M **SetOnlyFirst**(M m)</code>: returns mask that has only lane `k`
+    set to true if any lanes are set in `m` and all other lanes set to false,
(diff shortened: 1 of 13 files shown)


g3doc/quick_reference.md
```diff
@@ -857,6 +857,23 @@ false is zero, true has all bits set:
     these operations. This op is for situations where the inputs are known to be
     mutually exclusive.
 
+*   <code>M **SetOnlyFirst**(M m)</code>: returns mask that has only lane `k`
```
jan-wassenberg, line 860: I mis-parsed this sentence. Maybe it would be more clear to say something like this?

If none of `m[i]` are true, returns all-true. Otherwise, only lane k is true, where k = FindKnownFirstTrue(m). In other words, sets to false any lanes with index greater than the first true lane, if it exists.

jan-wassenberg (approved):
Nice, these look useful, thanks for adding.

g3doc/quick_reference.md
johnplatts: I have updated the documentation of SetOnlyFirst and SetBeforeFirst and pushed the changes up to the repository.

jan-wassenberg (dismissed):
Nice, thanks for updating :)

jan-wassenberg (approved):
Thanks! Re-running tests..

jan-wassenberg:
hm, unfortunately still failing :( I dug deeper and saw in the bmiintrin.h contents that the _blsi function documentation mentions that x86intrin.h could be required. Indeed bmiintrin.h raises an #error, saying to include x86intrin.

But that still isn't enough because clang's bmiintrin has an `#if` inside it that skips blsi `#ifdef _MSC_VER`. By contrast, bmi2intrin does not and relies on immintrin.h to check that and then not include bmi2intrin; this is why we can get away with including bmi2intrin ourselves.

I think `#if HWY_COMPILER_CLANGCL` we will require some wrapper function that replaces blsi with x & -x. Otherwise we can use blsi and rely on immintrin to define it (so we can remove the bmiintrin.h). Would you mind making such a change?

johnplatts:
> hm, unfortunately still failing :( I dug deeper and saw in the bmiintrin.h contents that the _blsi function documentation mentions that x86intrin.h could be required. Indeed bmiintrin.h raises an #error, saying to include x86intrin.
> 
> But that still isn't enough because clang's bmiintrin has an `#if` inside it that skips blsi `#ifdef _MSC_VER`. By contrast, bmi2intrin does not and relies on immintrin.h to check that and then not include bmi2intrin; this is why we can get away with including bmi2intrin ourselves.
> 
> I think `#if HWY_COMPILER_CLANGCL` we will require some wrapper function that replaces blsi with x & -x. Otherwise we can use blsi and rely on immintrin to define it (so we can remove the bmiintrin.h). Would you mind making such a change?

The issue turns out to be that clang-cl does not include the _blsi_u32, _blsi_u64, _blsmsk_u32, and _blsmsk_u64 definitions in bmiintrin.h unless the `-mbmi` option is specified (or another option such as `-march=haswell` that enables BMI).

I have pushed down a workaround that fixes the issue with _blsi_u32, _blsi_u64, _blsmsk_u32, and _blsmsk_u64 being missing on clang-cl without specifying the `-mbmi` option.

Clang-cl 10.0.0 and later actually does include the bmiintrin.h header when the immintrin.h header is included, even if the `-mbmi` option (or another compiler option that enables BMI1) in not specified on the command line.

hwy/ops/x86_128-inl.h
```diff
@@ -8575,28 +8575,72 @@ HWY_API VFromD<D> LoadExpand(MFromD<D> mask, D d,
 // ------------------------------ Additional mask logical operations
 
 #if HWY_TARGET <= HWY_AVX3
+namespace detail {
+
+template <class T, HWY_IF_LANES_LE(sizeof(T), 4)>
```
jan-wassenberg, line 8580: In future let's use HWY_IF_NOT_T_SIZE or T_SIZE_ONE_OF so it's easier to search for occurrences and/or immediately obvious the constraint concerns the lane size. But no need to change now :)

jan-wassenberg:
Looks like tests are passing 👍 We will be able to merge soon.

PR #1429, opened 2023-05-29
Added ApproximateReciprocal and ApproximateReciprocalSqrt for F64 vectors

johnplatts:
There are some targets that have instructions for F64 ApproximateReciprocal and F64 ApproximateReciprocalSqrt, including AVX3, PPC8/PPC9/PPC10, NEON on AArch64, SVE, and RVV.

There are also fallback implementations of ApproximateReciprocal and ApproximateReciprocalSqrt in hwy/ops/generic_ops-inl.h for targets that support F64 vectors but do not have F64 ApproximateReciprocal/ApproximateReciprocalSqrt instructions.

### DIFF
--- a/hwy/ops/x86_256-inl.h
+++ b/hwy/ops/x86_256-inl.h
@@ -2207,6 +2207,18 @@ HWY_API Vec256<float> ApproximateReciprocal(Vec256<float> v) {
   return Vec256<float>{_mm256_rcp_ps(v.raw)};
 }
 
+#if HWY_TARGET <= HWY_AVX3
+#ifdef HWY_NATIVE_F64_APPROX_RECIP
+#undef HWY_NATIVE_F64_APPROX_RECIP
+#else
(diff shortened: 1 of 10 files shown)


hwy/ops/x86_256-inl.h
```diff
@@ -2207,6 +2207,18 @@ HWY_API Vec256<float> ApproximateReciprocal(Vec256<float> v) {
   return Vec256<float>{_mm256_rcp_ps(v.raw)};
 }
 
+#if HWY_TARGET <= HWY_AVX3
```
jan-wassenberg, line 2210: Are we sure this behaves as intended? Isn't it enough to toggle the HWY_NATIVE flag only in x86_128 because that is included from all other x86 headers? I think we can skip this toggle here.

jan-wassenberg (commented):
Nice, thanks for adding these! One question on the implementation:

jan-wassenberg (approved):
👍

PR #1431, opened 2023-05-30
Added I8/U8/I16/U16 DupEven/DupOdd and Per4LaneBlockShuffle

johnplatts:
Added DupEven and DupOdd for I8/U8/I16/U16 vector types.

Also fixed issues on RVV where Reverse2, Reverse4, Reverse8, and ReverseBlocks sometimes fail to compile for vector types that have a fractional LMUL.

Also added a new Per4LaneBlockShuffle operation as Per4LaneBlockShuffle is usually more efficient than TableLookupBytes or TableLookupLanes for I16/U16/I32/U32/F32/I64/U64/F64 vectors.

Per4LaneBlockShuffle is equivalent to SSE2 `_mm_shufflelo_epi16(v, _MM_SHUFFLE(kIdx3, kIdx2, kIdx1, kIdx0))`, SSE2 `_mm_shuffle_epi32(v, _MM_SHUFFLE(kIdx3, kIdx2, kIdx1, kIdx0))`, and AVX2 `_mm256_permute4x64_epi64(v, _MM_SHUFFLE(kIdx3, kIdx2, kIdx1, kIdx0))`.

In the case of I64/U64/F64 vectors that have 4 or more lanes, the Per4LaneBlockShuffle operation does a per 256-bit block shuffle of `v` (same behavior as AVX2 `_mm256_permute4x64_epi64` and AVX512 `_mm512_permutex_epi64`).

Also added <code>ForPartialFixedOrFullScalableVectors&lt;Test&gt;</code> in hwy/tests/test_util-inl.h, which only calls `Test` for for each `D` where
`MaxLanes(D()) == MaxLanes(DFromV<VFromD<D>>())`.

ForPartialFixedOrFullScalableVectors allows a test to be invoked for all of the possible vector types for a lane type `T`, but skips over cases where `MaxLanes(D()) < MaxLanes(DFromV<VFromD<D>>())`, which is possible on RVV/SVE targets.

`ForAllTypes(ForPartialVectors<TestPer4LaneBlockShuffle>())` causes the compilation of hwy/tests/swizzle_test.cc to get stuck on the HWY_RVV target, but replacing `ForAllTypes(ForPartialVectors<TestPer4LaneBlockShuffle>())` with `ForAllTypes(ForPartialFixedOrFullScalableVectors<TestPer4LaneBlockShuffle>())` fixes the problem on the HWY_RVV target.

On targets where `MaxLanes(D()) == MaxLanes(DFromV<VFromD<D>>())` is always true, `ForPartialFixedOrFullScalableVectors<Test>` is simply a typedef for `ForPartialVectors<Test>`.

### DIFF
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -3119,179 +3119,288 @@ HWY_API V ReverseBits(V v) {
 
 // ------------------------------ Per4LaneBlockShuffle
 
+#if (defined(HWY_NATIVE_PER4LANEBLKSHUF_DUP32) == defined(HWY_TARGET_TOGGLE))
+#ifdef HWY_NATIVE_PER4LANEBLKSHUF_DUP32
+#undef HWY_NATIVE_PER4LANEBLKSHUF_DUP32
+#else
+#define HWY_NATIVE_PER4LANEBLKSHUF_DUP32
+#endif
+
 #if HWY_TARGET != HWY_SCALAR
 namespace detail {
 
-template <class V, HWY_IF_LANES_GT_D(DFromV<V>, 1)>
-HWY_INLINE V Per2LaneBlockShuffle(hwy::SizeTag<0> /*idx_0_tag*/,
-                                  hwy::SizeTag<0> /*idx_1_tag*/, V v) {
+template <class D>
+HWY_INLINE Vec<D> Per4LaneBlkShufDupSet4xU32(D d, const uint32_t x3,
+                                             const uint32_t x2,
+                                             const uint32_t x1,
+                                             const uint32_t x0) {
+  alignas(16) const uint32_t lanes[4] = {x0, x1, x2, x3};
+
+#if HWY_TARGET == HWY_RVV
+  constexpr int kPow2 = d.Pow2();
+  constexpr int kLoadPow2 = HWY_MAX(kPow2, -1);
+  const ScalableTag<uint32_t, kLoadPow2> d_load;
+#else
+  constexpr size_t kMaxBytes = d.MaxBytes();
+#if HWY_TARGET == HWY_NEON || HWY_TARGET == HWY_NEON_WITHOUT_AES
+  constexpr size_t kMinLanesToLoad = 2;
+#else
+  constexpr size_t kMinLanesToLoad = 4;
+#endif
+  constexpr size_t kNumToLoad =
+      HWY_MAX(kMaxBytes / sizeof(uint32_t), kMinLanesToLoad);
+  const CappedTag<uint32_t, kNumToLoad> d_load;
+#endif
+
+  return ResizeBitCast(d, LoadDup128(d_load, lanes));
+}
+
+}  // namespace detail
+#endif
+
+#endif  // HWY_NATIVE_PER4LANEBLKSHUF_DUP32
+
+#if HWY_TARGET != HWY_SCALAR
+namespace detail {
+
+template <class V>
+HWY_INLINE V Per2LaneBlockShuffle(hwy::SizeTag<0> /*idx_10_tag*/, V v) {
   return DupEven(v);
 }
 
-template <class V, HWY_IF_LANES_GT_D(DFromV<V>, 1)>
-HWY_INLINE V Per2LaneBlockShuffle(hwy::SizeTag<1> /*idx_0_tag*/,
-                                  hwy::SizeTag<1> /*idx_1_tag*/, V v) {
-  return DupOdd(v);
-}
-
-template <size_t kIdx0, class V, HWY_IF_LANES_GT_D(DFromV<V>, 1)>
-HWY_INLINE V Per2LaneBlockShuffle(hwy::SizeTag<kIdx0> /*idx_0_tag*/,
-                                  hwy::SizeTag<0> /*idx_1_tag*/, V v) {
-  static_assert(kIdx0 != 0, "kIdx0 != 0 should be true");
+template <class V>
+HWY_INLINE V Per2LaneBlockShuffle(hwy::SizeTag<1> /*idx_10_tag*/, V v) {
   const DFromV<decltype(v)> d;
   return Reverse2(d, v);
 }
 
-template <size_t kIdx0, size_t kIdx1, class V>
-HWY_INLINE V Per2LaneBlockShuffle(hwy::SizeTag<kIdx0> /*idx_0_tag*/,
-                                  hwy::SizeTag<kIdx1> /*idx_1_tag*/, V v) {
-  static_assert(
-      HWY_MAX_LANES_V(V) <= 1 || kIdx1 != 0,
-      "kIdx1 != 0 should be true if HWY_MAX_LANES_V(V) is greater than 1");
-  static_assert(HWY_MAX_LANES_V(V) <= 1 || kIdx0 != 1 || kIdx1 != 1,
-                "kIdx0 != 1 or kIdx1 != 1 should be true if HWY_MAX_LANES_V(V) "
-                "is greater than 1");
+template <class V>
+HWY_INLINE V Per2LaneBlockShuffle(hwy::SizeTag<2> /*idx_10_tag*/, V v) {
   return v;
 }
 
-template <size_t kLaneSize, class VT, class VI>
-HWY_INLINE VT Per4LaneBlockShufTblLookup(
-    hwy::SizeTag<kLaneSize> /* lane_size_tag */, VT vt, VI vi) {
-  return TableLookupBytes(vt, vi);
+template <class V>
+HWY_INLINE V Per2LaneBlockShuffle(hwy::SizeTag<3> /*idx_10_tag*/, V v) {
+  return DupOdd(v);
 }
 
-template <size_t kIdx0, size_t kIdx1, size_t kIdx2, size_t kIdx3, class D>
-HWY_INLINE VFromD<D> Per4LaneBlockShufIdx(hwy::SizeTag<1> /* lane_size_tag */,
-                                          D d,
-                                          hwy::SizeTag<kIdx0> /*idx_0_tag*/,
-                                          hwy::SizeTag<kIdx1> /*idx_1_tag*/,
-                                          hwy::SizeTag<kIdx2> /*idx_2_tag*/,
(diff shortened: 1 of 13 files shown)


hwy/ops/generic_ops-inl.h
```diff
@@ -3117,6 +3117,321 @@ HWY_API V ReverseBits(V v) {
...
+  const DFromV<decltype(v)> d;
+  const RepartitionToWide<decltype(d)> dw;
+  return BitCast(dw, v);
+}
+
+template <class V, HWY_IF_LANES_GT_D(DFromV<V>, 2),
+          HWY_IF_T_SIZE_ONE_OF_V(V, (1 << 1) | (1 << 2) |
+                                        (HWY_HAVE_INTEGER64 ? (1 << 4) : 0))>
+HWY_INLINE V Per4LaneBlockShuffle(hwy::SizeTag<0> /*idx_0_tag*/,
+                                  hwy::SizeTag<1> /*idx_1_tag*/,
```
jan-wassenberg, line 3330: Shouldn't DupEven be idx3..0 = 2200?

hwy/ops/generic_ops-inl.h
```diff
@@ -3117,6 +3117,321 @@ HWY_API V ReverseBits(V v) {
...
+  const DFromV<decltype(v)> d;
+  const auto vw = Per4LaneBlockShufCastToWide(
+      hwy::IsFloatTag<TFromV<V>>(), hwy::SizeTag<sizeof(TFromV<V>)>(), v);
+  return BitCast(d, DupEven(vw));
+}
+
+template <class V, HWY_IF_LANES_GT_D(DFromV<V>, 2),
+          HWY_IF_T_SIZE_ONE_OF_V(V, (1 << 1) | (1 << 2) |
+                                        (HWY_HAVE_INTEGER64 ? (1 << 4) : 0))>
+HWY_INLINE V Per4LaneBlockShuffle(hwy::SizeTag<2> /*idx_0_tag*/,
```
jan-wassenberg, line 3342: And here also: if idx3 is the most-significant, wouldn't Reverse2 idx3..0 = 2301? (idx3=2)

hwy/ops/generic_ops-inl.h
```diff
@@ -3117,6 +3117,321 @@ HWY_API V ReverseBits(V v) {
...
+      hwy::IsFloatTag<TFromV<V>>(), hwy::SizeTag<sizeof(TFromV<V>)>(), v);
+  const DFromV<decltype(vw)> dw;
+  return BitCast(d, Reverse2(dw, vw));
+}
+
+template <class V, HWY_IF_LANES_GT_D(DFromV<V>, 2),
+          HWY_IF_T_SIZE_ONE_OF_V(V, (1 << 1) | (1 << 2) |
+                                        (HWY_HAVE_INTEGER64 ? (1 << 4) : 0))>
+HWY_INLINE V Per4LaneBlockShuffle(hwy::SizeTag<2> /*idx_0_tag*/,
+                                  hwy::SizeTag<3> /*idx_1_tag*/,
```
jan-wassenberg, line 3357: idx3..0 = 3311?

hwy/ops/generic_ops-inl.h
```diff
@@ -3117,6 +3117,321 @@ HWY_API V ReverseBits(V v) {
...
+HWY_INLINE V Per4LaneBlockShuffle(hwy::SizeTag<0> /*idx_0_tag*/,
+                                  hwy::SizeTag<0> /*idx_1_tag*/,
+                                  hwy::SizeTag<1> /*idx_2_tag*/,
+                                  hwy::SizeTag<1> /*idx_3_tag*/, V v) {
+  const DFromV<decltype(v)> d;
+  return InterleaveLower(d, v, v);
+}
+
+template <class V, HWY_IF_LANES_GT_D(DFromV<V>, 2), HWY_IF_T_SIZE_V(V, 4)>
+HWY_INLINE V Per4LaneBlockShuffle(hwy::SizeTag<0> /*idx_0_tag*/,
```
jan-wassenberg, line 3386: Can we merge this with the above overload?

hwy/ops/wasm_256-inl.h
```diff
@@ -1267,6 +1267,89 @@ HWY_API Vec256<T> ReverseBlocks(D /* tag */, const Vec256<T> v) {
...
 
+// ------------------------------ Per4LaneBlockShuffle
+namespace detail {
+
+template <size_t kIdx0, size_t kIdx1, size_t kIdx2, size_t kIdx3, class V>
+HWY_INLINE V Per4LaneBlockShuffle(hwy::SizeTag<kIdx0> /*idx_0_tag*/,
+                                  hwy::SizeTag<kIdx1> /*idx_1_tag*/,
+                                  hwy::SizeTag<kIdx2> /*idx_2_tag*/,
+                                  hwy::SizeTag<kIdx3> /*idx_3_tag*/,
+                                  hwy::SizeTag<1> /*lane_size_tag*/,
```
jan-wassenberg, line 1278: We could handle sizes <= 4 by calling ::Per4LaneBlockShuffle() for each 128-bit half?

hwy/tests/swizzle_test.cc
```diff
@@ -343,6 +343,154 @@ HWY_NOINLINE void TestAllTwoTablesLookupLanes() {
...
+        And(GenerateTestVect(hwy::NonFloatTag(), du), Set(du, kIntBitsMask));
+    return Or(flt_iota, BitCast(d, int_iota));
+  }
+
+  template <class D>
+  static HWY_INLINE void DoTestPer4LaneBlkShuffles(
+      D d, const size_t N, const VFromD<D> v, TFromD<D>* HWY_RESTRICT src_lanes,
+      TFromD<D>* HWY_RESTRICT expected) {
+    Store(v, d, src_lanes);
+    DoTestPer4LaneBlkShuf<0, 0, 0, 0>(d, N, v, src_lanes, expected);
```
jan-wassenberg, line 441: How about we make this truly exhaustive for safety? To avoid 256 lines, we can have a function ForeachIdx0 that calls ForeachIdx1<i> for i=0..3 which then calls ForeachIdx2<i, j> etc.

jan-wassenberg (commented):
Wow! I like your idea for this op, seems useful.

hwy/tests/swizzle_test.cc
johnplatts: The issue that I was running to with testing all 256 cases of `Per4LaneBlockShuffle<kIdx3, kIdx2, kIdx1, kIdx0>(v)` using the `ForeachIdx1<i>`, `ForeachIdx2<i, j>`, and `ForeachIdx3<i, j, k>` approach is that the compilation of hwy/tests/swizzle_test.cc is slow and consumes too much memory.

I had refactored the default implementation of `Per4LaneBlockShuffle<kIdx3, kIdx2, kIdx1, kIdx0>(v)` (which does table lookup) to detail::TblLookupPer4LaneBlkShuf, which takes a non-template `idx3210` parameter to allow detail::TblLookupPer4LaneBlkShuf to be unit tested using a loop.

I had also added a TestTblLookupPer4LaneBlkShuf function in TestPer4LaneBlockShuffle that checks the results of detail::TblLookupPer4LaneBlkShuf for all 256 cases using a loop.

Here is the loop that is used to check all 256 cases of detail::TblLookupPer4LaneBlkShuf:
```
    for (size_t idx3210 = 0; idx3210 <= 0xFF; idx3210++) {
      const size_t idx3 = (idx3210 >> 6) & 3;
      const size_t idx2 = (idx3210 >> 4) & 3;
      const size_t idx1 = (idx3210 >> 2) & 3;
      const size_t idx0 = idx3210 & 3;

      const auto actual = detail::TblLookupPer4LaneBlkShuf(v, idx3210);
      DoCheckPer4LaneBlkShufResult(d, N, actual, src_lanes, expected, idx3,
                                   idx2, idx1, idx0);
    }
```

hwy/ops/generic_ops-inl.h
```diff
@@ -3161,6 +3161,470 @@ HWY_API V ReverseBits(V v) {
...
+
+#if HWY_TARGET == HWY_EMU128
+  constexpr size_t kN = HWY_MAX_LANES_D(D);
+  if (kN < 4) {
+    idx_in_blk = And(idx_in_blk, Set(du, static_cast<TU>(kN - 1)));
+  }
+#endif
+
+#if HWY_TARGET == HWY_RVV
+  const auto blk_offsets = AndS(Iota0(du), static_cast<TU>(~TU{3}));
```
jan-wassenberg, line 3413: This pattern seems frequent enough to optimize; SVE also has something similar. No need to change in this patch but I am considering adding an op for And/Xor-with-constant; probably also comparisons and add/sub, maybe min/max. What do you think? Any preference on the naming, e.g. AndS (RVV: 'scalar') or AndN (SVE) or AndK ('constant')?

jan-wassenberg (dismissed):
Nice, thanks for making the changes. One comment for future consideration, but let's merge this already.

jan-wassenberg (approved):

jan-wassenberg:
Unfortunately it's also failing on AVX2, and also would on SSE4. Sorry I didn't realize that.

The IndicesFromVec actually is documented that it requires in-range inputs. Isn't it safest then if we do the And() on all platforms, not just the EMU128 and X86 that check this precondition?

jan-wassenberg (approved):
👍

johnplatts:
> @johnplatts sorry this has required rather more iterations, we often get results one at a time from CI :/ There's an issue with test_util-inl.h:579:5 on fixed-size SVE targets: `error: no template named 'ForeachPow2' in namespace 'hwy::N_SVE_256` This is because ForeachPow2 is only defined `#if HWY_HAVE_SCALABLE` and SVE_256 is considered fixed-size, not scalable. I think it's reasonable to keep the "fixed-size is not scalable", so options include always defining ForeachPow2, or only using this ForeachPow2 if HWY_HAVE_SCALABLE. Any preference?

I have pushed down a fix that fixes the compilation issue for the HWY_SVE_256 and HWY_SVE2_128 targets that defines `ForPartialFixedOrFullScalableVectors` as a typedef for `ForGEVectors<HWY_MAX_BYTES * 8, Test>` on the HWY_SVE_256 and HWY_SVE2_128 targets.

jan-wassenberg (approved):
Looks good, thanks for making the change :)

PR #1434, opened 2023-06-02
Fix compilation error with bench_sort.cc on Clang 6

johnplatts:
Fixed issue where bench_sort.cc failed to compile on Clang 6 as the RandomState constructor takes a uint64_t argument.


jan-wassenberg (approved):
Thank you!

PR #1437, opened 2023-06-06
Added InsertBlock, ExtractBlock, BroadcastBlock, BroadcastLane, and NumOfBlocks

johnplatts:
Added the following new operations:
- InsertBlock - inserts a block of `HWY_MIN(Lanes(DFromV<V>()), 16 / sizeof(TFromV<V>))` lanes to the result vector if `kBlock * 16` is less than `Lanes(DFromV<V>()) * sizeof(TFromV<V>)`. Otherwise, returns `v`.
- ExtractBlock - extracts a block of `HWY_MIN(Lanes(DFromV<V>()), 16 / sizeof(TFromV<V>))` lanes from `v`
- BroadcastBlock - broadcasts a block of `HWY_MIN(Lanes(DFromV<V>()), 16 / sizeof(TFromV<V>))` lanes to all of the other blocks of the result vector
- BroadcastLane - broadcasts a lane to all of the other lanes of the result vector (even in the case of vectors that are larger than 16 bytes)
- NumOfBlocks - returns the number of blocks with `HWY_MIN(Lanes(DFromV<V>()), 16 / sizeof(TFromV<V>))` lanes

Also added `BlockDFromD`, which returns a SIMD tag for a block of `HWY_MIN(Lanes(D()), 16 / sizeof(TFromD<D>))` lanes

Also added I8/U8 Broadcast for the SSE2/SSSE3/SSE4/AVX2/AVX3, NEON, and WASM targets (I8/U8 was already implemented on the SVE, RVV, PPC, EMU128, and SCALAR targets).

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -224,6 +224,9 @@ lane count, thus avoiding the need for a second loop to handle remainders.
 *   The result of `UpperHalf`/`LowerHalf` has half the lanes. To obtain a
     corresponding `d`, use `Half<decltype(d)>`; the opposite is `Twice<>`.
 
+*   `BlockDFromD<D>` returns a `d` with a lane type of `TFromD<D>` and
+    `HWY_MIN(HWY_MAX_LANES_D(D), 16 / sizeof(TFromD<D>))` lanes.
+
 User-specified lane counts or tuples of vectors could cause spills on targets
 with fewer or smaller vectors. By contrast, Highway encourages vector-length
 agnostic code, which is more performance-portable.
@@ -402,6 +405,20 @@ time-critical code:
     lane `i` is set to `t`. `i` must be in `[0, Lanes(DFromV<V>()))`.
     Potentially slow, it may be better set all elements of an aligned array and
     then `Load` it.
+
+### Getting/setting blocks
+
+*   <code>Vec<BlockDFromD<DFromV<V>>> **ExtractBlock**&lt;int kBlock&gt;(V)
+    </code>: returns block `kBlock` of V, where `kBlock` is an index to a block
+    that is `HWY_MIN(DFromV<V>().MaxBytes(), 16)` bytes.
+
+    `kBlock` must be in `[0, DFromV<V>().MaxBlocks())`.
+
+*   <code>V **InsertBlock**&lt;int kBlock&gt;(V v, Vec<BlockDFromD<DFromV<V>>>
+    blk_to_insert)</code>: Inserts `blk_to_insert` at lane offset
+    `kBlock * (16 / sizeof(TFromV<V>))`.
+
+    `kBlock` must be in `[0, DFromV<V>().MaxBlocks())`.
 
 ### Printing
 
@@ -1338,8 +1355,7 @@ more expensive on AVX2/AVX-512 than per-block operations.
 **Note**: if vectors are larger than 128 bits, the following operations split
 their operands into independently processed 128-bit *blocks*.
 
-*   `V`: `{u,i}{16,32,64}, {f}` \
-    <code>V **Broadcast**&lt;int i&gt;(V)</code>: returns individual *blocks*,
+*   <code>V **Broadcast**&lt;int i&gt;(V)</code>: returns individual *blocks*,
     each with lanes set to `input_block[i]`, `i = [0, 16/sizeof(T))`.
 
 All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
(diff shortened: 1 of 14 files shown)


g3doc/quick_reference.md
```diff
@@ -403,6 +406,29 @@ time-critical code:
...
+    blk_to_insert)</code>: Inserts `blk_to_insert`, with `blk_to_insert[i]`
+    inserted into lane `kBlock * (16 / sizeof(TFromV<V>)) + i` of the result
+    vector, if `kBlock * 16 < Lanes(DFromV<V>()) * sizeof(TFromV<V>)` is true.
+
+    Otherwise, returns `v` if `kBlock * 16` is greater than or equal to
+    `Lanes(DFromV<V>()) * sizeof(TFromV<V>)`.
+
+    `kBlock` must be in `[0, DFromV<V>().MaxBlocks())`.
+
+*   <code>size_t **NumOfBlocks**(D d)</code>: Returns the number of 16-byte
```
jan-wassenberg, line 427: Would you mind renaming this to Blocks(d) for consistency with Lanes(d)?

hwy/ops/arm_neon-inl.h
```diff
@@ -4537,6 +4567,16 @@ HWY_API Vec128<uint64_t> Broadcast(Vec128<uint64_t> v) {
 
 // Signed
 template <int kLane>
+HWY_API Vec128<int8_t> Broadcast(Vec128<int8_t> v) {
+  static_assert(0 <= kLane && kLane < 16, "Invalid lane");
+  return Vec128<int8_t>(vdupq_n_s8(vgetq_lane_s8(v.raw, kLane)));
```
jan-wassenberg, line 4572: I think we can skip the vgetq by using vdupq_laneq. Would you mind also updating the other instances of this?

hwy/ops/arm_sve-inl.h
```diff
@@ -2678,6 +2678,92 @@ HWY_API VFromD<D> Reverse8(D d, const VFromD<D> v) {
...
 
+// ------------------------------ Block insert/extract/broadcast ops
+#ifdef HWY_NATIVE_BLOCKDFROMD
+#undef HWY_NATIVE_BLOCKDFROMD
+#else
+#define HWY_NATIVE_BLOCKDFROMD
+#endif
+
+template <class D>
+using BlockDFromD =
```
jan-wassenberg, line 2689: Would it make sense to define this in shared-inl.h for targets except RVV?

hwy/ops/rvv-inl.h
```diff
@@ -3094,6 +3098,150 @@ HWY_API V Broadcast(const V v) {
...
+#endif
+
+template <int kLane, class V,
+          hwy::EnableIf<(sizeof(TFromV<V>) != 1 || kLane <= 255)>* = nullptr>
+HWY_API V BroadcastLane(V v) {
+  static_assert(0 <= kLane && kLane < HWY_MAX_LANES_V(V), "Invalid lane");
+  const DFromV<V> d;
+  const RebindToUnsigned<decltype(d)> du;
+  using TU = TFromD<decltype(du)>;
+  return TableLookupLanes(v, Set(du, static_cast<TU>(kLane)));
```
jan-wassenberg, line 3115: RVV's rgather_vx is likely much more efficient here. Let's use that in all three overloads here?
https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/auto-generated/policy_funcs/intrinsic_funcs/10_vector_permutation_functions.md

jan-wassenberg (changes requested):
Thanks for adding these, looks useful and also nice to fill in the 8-bit gap for Broadcast :)
Some minor suggestions for updates:

hwy/ops/arm_neon-inl.h
johnplatts: The issue is that ARMv7 NEON does not have the vdupq_laneq intrinsics that are available on AArch64.

Need to keep the `vdupq_n_s8(vgetq_lane_s8(v.raw, kLane))` for ARMv7.

jan-wassenberg: Oh, I see - the `#if` and especially the "No vdupq_laneq" comment were not visible in the diff, sorry about that. Never mind :)

g3doc/quick_reference.md
johnplatts: Made the changes

hwy/ops/arm_sve-inl.h
johnplatts: Already moved the BlockDFromD implementation for non-RVV targets over to shared-inl.h

hwy/ops/rvv-inl.h
johnplatts: Re-implemented BroadcastLane using the RVV rgather_vx intrinsics.

jan-wassenberg (approved):
Nice, thanks for making the changes :)

jan-wassenberg:
CI ran into this:
HighwayTest.TestAllBlockDFromD/SVE
Abort at third_party/highway/hwy/highway_test.cc:506: Assert d_block_bytes <= d_bytes

Are you able to repro it?

johnplatts:
> CI ran into this: HighwayTest.TestAllBlockDFromD/SVE Abort at third_party/highway/hwy/highway_test.cc:506: Assert d_block_bytes <= d_bytes
> 
> Are you able to repro it?

I was able to reproduce the problem, but the problem turned out to be due to a bug in the Lanes implementation in hwy/ops/arm_sve-inl.h.

I have pushed a fix to the bug that is there with Lanes on SVE, and the fix I made to the SVE Lanes implementation does fix the problem with the TestAllBlockDFromD test on the HWY_SVE target.

jan-wassenberg (approved):
Thanks for fixing SVE Lanes(). It is a bit subtle: we do want to honor both the user's max N and the fraction, both of those come in through MaxLanes. But then we also want to scale the actual HW lanes by (clamped) pow2, so your code looks good.

PR #1438, opened 2023-06-08
Implemented fix for RVV Lanes

johnplatts:
Added fix for RVV Lanes operation to deal with the fractional `N` case.

Also updated TestAllMaxLanes in hwy/highway_test.cc to verify the result of `Lanes(d)` for the cases where `N` is a fractional N.

The fix to the SVE Lanes function that is in pull request #1437 needs to be applied to get the updated version of TestAllMaxLanes (which includes the checks for the fractional N cases) to pass on the HWY_SVE and HWY_SVE2 targets.

### DIFF
--- a/hwy/highway_test.cc
+++ b/hwy/highway_test.cc
@@ -81,7 +81,7 @@ static size_t* MaxLanesForSize(size_t sizeof_t) {
 
 struct TestMaxLanes {
   template <class T, class D>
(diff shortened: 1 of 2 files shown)


jan-wassenberg (approved):
Thanks, nice simplification of RVV Lanes(). FYI #1437 is currently stuck on testing, hopefully that will succeed overnight, then we can begin importing this one.

PR #1444, opened 2023-06-13
Sin cos

nicolacandussi:
- Added SinCos math function. It computes Sin and Cos at the same time at roughly the same price of a Sin call. 
- Implementation for both double and single precision.
- Added the corresponding tests.

Note: 
The tests adjust the max ulp based on the target. On some targets (in particular EMU128 and Intel SSE2/3/4) the precision drops.


nicolacandussi:
Had problems with passing the CLA check because I submitted some commits with the wrong author email address. Re-opening another pull request later...

PR #1448, opened 2023-06-13
Added SinCos

nicolacandussi:
- Added SinCos math function. It computes Sin and Cos at the same time at roughly the same price of a Sin call.
- Implementation for both double and single precision.
- Added the corresponding tests.

Note:
The tests adjust the max ulp based on the target. On some targets (in particular EMU128 and Intel SSE2/3/4) the precision drops.

### DIFF
--- a/hwy/contrib/math/math-inl.h
+++ b/hwy/contrib/math/math-inl.h
@@ -1025,13 +1025,9 @@ HWY_INLINE void SinCos3(D d,
   using M = Mask<D>;
 
   static constexpr size_t bits = sizeof(TI) * 8;
-  // equivalent to 0x80000000 for int32 and 0x8000000000000000L for int64
-  static constexpr TI mask = TI(1) << (bits - 1);
-  const VI sign_mask = Set(di, mask);
-  const VI inv_sign_mask = Set(di, ~mask);
+  const VI sign_mask = SignBit(di);
   const VI ci_0 = Zero(di);
   const VI ci_1 = Set(di, 1);
-  const VI ci_inv_1 = Set(di, ~1);
   const VI ci_2 = Set(di, 2);
   const VI ci_4 = Set(di, 4);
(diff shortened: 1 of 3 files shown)


hwy/contrib/math/math-inl.h
```diff
@@ -989,6 +1008,257 @@ HWY_INLINE V Log(const D d, V x) {
...
+  using TI = MakeSigned<T>;
+  using DI = Rebind<TI, D>;
+  const DI di;
+  using VI = decltype(Zero(di));
+  using M = Mask<D>;
+
+  static constexpr size_t bits = sizeof(TI) * 8;
+  // equivalent to 0x80000000 for int32 and 0x8000000000000000L for int64
+  static constexpr TI mask = TI(1) << (bits - 1);
+  const VI sign_mask = Set(di, mask);
```
jan-wassenberg, line 1138: Also SignBit?

hwy/contrib/math/math-inl.h
```diff
@@ -989,6 +1008,257 @@ HWY_INLINE V Log(const D d, V x) {
...
+  imm0 = ShiftLeft<bits-3>(imm0);
+
+  V swap_sign_bit_sin = BitCast(d, imm0);
+
+  /* get the polynomial selection mask for the sine*/
+  imm2 = And(imm2, ci_2);
+  imm2 = IfThenElseZero(Eq(imm2, ci_0), Set(di, -1));
+      
+  // Cast integer 0000 FFFF (negative int) to mmask type. Is there a better way?
+  M poly_mask = RebindMask(d, Lt(imm2, Zero(di)));
```
jan-wassenberg, line 1197: It seems we want a mask saying whether imm2 == 0. Why not just go to poly_mask = Eq(imm2, ci_0)? (also above)

jan-wassenberg (approved):
Nice, thanks for making the changes :)

PR #1473, opened 2023-06-18
Fixed HWY_TARGET <= HWY_SSSE3 check in UI8 Reverse

johnplatts:
Replaced `HWY_TARGET <= HWY_SSE3` with `HWY_TARGET <= HWY_SSSE3` in the UI8 Reverse implementation in x86_128-inl.h


jan-wassenberg (approved):
👍

PR #1478, opened 2023-06-20
Explicit public linking to hwy for hwy_contrib

malaterre:
Recent cmake version fails to handle both target_link_libraries signatures and reports:

CMake Error at CMakeLists.txt:335 (target_link_libraries):
  The plain signature for target_link_libraries has already been used with
  the target "hwy_contrib".  All uses of target_link_libraries with a target
  must be either all-keyword or all-plain.

  The uses of the plain signature are here:

   * CMakeLists.txt:324 (target_link_libraries)

This commit fixes this configuration error

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -321,7 +321,7 @@ endif()
 
 if (HWY_ENABLE_CONTRIB)
 add_library(hwy_contrib ${HWY_LIBRARY_TYPE} ${HWY_CONTRIB_SOURCES})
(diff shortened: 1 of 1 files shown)


jan-wassenberg:
Thank you :) Should we also do this for some of the other places, including hwy_test and hwy_list_targets? (Let me know if you'd prefer I do that in a separate change.)

jan-wassenberg:
Thanks, will do 👍

jan-wassenberg (approved):
👍

Issue #1489, opened 2023-06-21
arm64: math_test.cc:111:23: error: no matches converting function ‘SinCosSin’


malaterre:
Build logs on arm64 are a bit of mess because of an ICE, but it seems g++-13 fails to parse the code here:

```
[110/120] /usr/bin/g++-13 -DHWY_SHARED_DEFINE -I"/<<PKGBUILDDIR>>" -g -O2 -ffile-prefix-map=/<<PKGBUILDDIR>>=. -fstack-protector-strong -Wformat -Werror=format-security -DHWY_BROKEN_EMU128=0 -Wdate-time -D_FORTIFY_SOURCE=2 -fPIE -fvisibility=hidden -fvisibility-inlines-hidden -Wno-builtin-macro-redefined -D__DATE__=\"redacted\" -D__TIMESTAMP__=\"redacted\" -D__TIME__=\"redacted\" -fmerge-all-constants -Wall -Wextra -Wconversion -Wsign-conversion -Wvla -Wnon-virtual-dtor -fmath-errno -fno-exceptions -Werror -DHWY_IS_TEST=1 -DGTEST_HAS_PTHREAD=1 -MD -MT CMakeFiles/math_test.dir/hwy/contrib/math/math_test.cc.o -MF CMakeFiles/math_test.dir/hwy/contrib/math/math_test.cc.o.d -o CMakeFiles/math_test.dir/hwy/contrib/math/math_test.cc.o -c '/<<PKGBUILDDIR>>/hwy/contrib/math/math_test.cc'
FAILED: CMakeFiles/math_test.dir/hwy/contrib/math/math_test.cc.o 
[...]

PR #1490, opened 2023-06-21
Fix compilation errors with SinCosSin and SinCosCos in math_test.cc

johnplatts:
Resolves issue #1489


jan-wassenberg (approved):
Thanks for fixing the build :)

PR #1494, opened 2023-06-22
Fix compilation error in SVE debug build

johnplatts:
Fixed IndicesFromVec so that SVE debug build will successfully compile


jan-wassenberg (approved):
👍

Issue #1455, opened 2023-06-14
Automatic unrolling, hoisting and reordering for loops


matthewkolbe:
I have found myself needing to implement a lot of optimizations that compilers struggle with in the presence of intrinsic functions. I shared this example with @jan-wassenberg on hacker news: https://godbolt.org/z/cMafh4Pnv

Since I had to do all of the unrolling, hoisting, reordering and loop cleanup over and over, I created a sort of generalized optimization template to take basic information about the loop, and do a sensible set of optimizations for a lot of x86 chips.  

Below is a basic example using highway. I normally use it on intrinsics, so I just did a quick example in highway. I'll make it more thorough if this is interesting to you all.

https://gcc.godbolt.org/z/9hrrGs6We

The gist is it uses the CRTP with default implementations to provide an `unroller` function the necessary implementation details to optimize the loop. Then you pass the data and an instance of this function to the `unroller` function to run a reasonably well optimized loop for a typical use case. 

My questions: is this interesting? And if so, does it belong in highway? I have found it takes a lot of headache and confusion away when using intrinsics.

jan-wassenberg:
Hi @matthewkolbe , sorry to reply late. That's indeed interesting! It hides much of the difficulty of remainder handling.
We welcome such things in hwy/contrib :)

What you have is more flexible than the hwy/contrib/transform-inl.h, which does take care of masking but currently only manages to 'zip' several arrays and then apply a function to each vector, including remnants.
I like your idea to also support reduction.

For maskload/maskstore, check out transform-inl. Unfortunately we require an `#if` to check whether the platform's masking is even safe (AVX-512 yes, AVX2 no).

Sometimes I've seen better codegen from returning vectors rather than output parameters. Might be good to switch to that.

FYI Highway currently only requires C++11. It would be nice to retain support for that if possible, resorting to `#if` when necessary (see HWY_GENERIC_LAMBDA). Do you think that's feasible?

matthewkolbe:
@jan-wassenberg, it looks like the feature I most depend on is a C++20 feature. In order to have a default "base class" implementation within CRTP, I used the `requires` expression. I'll look into it more, but I'm not sure I can get around that, but if instead we use single-class inheritance, it would be just as performant, right?

[...]

PR #1511, opened 2023-06-29
Unroller

matthewkolbe:
@jan-wassenberg  I integrated all your suggestions from [our discussion](https://github.com/google/highway/issues/1455#issuecomment-1609036595) except partial D/tags. I'm not necessarily opposed to that, but I don't think I understand it. 

I integrated tests. I have some basic questions about debugging that I was hoping you could help with: My code is emitting a lot of warnings. What's your process for debugging the warnings?

### DIFF
--- a/hwy/contrib/unroller/README.md
+++ b/hwy/contrib/unroller/README.md
@@ -1,5 +1,7 @@
 # Unroller
 
+All contents of the `unroller` folder are experimental and subject to changes.
+
 `Unroller` is a templated function that automatically implements common optimizations that are usually handled by compilers when writing scalar code. Modern CPUs operate much more efficiently when non-dependent calculations are packed into an instruction pipeline. For scalar code, this often means a compiler will take a one-line loop, and compile it down to hundreds of lines of machine code in order to fully capture these efficiencies. 
 
 As of today (2023-07-06), compilers are not nearly as good at implementing these optimizations for code written in SIMD intrinsics. `Unroller` is a templated function that takes in an `UnrollerUnit` of SIMD instructions, and then implements unrolling, reordering, hoisting and tail-handling (URHT optimizations) of arrays of data being processed with SIMD intrinsics. 
@@ -11,7 +13,7 @@ As of today (2023-07-06), compilers are not nearly as good at implementing these
 ### Doubling values of an array example
 
 ```
-struct DoubleUnit : UnrollerUnit<DoubleUnit<int>, int, int> {
+struct DoubleUnit : UnrollerUnit<DoubleUnit, int, int> {
   using TT = ScalableTag<int>;
   inline Vec<TT> Func(ptrdiff_t idx, Vec<TT> x, Vec<TT> y) {
     TT d;
--- a/hwy/contrib/unroller/README.md
+++ b/hwy/contrib/unroller/README.md
@@ -23,7 +23,7 @@ struct DoubleUnit : UnrollerUnit<DoubleUnit<int>, int, int> {
 Leaving all other methods in their default state, the following code will double all the values in array `a` and place them in `r`
 
 ```
-DoubleUnit<a> dblunit;
+DoubleUnit dblunit;
 int r[N];
 Unroller(dblunit, a, r, N);
 ```
--- a/hwy/contrib/unroller/README.md
+++ b/hwy/contrib/unroller/README.md
@@ -1,0 +1,29 @@
+# Unroller
+
+`Unroller` is a templated function that automatically implements common optimizations that are usually handled by compilers when writing scalar code. Modern CPUs operate much more efficiently when non-dependent calculations are packed into an instruction pipeline. For scalar code, this often means a compiler will take a one-line loop, and compile it down to hundreds of lines of machine code in order to fully capture these efficiencies. 
+
+As of today (2023-07-06), compilers are not nearly as good at implementing these optimizations for code written in SIMD intrinsics. `Unroller` is a templated function that takes in an `UnrollerUnit` of SIMD instructions, and then implements unrolling, reordering, hoisting and tail-handling (URHT optimizations) of arrays of data being processed with SIMD intrinsics. 
+
+### `UnrollerUnit`
+
+`UnrollerUnit` and `UnrollerUnit2D` are a base classes of functions that `Unroller` needs implemented in order to properly handle URHT. `UnrollerUnit` has default implementations for all but the `Func` method, which defines the SIMD operation to be applied. Many examples of how to implement these functions are in the tests. 
+
+### Doubling values of an array example
+
+```
+struct DoubleUnit : UnrollerUnit<DoubleUnit<int>, int, int> {
+  using TT = ScalableTag<int>;
+  inline Vec<TT> Func(ptrdiff_t idx, Vec<TT> x, Vec<TT> y) {
+    TT d;
+    return Mul(x, Set(d, 2));
+  }
+};
+```
+
+Leaving all other methods in their default state, the following code will double all the values in array `a` and place them in `r`
+
+```
+DoubleUnit<a> dblunit;
+int r[N];
+Unroller(dblunit, a, r, N);
+```
--- a/hwy/contrib/unroller/unroller-inl.h
+++ b/hwy/contrib/unroller/unroller-inl.h
@@ -18,12 +18,25 @@ namespace HWY_NAMESPACE {
 
 namespace hn = hwy::HWY_NAMESPACE;
 
+// std::min isn't constexpr in c++11, so we use this
+template<typename T> 
+constexpr T constmin(const T a, const T b) {
+    return a < b ? a: b;
+}
+
 template <class DERIVED, typename IN_T, typename OUT_T> 
 struct UnrollerUnit
 {
     DERIVED* me() { return static_cast<DERIVED*>(this); }
-    using IT = hn::ScalableTag<IN_T>;
-    using OT = hn::ScalableTag<OUT_T>;
+    using ITT = hn::ScalableTag<IN_T>;
+    using OTT = hn::ScalableTag<OUT_T>;
+    
+    static constexpr inline size_t lanes() {
+        return constmin(HWY_MAX_LANES_D(ITT), HWY_MAX_LANES_D(OTT));
+    }
+
+    using IT = hn::CappedTag<IN_T, lanes()>;
+    using OT = hn::CappedTag<OUT_T, lanes()>;
     IT d_in;
     OT d_out;
 
@@ -33,12 +46,6 @@ struct UnrollerUnit
 
     inline hn::Vec<OT> func(int idx, hn::Vec<IT> x0, hn::Vec<IT> x1, hn::Vec<OT> y) {
         return me()->func(idx, x0, x1, y);
-    }
-
-    static inline size_t lanes() {
-        const IT d_in;
-        const OT d_out;
-        return std::min(hn::Lanes(d_in), hn::Lanes(d_out));
     }
 
     inline hn::Vec<IT> x0_init()
--- a/hwy/contrib/unroller/unroller-inl.h
+++ b/hwy/contrib/unroller/unroller-inl.h
@@ -86,11 +86,10 @@ struct UnrollerUnit {
                                   const ptrdiff_t places) {
     const hn::ScalableTag<ptrdiff_t> di;
     using TI = hn::TFromD<decltype(di)>;
-    auto mask = hn::RebindMask(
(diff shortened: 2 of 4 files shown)


hwy/contrib/unroller/unroller-inl.h
```diff
@@ -0,0 +1,496 @@
...
+
+  inline hn::Vec<IT> X0Init() { return me()->X0InitImpl(); }
+
+  inline hn::Vec<IT> X0InitImpl() { return hn::Zero(d_in); }
+
+  inline hn::Vec<OT> YInit() { return me()->YInitImpl(); }
+
+  inline hn::Vec<OT> YInitImpl() { return hn::Zero(d_out); }
+
+  inline hn::Vec<IT> Load(ptrdiff_t const& HWY_RESTRICT idx,
```
jan-wassenberg, line 63: Probably easier and faster to pass word-sized variables by value, without const&.

hwy/contrib/unroller/unroller-inl.h
```diff
@@ -0,0 +1,496 @@
...
+                     const ptrdiff_t n) {
+  const auto lane_sz = f.UnitLanes();
+
+  auto xx = f.X0Init();
+  auto yy = f.YInit();
+  ptrdiff_t i = 0;
+
+#if HWY_MEM_OPS_MIGHT_FAULT
+  if (n < lane_sz) {
+    // stack is maybe too small for this in RVV?
```
jan-wassenberg, line 334: We can update this comment: 'this may not fit on the stack for HWY_RVV, but we do not reach this code there'.

hwy/contrib/unroller/unroller_test.cc
```diff
@@ -0,0 +1,295 @@
...
+    if(min > pa[i])
+        min = pa[i];
+  }
+  return min;
+}
+
+template <typename T>
+struct accv : UnrollerUnit<accv<T>, T, T> {
+  using TT = hn::ScalableTag<T>;
+  inline hn::Vec<TT> Func(ptrdiff_t const& HWY_RESTRICT idx,
```
jan-wassenberg, line 66: We can omit the inline, it's implicit for functions defined inside a struct/class. If you want to really force inlining, then we could instead use HWY_INLINE.

hwy/contrib/unroller/unroller_test.cc
```diff
@@ -0,0 +1,295 @@
...
+  T min = std::numeric_limits<T>::max();
+  for (size_t i = 0; i < num; ++i) {
+    if(min > pa[i])
+        min = pa[i];
+  }
+  return min;
+}
+
+template <typename T>
+struct accv : UnrollerUnit<accv<T>, T, T> {
```
jan-wassenberg, line 64: PascalCase please, here and below :) A bit longer name would also help.

hwy/contrib/unroller/unroller_test.cc
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  inline ptrdiff_t MaskStoreImpl(const ptrdiff_t idx, T* HWY_RESTRICT to,
+                                 hn::Vec<TT> const& HWY_RESTRICT x,
+                                 ptrdiff_t const& HWY_RESTRICT places) const {
+    // no stores in a reducer
+    return 0;
+  }
+
+  inline ptrdiff_t ReduceImpl(hn::Vec<TT> const& HWY_RESTRICT x,
```
jan-wassenberg, line 85: Even Vec can be passed around by value (on MSVC that requires /Gv to work well). Then we can/should also remove the HWY_RESTRICT for vectors.

hwy/contrib/unroller/unroller_test.cc
```diff
@@ -0,0 +1,295 @@
...
+  TT d;
+
+  inline hn::Vec<TT> Func(ptrdiff_t const& HWY_RESTRICT idx,
+                          hn::Vec<TT> const& HWY_RESTRICT x,
+                          hn::Vec<TT> const& HWY_RESTRICT y) const {
+    return hn::Min(y, x);
+  }
+
+  inline hn::Vec<TT> YInitImpl() const {
+    return hn::Set(d, std::numeric_limits<T>::max());
```
jan-wassenberg, line 112: If you care about RVV on Godbolt, we can instead use `HighestValue<T>()`. (Some parts of libstdc++ are not available, and might also not work for float16_t, sigh.)

hwy/contrib/unroller/unroller_test.cc
```diff
@@ -0,0 +1,295 @@
...
+  inline hn::Vec<TT> MaskLoadImpl(ptrdiff_t const& HWY_RESTRICT idx,
+                                  T* HWY_RESTRICT from,
+                                  ptrdiff_t const& HWY_RESTRICT places) const {
+    const hn::ScalableTag<ptrdiff_t> di;
+    using TI = hn::TFromD<decltype(di)>;
+    auto mask = hn::FirstN(d, static_cast<size_t>(places));
+    auto maskneg = hn::Not(hn::FirstN(
+        d, static_cast<size_t>(places +
+                               static_cast<ptrdiff_t>(
+                                   UnrollerUnit<minv<T>, T, T>::UnitLanes()))));
```
jan-wassenberg, line 124: Should we encourage a typedef called Base in these classes? Or perhaps a constexpr size_t kMaxLanes at the beginning of the class next to the inheritance?

hwy/contrib/unroller/unroller_test.cc
```diff
@@ -0,0 +1,295 @@
...
+    const hn::ScalableTag<T> d;
+    auto minvect = hn::MinOfLanes(d, x);
+    (*to) = hn::ExtractLane(minvect, 0);
+    return 1;
+  }
+
+  inline void ReduceImpl(hn::Vec<TT> const& HWY_RESTRICT x0,
+                         hn::Vec<TT> const& HWY_RESTRICT x1,
+                         hn::Vec<TT> const& HWY_RESTRICT x2,
+                         hn::Vec<TT>& HWY_RESTRICT y) const {
```
jan-wassenberg, line 157: I think MSVC might complain about restrict on references. A pointer would be safer for the output arg.

hwy/contrib/unroller/unroller_test.cc
```diff
@@ -0,0 +1,295 @@
...
+    const hn::ScalableTag<T> d;
+    (*to) = hn::ReduceSum(d, x);
+    return 1;
+  }
+
+  inline void ReduceImpl(hn::Vec<TT> const& HWY_RESTRICT x0,
+                         hn::Vec<TT> const& HWY_RESTRICT x1,
+                         hn::Vec<TT> const& HWY_RESTRICT x2,
+                         hn::Vec<TT>& HWY_RESTRICT y) const {
+    y = y + x0 + x1 + x2;
```
jan-wassenberg, line 200: To work on SVE/RVV, we'd have to use Add() here :/

hwy/contrib/unroller/unroller_test.cc
```diff
@@ -0,0 +1,295 @@
...
+  }
+};
+
+template <typename T>
```
jan-wassenberg, line 204: Let's move this to test_util.h?

hwy/contrib/unroller/unroller-inl.h
```diff
@@ -0,0 +1,496 @@
...
+    xx = f.MaskLoad(n - lane_sz, x, i - n);
+    yy = f.Func(n - lane_sz, xx, yy);
+    f.MaskStore(n - lane_sz, y, yy, i - n);
+  }
+
+  f.Reduce(yy, y);
+}
+
+template <class FUNC, typename IN0_T, typename IN1_T, typename OUT_T>
+inline void Unroller(FUNC& HWY_RESTRICT f, IN0_T* HWY_RESTRICT x0,
```
jan-wassenberg, line 401: Are we expecting FUNC to record any state? If not, it would be convenient for callers if they could pass FUNC(), which would require the f arg here to be const&.

Also, if the reducer always returns one value, can we have that returned from this function? That would allow `const T result = Unroller();` rather than `T result; Unroller(.., &result)`.

jan-wassenberg (changes requested):
Nice, great to see a concrete example :)

jan-wassenberg:
The checks/builders are now activated here so we should be able to [see warnings/errors here](https://github.com/google/highway/actions/runs/5416172582/jobs/9863077600) also in case that helps.

Generally we aim to have clean builds. How to debug them? Often if there is a flood, I find it useful to sort the output, keep only unique lines (SublimeText editor can do this), then often there are only a few root causes to look at.

Happy to advise further if you have examples? The current ones seem to just be unused-arg warnings, we (void)arg those.

hwy/contrib/unroller/unroller-inl.h
matthewkolbe: > Are we expecting FUNC to record any state?

I don't think it will record state. I'm having difficulty thinking of a reason it would. It can contain state, but I don't think there's a reason to allow it to alter state. I was trying to do this a couple days ago, and I encountered a problem. I don't exactly remember what the problem was, but I had to make all the member functions `const`, and I vaguely remember something like the `D d` member variables of the class being passed into non-const highway functions was causing the builds to fail. Does that sound right?

> if the reducer always returns one value

It returns one value in the examples I made (I'll make more), but it returns `void` in non-reducer loops.

hwy/contrib/unroller/unroller_test.cc
matthewkolbe: I don't quite understand how to do this yet.

hwy/contrib/unroller/unroller_test.cc
matthewkolbe: I'm thinking about the right way to handle this.

matthewkolbe:
@jan-wassenberg I'm heading into nature for a week. For every one of your change requests, I did a thumbs up if it's now integrated into the head, or I gave a comment as to why it's not in there. I'll check back in with you when I'm back.

hwy/contrib/unroller/unroller_test.cc
jan-wassenberg: Ah, was just thinking: this pair of functions is now used in multiple tests. We can avoid the duplication by moving both of them to test_util.h, and adding HWY_INLINE before the void just to be sure.

matthewkolbe:
@jan-wassenberg I notice it's failing to build on clang 6.0. I'm not seeing the same issue on Godbolt: https://godbolt.org/z/PsY3Wsfex

I don't have clang 6.0, so I cannot test locally.

hwy/contrib/unroller/unroller_test.cc
```diff
@@ -88,6 +88,77 @@ struct ConvertUnit : UnrollerUnit<ConvertUnit<FROM_T, TO_T>, FROM_T, TO_T> {
...
 
+template <typename T>
+struct FindUnit : UnrollerUnit<FindUnit<T>, T, ptrdiff_t> {
+  static constexpr size_t UnitLanes() {
+    return HWY_MIN(HWY_MAX_LANES_D(hn::ScalableTag<T>),
+                   HWY_MAX_LANES_D(hn::ScalableTag<ptrdiff_t>));
+  }
+
+  using TT = hn::CappedTag<T, UnitLanes()>;
+  using TTSZT = hn::CappedTag<ptrdiff_t, UnitLanes()>;
```
jan-wassenberg, line 99: Strictly speaking there is no guarantee this lane type is supported. How about `MakeSigned<sizeof(ptrdiff_t)>`?

hwy/contrib/unroller/README.md
```diff
@@ -23,7 +23,7 @@ struct DoubleUnit : UnrollerUnit<DoubleUnit<int>, int, int> {
 Leaving all other methods in their default state, the following code will double all the values in array `a` and place them in `r`
```
jan-wassenberg, line 23: How about we add a comment that this interface is considered experimental and may still change? Then we can commit them already and gather practical experience with it, while still allowing later improvements.

jan-wassenberg:
Looks like clang 6 is now working? :) Nice.

jan-wassenberg (changes requested):
👍

hwy/contrib/unroller/unroller_test.cc
matthewkolbe: I used `SignedFromSize` instead, but otherwise this was integrated in.

jan-wassenberg: Oops, that is what I meant 👍

jan-wassenberg:
Nice, thanks for making the changes!

jan-wassenberg (approved):

PR #1515, opened 2023-06-30
Use blend instead of mov for IfThenElse operation on SSE/AVX2/AVX512

ag01:


jan-wassenberg (approved):
Thanks for proposing and implementing this optimization :)

PR #1518, opened 2023-07-05
Remove a simple warning on GNU/Hurd system (non-linux)

malaterre:
Fixes:

  hwy/contrib/sort/vqsort.cc:82:30: error: unused parameter ‘bytes’ [-Werror=unused-parameter]

Also fixes a tiny typo in the doc.


jan-wassenberg (approved):
Thank you!

jan-wassenberg:
The idea is to skip the secure seeding on RVV because we cannot even include features.h there. If both RVV and Linux, I think the RVV will 'win' by defining VQSORT_GETRANDOM, so we skip the branch below.
Are you seeing any problems related to this?

jan-wassenberg:
I see, thanks. Looks like Godbolt does not have HWY_OS_LINUX set. Thus we can change the check to RVV && !HWY_OS_LINUX - I will do so :)

PR #1530, opened 2023-07-06
Zero: use __m512i version of setzero for int based floats

1480c1:
Fixes a compiler error of being unable to convert __m512 to long long int


jan-wassenberg (approved):
Oops, right, thanks for fixing this :)
I've now double-checked that x86_128/256 do this correctly.

PR #1540, opened 2023-07-10
Use defined marco to determine whether to use HWY_RVV_AVOID_VXRM

bhbruce:
The origin purpose of `HWY_RVV_AVOID_VXRM` is to detect whether VXRM is embedded in an intrinsic function argument.
If not embedded into an argument, `HWY_RVV_AVOID_VXRM` is defined.

This PR changes to use a predefined macro to check to make it available for both clang & gcc.

### DIFF
--- a/hwy/ops/rvv-inl.h
+++ b/hwy/ops/rvv-inl.h
@@ -943,8 +943,8 @@ HWY_RVV_FOREACH_U(HWY_RVV_RETV_ARGVV, SaturatedSub, ssubu, _ALL)
 HWY_RVV_FOREACH_I(HWY_RVV_RETV_ARGVV, SaturatedSub, ssub, _ALL)
 
 // ------------------------------ AverageRound
-// Assume that GCC-13 default to legacy vxrm arguments. Tested with GCC 13.1.0
-#if HWY_COMPILER_GCC_ACTUAL && HWY_COMPILER_GCC_ACTUAL < 1400
+// VXMR is embedded into intrinsic function argument since intrinsic v1.0.
+#if __riscv_v_intrinsic == 11000
 #define HWY_RVV_AVOID_VXRM
 #endif
 


jan-wassenberg (changes requested):
Hi @bhbruce , thanks for suggesting this change!
Unfortunately we have __riscv_v_intrinsic = 11000 on LLVM trunk, but it still requires the new argument, so we must not define AVOID.

Is there any other way to detect this?

Perhaps one alternative would be to change the AVOID macro into a tri-state, where undefined = default (= avoid on GCC<14), 1=always avoid, 0=never avoid?

jan-wassenberg:
We could use HWY_COMPILER_CLANG (which is still correct in the case of Apple LLVM).
Is your proposal: HWY_COMPILER_CLANG || __riscv_v_intrinsic < 11000: add arg; else AVOID? I think that could work.

bhbruce:
> Is your proposal: HWY_COMPILER_CLANG || __riscv_v_intrinsic < 11000: add arg; else AVOID? I think that could work.
I am not sure highway library support v0.10 intrinsic or not.

My proposal may look like this one.
```
#if HWY_COMPILER_CLANG == 1600 && __riscv_v_intrinsic == 11000 
    #define HWY_RVV_AVOID_VXRM
#elif HWY_COMPILER_GCC_ACTUAL && HWY_COMPILER_GCC_ACTUAL < 1400
    #define HWY_RVV_AVOID_VXRM
#end
```

jan-wassenberg:
Sorry to go back and forth :) There is still an issue because LLVM trunk is detected as 1600, and has intrinsic version 11000, and requires AVOID not to be set.
What we can do is AVOID on earlier clang than 1600, or intrinsics < 11000, and keep the same logic you have for GCC.
I'll send a change for this :)

PR #1552, opened 2023-07-11
Enable inlining of DoTestSlideDownLanes on x86/NEON/PPC/WASM targets

johnplatts:
Updated slide_up_down_test.cc to enable inlining of the DoTestSlideDownLanes on SSE2/SSSE3/SSE4/AVX2/AVX3/NEON/PPC/WASM targets as DoTestSlideDownLanes needs to be inlined on the SSE2/SSSE3/SSE4/AVX2/AVX3/NEON/WASM targets to ensure thorough testing of the SlideDownLanes implementation on SSE2/SSSE3/SSE4/AVX2/AVX3/NEON/WASM.

DoTestSlideDownLanes should also not be inlined on the RVV target to work around miscompilation on the RVV target.

The HWY_SLIDE_DOWN_TEST_INLINE define was added to slide_up_down_test.cc to enable inlining of DoTestSlideDownLanes on SSE2/SSSE3/SSE4/AVX2/AVX3/NEON/PPC/WASM targets but disable inlining of DoTestSlideDownLanes on other targets (including RVV where inlining of DoTestSlideDownLanes should be disabled).


jan-wassenberg (dismissed):
👍

PR #1560, opened 2023-07-13
Fixed compilation errors with unroller_test on Clang

johnplatts:
Updated unroller-inl.h and unroller_test.cc to fix compilation errors with the Highway test suite on clang.


jan-wassenberg (approved):
Thank you for fixing these :)

PR #1564, opened 2023-07-13
Fix unroller_test FindUnit

johnplatts:
Fixes bug in FindUnit in unroller_test.cc that causes UnrollerTest to fail on AVX2


jan-wassenberg (approved):
Nice improvements, and sorry I did not catch the "Vec<> inside class" during code review :)

PR #1568, opened 2023-07-17
Add Set2/Set3/Set4 functions as new tuple operations

ag01:


jan-wassenberg (approved):
Looks good, thanks for adding this!

PR #1576, opened 2023-07-19
Fix TernaryLogic inline assembly in ppc_vsx-inl.h

johnplatts:
Updated GCC PPC10 TernaryLogic inline assembly workaround in ppc_vsx-inl.h to prevent kTernLogOp from being incorrectly sign-extended if kTernLogOp >= 128 is true.


jan-wassenberg (approved):
👍

PR #1578, opened 2023-07-19
Made enhancements to x86 ExtractLane operation

johnplatts:
Made the following enhancements to ExtractLane on x86:
- Updated implementation of detail::ExtractLane in hwy/ops/x86_128-inl.h for I32/U32/F32 vectors on SSE2/SSSE3 targets to do an shuffle followed by a _mm_cvtsi128_si32 or _mm_cvtss_f32
- Updated AVX2/AVX3 implementation of ExtractLane for 32-byte and 64-byte vectors to call ExtractLane on the extracted 16-byte block if `__builtin_constant_p(i < kLanesPerBlock) && (i < kLanesPerBlock)` is true


jan-wassenberg (approved):
Nice :)

PR #1579, opened 2023-07-20
Replaced memcpy with 3-argument CopyBytes

johnplatts:
Added a new 3-argument CopyBytes function in hwy/base.h that takes an additional num_of_bytes_to_copy argument.

Also replaced `memcpy(to, from, N)` calls with `CopyBytes(from, to, N)` calls (where `N` is the number of bytes to copy) in hwy/contrib/sort/algo-inl.h, hwy/contrib/sort/vqsort-inl.h, hwy/contrib/unroller/unroller-inl.h, hwy/ops/emu128-inl.h, 
hwy/ops/x86_128-inl.h, and hwy/ops/x86_256-inl.h.

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -282,6 +282,15 @@ HWY_API void CopyBytes(const From* from, To* to) {
 #endif
 }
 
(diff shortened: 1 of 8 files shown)


jan-wassenberg (approved):
👍

PR #1585, opened 2023-07-21
Applied code reformatting using clang-format

johnplatts:
Reformatted Highway source code using clang-format


jan-wassenberg (approved):
Thanks! FYI our clang-format runs only on files we edit, thus there may be some differences.
I also am not 100% certain we have the same .clang-format file as our internal clang-format. Let's see if any of this changes back.

PR #1586, opened 2023-07-21
Fixed UI64 ScalarAbsDiff in TestIntegerAbsDiff

johnplatts:
Updated UI64 ScalarAbsDiff in TestIntegerAbsDiff to cast its operands to uint64_t before subtraction as subtraction overflow in int64_t ScalarAbsDiff is possible on the RVV target if `Lanes(ScalableTag<int64_t>()) >= 8` is true (which implies that `Lanes(ScalableTag<int64_t, 3>()) >= 64` is true on the RVV target).


jan-wassenberg (approved):
Thanks for fixing this :)

PR #1588, opened 2023-07-21
Made fixes to TestInsertLane and TestExtractLane

johnplatts:
Fixed TestInsertLane as follows:
- Added DoTestInsertLaneWithConstAmt to verify correctness of InsertLane that is called with a constant amount for vectors that are 16 bytes or smaller on x86/NEON/PPC/WASM targets
- Added a loop that starts with a zero vector and inserts each value one lane at a time
- Added a check to make sure that `v` is equal to `v2` after the first loop
- Updated the InsertLane operation in the second loop (which was previously the only loop in TestInsertLane) to use the reloaded `lanes` vector instead of `v`

Also added extra checks to TestExtractLane to verify correctness of ExtractLane on x86/NEON/PPC/WASM targets when called with a constant amount that is less than `HWY_MIN(Lanes(d), 16 / sizeof(TFromD<D>))`.

### DIFF
--- a/hwy/tests/swizzle_test.cc
+++ b/hwy/tests/swizzle_test.cc
@@ -36,9 +36,68 @@ HWY_NOINLINE void TestAllGetLane() {
 }
 
 struct TestExtractLane {
+#if !HWY_HAVE_SCALABLE && HWY_TARGET < HWY_EMU128 && \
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
👍

jan-wassenberg (approved):

jan-wassenberg:
Apologies, the import process somehow got stuck. I've restarted it.

Issue #1570, opened 2023-07-17
Build error when cross-building for Aarch64 (ARM64) with GCC 12.3 (from Yocto Mickledore 4.2): hwy/base.h:1155:16: error: inlining failed in call to 'always_inline' 'size_t hwy::Num0BitsAboveMS1Bit_Nonzero32(uint32_t)': target specific option mismatch


clopez:
I tried both with last stable release 1.0.4 (46e365d6770f5d7a4240d8ac9d8e928a520478ea) and with master as of today (7233df1b4e29a04ecfd3a10a1da14c802f08c3fd)

And in both cases I get this build error:

```
[  5%] Building CXX object CMakeFiles/hwy_contrib.dir/hwy/contrib/sort/vqsort_128a.cc.o
Reaping winning child 0x557a0072df70 PID 88468 
[...]

Issue #1460, opened 2023-06-15
vqsort_128a.cc.o build error for aarch64 with gcc


Ray-V:
Built to commit f5f35e144c6e45beeb3caa28c6b1ee460705a86b
```
CXX=aarch64-linux-gnu-g++  cmake -DBUILD_TESTING=no -DBUILD_SHARED_LIBS=yes ..
make

..
[ 18%] Linking CXX shared library libhwy.so
[ 18%] Built target hwy
[ 21%] Building CXX object CMakeFiles/hwy_contrib.dir/hwy/contrib/sort/vqsort_128a.cc.o
In file included from highway/hwy/contrib/sort/vqsort_128a.cc:23,
                 from highway/hwy/foreach_target.h:152,
                 from highway/hwy/contrib/sort/vqsort_128a.cc:20:
highway/hwy/contrib/sort/traits128-inl.h: In member function ‘hwy::N_NEON::Vec<D> hwy::N_NEON::detail::Traits128<Base>::SortPairsReverse4(D, hwy::N_NEON::Vec<D>) const’:
highway/hwy/contrib/sort/traits128-inl.h:508:36: warning: requested alignment 64 is larger than 16 [-Wattributes]
  508 |     alignas(64) uint64_t kIndices[8] = {7, 7, 5, 5, 5, 5, 7, 7};
      |                                    ^
In file included from highway/hwy/contrib/sort/vqsort_128a.cc:23:
highway/hwy/contrib/sort/traits128-inl.h: In member function ‘hwy::N_NEON_WITHOUT_AES::Vec<D> hwy::N_NEON_WITHOUT_AES::detail::Traits128<Base>::SortPairsReverse4(D, hwy::N_NEON_WITHOUT_AES::Vec<D>) const’:
highway/hwy/contrib/sort/traits128-inl.h:508:36: warning: requested alignment 64 is larger than 16 [-Wattributes]
  508 |     alignas(64) uint64_t kIndices[8] = {7, 7, 5, 5, 5, 5, 7, 7};
      |                                    ^
In file included from highway/hwy/contrib/sort/vqsort.h:27,
                 from highway/hwy/contrib/sort/vqsort_128a.cc:16:
[...]

PR #1589, opened 2023-07-22
HWY_NEON: Enable the crypto extension for GCC without setting the full arch

clopez:
This reverts commits 60241b99 and f8a6ea58.

GCC understands (like Clang) appending an architecture extension on its own. This can be used to turn on and off particular architectural extensions without having to specify a particular architecture version or core.

GCC supports this since at least 6.5. Check:
https://gcc.gnu.org/onlinedocs/gcc-6.5.0/gcc/AArch64-Function-Attributes.html
https://gcc.gnu.org/onlinedocs/gcc/AArch64-Function-Attributes.html

This way we can enable the crypto architecture extension for `HWY_NEON` without overriding the default architecture of the compiler potentially disabling other architectural extensions.

This fixes a build error that happens with the RPi4 and a Yocto-based toolchain that by default builds with `-march=armv8-a+crc` (the RPi4 has no crypto arch extension).
So before this commit we ended building `HWY_NEON` with `-march=armv8-a+crypto` and that failed to build when it tried to inline code from other parts that by default is compiled with `-march=armv8-a+crc` (due to the toolchain in this example).

The solution is to append the crypto extension, so in this example of the RPi4 we end compiling `HWY_NEON` with `-march=armv8-a+crc+crypto` allowing it to inline code from any other part that is built with `-march=armv8-a+crc`. See the documentation from GCC for the `Inlining rules` in the links from above.

Issues related: #1570 and #1460.

### DIFF
--- a/hwy/ops/set_macros-inl.h
+++ b/hwy/ops/set_macros-inl.h
@@ -360,11 +360,7 @@
 #if HWY_TARGET == HWY_NEON_WITHOUT_AES
 // Do not define HWY_TARGET_STR (no pragma).
 #else
-#if HWY_COMPILER_GCC_ACTUAL
-#define HWY_TARGET_STR "arch=armv8-a+crypto"
-#else  // clang
 #define HWY_TARGET_STR "+crypto"
-#endif  // HWY_COMPILER_*
 #endif  // HWY_TARGET == HWY_NEON_WITHOUT_AES
 
 #endif  // HWY_ARCH_ARM_V7


jan-wassenberg (approved):
Thank you for sending the pull request!
Unfortunately I do not remember, nor does the commit message mention, why my change in March added GCC-specific flags. From a [quick test](https://gcc.godbolt.org/z/o341o4cdf), this does indeed work for all GCC versions tested. 👍

PR #1592, opened 2023-07-24
Made fixes and enhancements to IfNegativeThenElse

johnplatts:
Made the following bug fixes to IfNegativeThenElse:
- Updated SSE2/SSSE3 implementations to broadcast the sign bit
- Updated EMU128/SCALAR implementations to use `vi` (which is `v` bitcasted to `Vec<RebindToSigned<DFromV<decltype(v)>>>`) to match the behavior of x86/NEON/PPC IfNegativeThenElse

Made the following enhancements to IfNegativeThenElse:
- Updated RVV/SVE IfNegativeThenElse to avoid broadcasting the sign bit and simply do a signed integer less-than comparison (as RVV/SVE MaskFromVec simply does a comparison against zero)
- Updated PPC10 IfNegativeThenElse to do a vec_blend (which has the same behavior as SSE4 _mm_blendv_epi8, _mm_blendv_ps, and _mm_blend_pd)

Also updated TestIfNegative to add additional checks that would expose the bugs that were previously there with the SSE2/SSSE3 IfNegativeThenElse implementation.

Also updated I32/I64 SaturatedAdd and SaturatedSub implementations in ppc_vsx-inl.h and generic_ops-inl.h and to use IfNegativeThenElse instead of IfThenElse.

### DIFF
--- a/hwy/ops/arm_sve-inl.h
+++ b/hwy/ops/arm_sve-inl.h
@@ -3566,7 +3566,7 @@ HWY_API V IfNegativeThenElse(V v, V yes, V no) {
   const DFromV<V> d;
   const RebindToSigned<decltype(d)> di;
 
-  const svbool_t m = MaskFromVec(BitCast(d, BroadcastSignBit(BitCast(di, v))));
+  const svbool_t m = detail::LtN(BitCast(di, v), 0);
   return IfThenElse(m, yes, no);
 }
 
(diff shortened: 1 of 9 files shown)


jan-wassenberg (approved):
Thanks for fixing the warnings :)

jan-wassenberg:
Nice codegen improvement also for SaturatedAdd :)

PR #1604, opened 2023-07-26
Fix arm64 build on Windows with Visual Studio.

dlemstra:
We are using this project in an arm64 build of the @ImageMagick library and this no longer works with the latest release. This is happening because `HWY_NEON_HAVE_FLOAT16C` becomes `1` while it shouldn't. And it also looks like the current code breaks the next check `(HWY_SVE_HAVE_BFLOAT16)` because the first one will always be true on arm64. This pull request resolves that issue.

p.s. Would you be interested in a Visual Studio x86?/x64/arm64 build to make sure that this project continues to work on the Windows platform?

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -340,7 +340,7 @@ static constexpr HWY_MAYBE_UNUSED size_t kMaxVectorSize = 16;
 
 // float16_t load/store/conversion intrinsics are always supported on Armv8 and
 // VFPv4. On Armv7 Clang requires __ARM_FP & 2; GCC requires -mfp16-format=ieee.
-#if HWY_ARCH_ARM_A64 ||                                            \
+#if HWY_ARCH_ARM_A64 &&                                            \
     (HWY_COMPILER_CLANG && defined(__ARM_FP) && (__ARM_FP & 2)) || \
     HWY_COMPILER_GCC_ACTUAL && defined(__ARM_FP16_FORMAT_IEEE)
 #define HWY_NEON_HAVE_FLOAT16C 1


jan-wassenberg:
Hi @dlemstra , thanks for reaching out and letting us know. Can you help me understand why the FCVTN instruction is not available on Armv7? The [intrinsics guide](https://developer.arm.com/architectures/instruction-sets/intrinsics/#f:@navigationhierarchiessimdisa=[Neon]&q=vcvt_f16) claims it should be. My understanding is that F16C is always available on aarch64, and only conditionally on Armv7. Perhaps our checks for that in the next two lines are too optimistic?

> Would you be interested in a Visual Studio x86?/x64/arm64 build to make sure that this project continues to work on the Windows platform?

Absolutely, thanks. What lines should we add to .github/workflows/build_test?

dlemstra:
I completely missed that comment and I should have taken a bit more time investigating this before creating the PR. I could not find anything in the MSDN documentaton of Microsoft about `__fp16` so this doesn't seem to be available on that platform. Maybe we could change this PR to this:

```c
#if (HWY_ARCH_ARM_A64 && !HWY_COMPILER_MSVC) ||                    \
```

jan-wassenberg:
:) Oh, that makes sense - looks like MSVC doesn't completely implement the ACLE interface.
Sure, I'm happy to make your proposed change.
Sounds good with the workflow, thanks for looking into it!

dlemstra:
I would not have minded to make that pull request. Will close this now.

PR #1620, opened 2023-07-31
Added LoadN and StoreN operations

johnplatts:
Added the LoadN operation to allow only the first (lowest-indexed) `HWY_MIN(Lanes(d), max_lanes_to_load)` lanes to be loaded without faulting.

Also added the StoreN operation to allow only the first (lowest-indexed) `HWY_MIN(Lanes(d), max_lanes_to_store)` lanes to be stored without faulting or modifying any memory past `p + HWY_MIN(Lanes(d), max_lanes_to_store) - 1`.

### DIFF
--- a/hwy/tests/memory_test.cc
+++ b/hwy/tests/memory_test.cc
@@ -324,6 +324,254 @@ HWY_NOINLINE void TestAllCache() {
   Pause();
 }
 
+class TestLoadN {
+ private:
+  template <class T, HWY_IF_NOT_FLOAT_NOR_SPECIAL(T)>
+  static HWY_INLINE T GenerateNonZeroValue(size_t val) {
+    const T conv_val = static_cast<T>(val);
+    return (conv_val == 0) ? static_cast<T>(-17) : conv_val;
+  }
+  template <class T, HWY_IF_FLOAT3264(T)>
+  static HWY_INLINE T GenerateNonZeroValue(size_t val) {
+    const T flt_val = static_cast<T>(val);
+    return (flt_val == T{0} ? static_cast<T>(0.5426808228865735) : flt_val);
+  }
+  template <class T, hwy::EnableIf<IsSame<T, hwy::bfloat16_t>()>* = nullptr>
+  static HWY_INLINE T GenerateNonZeroValue(size_t val) {
+    return BF16FromF32(GenerateNonZeroValue<float>(val));
+  }
+  template <class T, hwy::EnableIf<IsSame<T, hwy::float16_t>()>* = nullptr>
+  static HWY_INLINE T GenerateNonZeroValue(size_t val) {
+    return F16FromF32(GenerateNonZeroValue<float>(val));
+  }
+
+ public:
+  template <class T, class D>
+  HWY_NOINLINE void operator()(T /*unused*/, D d) {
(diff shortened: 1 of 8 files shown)


hwy/tests/memory_test.cc
```diff
@@ -324,6 +324,254 @@ HWY_NOINLINE void TestAllCache() {
...
+
+    auto load_buf = AllocateAligned<T>(load_buf_len);
+    auto expected = AllocateAligned<T>(N);
+    HWY_ASSERT(load_buf && expected);
+
+    for (size_t i = 0; i < load_buf_len; i++) {
+      load_buf[i] = GenerateNonZeroValue<T>(i + 1);
+    }
+
+    std::fill(expected.get(), expected.get() + N, T(0));
```
jan-wassenberg, line 367: We have problems with libc++ on some platforms. In future, let's use ZeroBytes instead. No need to change now, I will replace it when it becomes necessary.

jan-wassenberg (dismissed):
Nice, very useful, thanks for adding this!

jan-wassenberg:
We're seeing MSVC AVX2 fail:
```
[ RUN      ] HwyMemoryTestGroup/HwyMemoryTest.TestAllStoreN/AVX2
i8x132 expect [33+ ->]:
  0x04,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,
i8x132 actual [33+ ->]:
  0x04,0xFF,0x02,0xFF,0xFF,0xFF,0xFF,
Abort at hwy/tests/memory_test.cc:537: AVX2, i8x132 lane 35 mismatch: expected '0xFF', got '0x02'.
```
The array length 132 appears legitimate (32+3*32+4). Any thoughts on what might be causing this or a workaround?

johnplatts:
> We're seeing MSVC AVX2 fail:
> 
> ```
> [ RUN      ] HwyMemoryTestGroup/HwyMemoryTest.TestAllStoreN/AVX2
> i8x132 expect [33+ ->]:
>   0x04,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,
> i8x132 actual [33+ ->]:
>   0x04,0xFF,0x02,0xFF,0xFF,0xFF,0xFF,
> Abort at hwy/tests/memory_test.cc:537: AVX2, i8x132 lane 35 mismatch: expected '0xFF', got '0x02'.
> ```
> 
> The array length 132 appears legitimate (32+3*32+4). Any thoughts on what might be causing this or a workaround?

I have added a HWY_FENCE before and after the stores in the AVX2 StoreN implementation when compiled with MSVC, and adding HWY_FENCE before and after the stores did fix the problem.

HwyMemoryTestGroup/HwyMemoryTest.TestAllStoreN/AVX2 now passes when compiled with MSVC with the HWY_FENCE fix.

jan-wassenberg (approved):
Nice that you found a good workaround 👍 Am curious how you discovered that a fence would be required? I don't remember encountering such a bug on MSVC before.

johnplatts:
> Nice that you found a good workaround 👍 Am curious how you discovered that a fence would be required? I don't remember encountering such a bug on MSVC before.

I discovered that a fence would be required through trial and error.

jan-wassenberg:
Impressive :)

PR #1622, opened 2023-08-01
Added SatWidenMulPairwiseAdd

johnplatts:
Added U8xI8 SatWidenMulPairwiseAdd operation, which is a saturated pairwise sum of two adjacent U8xI8 products.

SSSE3/SSE4/AVX2/AVX3 can do an U8xI8 SatWidenMulPairwiseAdd operation using a single pmaddubsw instruction.

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -604,6 +604,13 @@ All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
     <code>Vec&lt;D&gt; **WidenMulPairwiseAdd**(D d, V a, V b,)</code>: widens `a`
     and `b` to `TFromD<D>` and computes `a[2*i+1]*b[2*i+1] + a[2*i+0]*b[2*i+0]`.
 
+*   `V`: `u8`, `VU`: `Vec<RebindToUnsigned<DFromV<V>>>`,
+    `D`: `RepartitionToWide<DFromV<V>>` \
+    <code>Vec&lt;D&gt; **SatWidenMulPairwiseAdd**(D d, VU a, V b)</code>:
+    widens `a` and `b` to `TFromD<D>` and computes
+    `a[2*i+1]*b[2*i+1] + a[2*i+0]*b[2*i+0]`, saturated to the range of
+    `TFromD<D>`.
+
 *   `V`: `{bf,i}16`, `D`: `RepartitionToWide<DFromV<V>>`, `VW`: `Vec<D>` \
     <code>VW **ReorderWidenMulAccumulate**(D d, V a, V b, VW sum0, VW&
     sum1)</code>: widens `a` and `b` to `TFromD<D>`, then adds `a[i] * b[i]` to
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -604,11 +604,11 @@ All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
     <code>Vec&lt;D&gt; **WidenMulPairwiseAdd**(D d, V a, V b,)</code>: widens `a`
     and `b` to `TFromD<D>` and computes `a[2*i+1]*b[2*i+1] + a[2*i+0]*b[2*i+0]`.
 
-*   `V`: `u8`, `VU`: `Vec<RebindToUnsigned<DFromV<V>>>`,
-    `D`: `RepartitionToWide<DFromV<V>>` \
-    <code>Vec&lt;D&gt; **SatWidenMulPairwiseAdd**(D d, VU a, V b)</code>:
-    widens `a` and `b` to `TFromD<D>` and computes
-    `a[2*i+1]*b[2*i+1] + a[2*i+0]*b[2*i+0]`, saturated to the range of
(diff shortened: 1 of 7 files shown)


g3doc/quick_reference.md
```diff
@@ -604,6 +604,13 @@ All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
     <code>Vec&lt;D&gt; **WidenMulPairwiseAdd**(D d, V a, V b,)</code>: widens `a`
     and `b` to `TFromD<D>` and computes `a[2*i+1]*b[2*i+1] + a[2*i+0]*b[2*i+0]`.
 
+*   `V`: `u8`, `VU`: `Vec<RebindToUnsigned<DFromV<V>>>`,
```
jan-wassenberg, line 607: I think V is i8 here. Let's also rename V to VI and D to DI to make this more obvious?
Maybe also in the parameter names, such as au/bi or a_u and b_i?

hwy/ops/scalar-inl.h
```diff
@@ -1784,6 +1784,21 @@ HWY_API Vec1<int32_t> WidenMulPairwiseAdd(D32 /* tag */, Vec1<int16_t> a,
...
+#ifdef HWY_NATIVE_U8_I8_SATWIDENMULPAIRWISEADD
+#undef HWY_NATIVE_U8_I8_SATWIDENMULPAIRWISEADD
+#else
+#define HWY_NATIVE_U8_I8_SATWIDENMULPAIRWISEADD
+#endif
+
+template <class DI16, HWY_IF_I16_D(DI16)>
+HWY_API Vec1<int16_t> SatWidenMulPairwiseAdd(DI16 /* tag */, Vec1<uint8_t> a,
+                                             Vec1<int8_t> b) {
+  return Vec1<int16_t>(static_cast<int16_t>(a.raw) *
```
jan-wassenberg, line 1798: I think we'd have to add saturation here before passing to the Vec1 ctor. Or perhaps just remove this specialization and let generic_ops handle it?

hwy/ops/x86_256-inl.h
```diff
@@ -5307,6 +5307,16 @@ HWY_API Vec256<int32_t> WidenMulPairwiseAdd(D /*d32*/, Vec256<int16_t> a,
   return Vec256<int32_t>{_mm256_madd_epi16(a.raw, b.raw)};
 }
 
+// ------------------------------ SatWidenMulPairwiseAdd
+
+template <class DI16, class VU8, class VI8, HWY_IF_I16_D(DI16),
+          HWY_IF_V_SIZE_D(DI16, 32), HWY_IF_U8_D(DFromV<VU8>),
+          HWY_IF_I8_D(DFromV<VI8>), HWY_IF_LANES_D(DFromV<VU8>, 32),
```
jan-wassenberg, line 5314: Here and in the other x86*, if you want, we could skip the HWY_IF_U8 and HWY_IF_LANES by deriving the a/b arg types from DI16. The code would simplify to
```
template <class DI16, HWY_IF_I16_D(DI16), HWY_IF_V_SIZE_D(DI16, 32)>
HWY_API VFromD<DI16> SatWidenMulPairwiseAdd(DI16 /* tag */,
    VFromD<Repartition<uint8_t, DI16>> a, VFromD<Repartition<int8_t, DI16>> b) {
  return VFromD<DI16>{_mm256_maddubs_epi16(a.raw, b.raw)};
}
```

jan-wassenberg (changes requested):
Nice to provide this missing op that some x86 code uses. Thanks for adding it!
Some suggestions:

hwy/ops/scalar-inl.h
johnplatts: If `a` is a 1-lane uint8_t vector and `b` is a 1-lane int8_t vector, `a[0] * b[0]` is within the range of an int16_t as `a[0] * b[0]` is between -32640 and 32385.

`a[0] * b[0] + a[1] * b[1]` can only overflow if `a[0]`, `a[1]`, `b[0]`, and `b[1]` are all non-zero and `b[0]` and `b[1]` both have the same sign.

jan-wassenberg: Oh, I see; it is only the addition that can overflow, and it doesn't happen in the scalar target. Thanks for explaining and adding the helpful comment 👍

jan-wassenberg (approved):

jan-wassenberg:
Sorry about the delay, I've checked up on our internal review, should be moving forward now.

johnplatts:
> Looks like this is hitting the generic implementation for 8x8=16 bit. It's not clear why this would fail only on PPC. The ops involved are SaturatedAdd which is just vec_adds, and shifts/and/mul. Any idea what's going on?

I have compiled this against GCC 11, GCC 12, GCC 13.2, and Clang 16, and TestAllSatWidenMulPairwiseAdd passes on PPC9 when compiled with GCC 11 or Clang 16 but fails when compiled with GCC 12 or GCC 13.2.

jan-wassenberg:
Thanks for testing and confirming :) Any thoughts on whether we should disable the test for GCC12/13, or mark PPC9 as broken for those compilers?

jan-wassenberg:
Great, thank you :D Looks like a more direct pre-computation of the expected values 👍

PR #1635, opened 2023-08-05
Fix GCC10 compilation errors with x86 Load/Store operations

johnplatts:
Fix compilation errors with x86 Load/Store when compiled with GCC 10.

Also fixed AVX2/AVX3 F16 LoadU operation to invoke the U16 LoadU operation instead of the U16 Load operation.


jan-wassenberg (approved):
Thank you for fixing these :) Some are also included in #1633 but that is not yet reviewed, so I will update that if required.

PR #1636, opened 2023-08-06
Fix return type of MulOdd

osamu620:
I believe the return type of MulOdd shall be V2.


osamu620:
Oops! I missed that "V2 is a vector with double-width lanes, or the same as V for 64-bit inputs."

jan-wassenberg:
Ah yes. Thank you for pointing to this, always happy to have a look. The return type here is a bit tricky but working as intended.

PR #1638, opened 2023-08-07
Fixed PPC9 LoadN compilation error with older GCC/Clang releases

johnplatts:
Fixed compilation error in PPC9 LoadN with GCC 9 or earlier and Clang 10 and earlier by adding a const_cast to `unsigned char*`


jan-wassenberg (approved):
👍

PR #1645, opened 2023-08-09
Updated TestSatWidenMulPairwiseAdd to work around PPC9 GCC 12 codegen issue

johnplatts:
I had updated TestSatWidenMulPairwiseAdd to work around the TestAllSatWidenMulPairwiseAdd failures that were happening on the HWY_PPC9 target when compiled with GCC 12 or GCC 13 (but not GCC 11 or Clang 16), and TestAllSatWidenMulPairwiseAdd will now pass on the PPC9 target when compiled with GCC 12 with the fixes I made to TestSatWidenMulPairwiseAdd in this pull request.


jan-wassenberg (approved):
Great that you have found a workaround :D

Issue #1663, opened 2023-08-17
i686-w64-mingw32 build failed since commit 9b80d37


nanake:
Last good commit 4d7947a69ce09935bf1792783bd38c892fe0cf36
First bad commit 9b80d372f5bd1c51aa806ede430a48b060180657

```console
$ docker run --rm -it ghcr.io/nanake/win32-nonfree:latest
# git -C /opt clone https://github.com/google/highway
# cd /opt/highway
# mkdir build && cd build
# cmake -DCMAKE_TOOLCHAIN_FILE="$FFBUILD_CMAKE_TOOLCHAIN" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="$FFBUILD_PREFIX" -DHWY_ENABLE_TESTS=0 -DHWY_ENABLE_CONTRIB=0 -DHWY_ENABLE_EXAMPLES=0 -GNinja ..
CMake Deprecation Warning at CMakeLists.txt:25 (cmake_policy):
  The OLD behavior for policy CMP0111 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.


-- The C compiler identification is GNU 13.2.1
-- The CXX compiler identification is GNU 13.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/i686-w64-mingw32-gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/i686-w64-mingw32-g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Performing Test ATOMICS_LOCK_FREE_INSTRUCTIONS
-- Performing Test ATOMICS_LOCK_FREE_INSTRUCTIONS - Success
-- Performing Test HWY_EMSCRIPTEN
-- Performing Test HWY_EMSCRIPTEN - Failed
-- Performing Test HWY_RISCV
-- Performing Test HWY_RISCV - Failed
-- Looking for sys/auxv.h
-- Looking for sys/auxv.h - not found
-- Looking for asm/hwcap.h
-- Looking for asm/hwcap.h - not found
[...]

PR #1664, opened 2023-08-17
Fixed compilation issue on i686 target

johnplatts:
Updated hwy/base.h to fix compilation issue on i686 target with SSE2 disabled.

Resolves issue #1663.


jan-wassenberg (approved):
Thank you for fixing this :) The _Float16 has unfortunately caused numerous issues.

PR #1669, opened 2023-08-19
Made enhancements to Compress and Expand operations on PPC9/PPC10

johnplatts:
Made the following enhancements to the Compress and Expand operations on PPC9/PPC10:
- Added detail::CompressOrExpandIndicesFromMask on the HWY_PPC10 target, which is a wrapper around the PPC10 xxgenpcvbm/xxgenpcvhm/xxgenpcvwm instructions
- Reimplemented I8/U8 Compress on PPC10 to do a TableLookupBytes using indices returned by detail::CompressOrExpandIndicesFromMask
- Updated CompressBlendedStore implementations in hwy/ops/ppc_vsx-inl.h to use StoreN on PPC9/PPC10 instead of BlendedStore as StoreN is more efficient than BlendedStore on PPC9/PPC10
- Reimplemented Expand on HWY_PPC10 target to use TableLookupBytes using indices returned by detail::CompressOrExpandIndicesFromMask for vectors that have 1-byte, 2-byte, or 4-byte lanes

### DIFF
--- a/hwy/ops/ppc_vsx-inl.h
+++ b/hwy/ops/ppc_vsx-inl.h
@@ -3981,6 +3981,48 @@ HWY_API intptr_t FindLastTrue(D d, MFromD<D> mask) {
 
 namespace detail {
 
+#if HWY_PPC_HAVE_10
+template <bool kIsCompress, class D, HWY_IF_T_SIZE_D(D, 1)>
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Nice, this looks like a handy POWER10 instruction 👍

PR #1670, opened 2023-08-20
Added platform-specific PromoteUpperTo implementation for PPC

johnplatts:
Added a new platform-specific PromoteUpperTo implementation for PPC as a platform-specific PromoteUpperTo implementation is more efficient than the generic PromoteUpperTo implementation for most of the half-vector to full-vector PromoteUpperTo cases on PPC.

PPC has instructions for I8->I16, I16->I32, and I32->I64 half-vector to full-vector PromoteUpperTo.

PPC9/PPC10 also have instructions for half-vector F16 to full-vector F32 PromoteUpperTo.


jan-wassenberg (approved):
Nice :)

PR #1672, opened 2023-08-22
Rewrap LICENSE-BSD3 file

kleisauke:
To improve readability.

### DIFF
--- a/LICENSE-BSD3
+++ b/LICENSE-BSD3
@@ -1,21 +1,15 @@
 Copyright (c) the Highway Project Authors.
 All rights reserved.
 
-Redistribution and use in source and binary forms,
-with or without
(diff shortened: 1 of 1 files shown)


jan-wassenberg (commented):
Thanks, unfortunately there are some technical reasons for this :)
We do not have the new/second license file in our internal repo because the internal tooling has some difficulty with that. I've had to patch it in via the sync tool, but in pieces so that the patches are not themselves detected as being a license file. And merging the pieces seems to insert newlines, so here we are :/
Is it OK if we keep as-is?

kleisauke:
Thanks for the clarification! It's probably fine to keep it this way, I think most automatic license checking tools will ignore newlines.

PR #1673, opened 2023-08-22
Docs: update amount of supported architectures

kleisauke:
Update the amount of supported architectures now that PowerPC is supported. 

Also, avoid stating how many instruction sets are supported to prevent the need of changes in this text.


jan-wassenberg (approved):
Thanks for updating this :)

PR #1688, opened 2023-08-29
Made fixes to RVV LoadN/LoadNOr

johnplatts:
Updated RVV LoadN/LoadNOr to use a tail-undisturbed load instead of a mask-undisturbed load.

A tail-undisturbed load (_tu) modifies the first `CappedLanes(num_lanes)` elements of the result vector and leaves any lanes past the first (lowest-index) `CappedLanes(num_lanes)` unchanged.

A mask-undisturbed tail-agnostic load (_mu) leaves inactive elements (those whose mask bit is not set) that are within the first (lowest-index) `CappedLanes(num_lanes)` unchanged but may possibly change any lanes past the first (lowest-index) `CappedLanes(num_lanes)` to all 1s.

Using a tail-undisturbed load instead of a mask-undisturbed load in the RVV LoadN and LoadNOr implementations fixes the errors that were encountered with TestAllLoadN and TestAllLoadNOr on the RVV target.

### DIFF
--- a/hwy/ops/rvv-inl.h
+++ b/hwy/ops/rvv-inl.h
@@ -1605,43 +1605,28 @@ HWY_RVV_FOREACH(HWY_RVV_MASKED_LOAD, MaskedLoad, le, _ALL_VIRT)
 #define HWY_NATIVE_LOAD_N
 #endif
 
-namespace detail {
-
-// Passing avl here would be ideal for FirstN, but it does not work because
-// mask ops are tail-agnostic, which (unpredictably) writes a 1 or the result of
-// the mask operation.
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Good catch, thanks. I'm curious under what environment you saw a failure?
Our QEMU didn't trigger a failure.

johnplatts:
> Good catch, thanks. I'm curious under what environment you saw a failure? Our QEMU didn't trigger a failure.

I was running the tests with QEMU 8.0.4 and discovered that TestAllLoadN and TestAllLoadNOr were failing.

jan-wassenberg:
Interesting, thanks; FYI we are still on version 7.2 at the moment.

jan-wassenberg:
Thanks for sharing! Made a note to be able to reproduce such cases.

PR #1704, opened 2023-08-31
Added PromoteEvenTo and PromoteOddTo operations

johnplatts:
The PromoteEvenTo and PromoteOddTo operations were added as SVE/PPC have instructions for some of the PromoteEvenTo and PromoteOddTo operations.

SSE2/SSSE3/SSE4/AVX2 also have platform-specific implementations of the I32->I64 PromoteEvenTo and PromoteOddTo operations to avoid the overhead that is there with I64 ShiftRight on SSE2/SSSE3/SSE4/AVX2.

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -1405,6 +1405,38 @@ The following may be more convenient or efficient than also calling `LowerHalf`
     lanes as `D` and the return value. Only available if
     `HWY_TARGET != HWY_SCALAR`.
 
+The following may be more convenient or efficient than also calling `ConcatEven`
+or `ConcatOdd` followed by `PromoteLowerTo`:
+*   `D`:`{u,i}{16,32,64},f{16,32,64}`, `V`:`Vec<RepartitionToNarrow<D>>`
+    <code>Vec&lt;D&gt; **PromoteEvenTo**(D, V v)</code>: promotes the even lanes
+    of `v` to `TFromD<D>`. Note that `V` has twice as many lanes as `D` and the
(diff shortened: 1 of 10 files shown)


jan-wassenberg (approved):
Nice :)

jan-wassenberg:
Internal CI reports arm7 build failure:

```
hwy/ops/arm_neon-inl.h:6274:10: error: no matching function for call to 'ConcatOdd'
  return detail::ConcatOdd(lo, hi);  // for T=hwy::float16_t
```

The issue appears to be that we only support `Concat*` if f16 is available, but we do allow `Promote*To` even if not, and their implementation may use `Concat*`. How about adding `Concat*` overloads for f16?

johnplatts:
> Internal CI reports arm7 build failure:
> 
> ```
> hwy/ops/arm_neon-inl.h:6274:10: error: no matching function for call to 'ConcatOdd'
>   return detail::ConcatOdd(lo, hi);  // for T=hwy::float16_t
> ```
> 
> The issue appears to be that we only support `Concat*` if f16 is available, but we do allow `Promote*To` even if not, and their implementation may use `Concat*`. How about adding `Concat*` overloads for f16?

I have added the missing float16_t overloads for ConcatEven, ConcatOdd, and InterleaveLower for ArmV7 NEON when `!HWY_HAVE_FLOAT16`, and adding the missing float16_t overloads for ConcatEven, ConcatOdd, and InterleaveLower fixes the compilation errors on ArmV7 NEON.

jan-wassenberg (approved):
Nice, thanks for adding the overloads :)

PR #1705, opened 2023-08-31
Really disables math_test

malaterre:
Fixes commit a02c342bf3b096b6c035165d3641f032e403506b


jan-wassenberg (approved):
Oops, thanks for fixing!

PR #1710, opened 2023-09-01
Add VQSelect

zbjornson:
The monotonous work is done and the new test suite works for `std::nth_element`, but fails for `VQSelect()` - it's skipping an entire partition. I don't see the cause so far though. There are a few other TODOs within the code as well.

Direct edits and comments both welcome.


hwy/contrib/sort/sort_test.cc
```diff
@@ -713,6 +713,89 @@ void TestAllSort() {
...
+          lanes[num_lanes + i] = hwy::HighestValue<LaneType>();
+        }
+#if HWY_IS_MSAN
+        __msan_poison(aligned.get(), misalign * sizeof(LaneType));
+        __msan_poison(lanes + num_lanes, kMaxMisalign * sizeof(LaneType));
+#endif
+        InputStats<LaneType> input_stats =
+            GenerateInput(dist, lanes, num_lanes);
+
+        const size_t k = num_lanes / 2;
```
jan-wassenberg, line 765: How about randomizing this k for better coverage?

hwy/contrib/sort/vqsort-inl.h
```diff
@@ -1660,11 +1660,21 @@ HWY_NOINLINE void Recurse(D d, Traits st, T* HWY_RESTRICT keys,
...
 
-  if (HWY_LIKELY(result != PivotResult::kIsFirst)) {
-    Recurse(d, st, keys, bound, buf, state, remaining_levels - 1);
-  }
-  if (HWY_LIKELY(result != PivotResult::kWasLast)) {
-    Recurse(d, st, keys + bound, num - bound, buf, state, remaining_levels - 1);
+  if (kIsSelect) {
+    if (HWY_LIKELY(result != PivotResult::kIsFirst) && k < bound) {
+      Recurse<true>(d, st, keys, bound, buf, state, remaining_levels - 1, k);
+    } else if (HWY_LIKELY(result != PivotResult::kWasLast) && k >= bound) {
```
jan-wassenberg, line 1666: If k == bound, aren't we done already?

jan-wassenberg (commented):
Some minor comments but I don't see a smoking gun. Have you tried defining VQSORT_PRINT=2 and reading through the output? This should show where we are recursing.

hwy/contrib/sort/vqsort-inl.h
zbjornson: Either no, or there's an off-by-one somewhere.

With this fix, the tests pass 1000 runs with random k:
https://github.com/google/highway/compare/136b58efccc870d90b6b1836ad2b29270fac05d8..8915b8caff1e203f9844d05790d37d6354840204#diff-adcb5fe8d4a8b73787547a9bff2dd40a1d60e93147153747d8ed317dfc37d670R1667-R1668

jan-wassenberg: Oh, good catch with the k - bound: we're shrinking the interval.

Aha, I think I understand the k==bound case: suppose (10, 20, 40, 30, 31), pivot 30, bound=2. So the 10,20 are skipped, we recurse into 40,30,31 but the first item is not yet correctly placed. So yes, >= is indeed necessary.

Looks good. Are you happy with the testing? How about we do a bit of final cleanup (other comment above, plus adding some documentation on what k means and that it's in [0, num]), then we can merge?

jan-wassenberg:
@zbjornson would you like to make the minor change suggested? Otherwise I'm happy to do it after we merge.

jan-wassenberg:
I understand, we'll take care of it after merging. Thanks for implementing this :)

jan-wassenberg (approved):

jan-wassenberg:
Sorry to drop the ball on this :/ The merge had failed because a rebase was required, but I missed the notification.
@zbjornson , any chance you could help update this PR? This would help @enum-class implement TopK in #2019 :)

hwy/contrib/sort/algo-inl.h
```diff
@@ -525,16 +529,26 @@ void Run(Algo algo, KeyType* HWY_RESTRICT inout, size_t num,
...
+    case Algo::kStdSort:
       if (Order().IsAscending()) {
         return std::sort(inout, inout + num, less);
       } else {
         return std::sort(inout, inout + num, greater);
       }
 
+    case Algo::kStdSelect:
+      if (Order().IsAscending()) {
+        return std::nth_element(inout, inout + k, inout + num, less);
```
enum-class, line 541: @zbjornson @jan-wassenberg Highway is meant to work with C++11, but std::nth_element was introduced in C++17. Is this okay?

jan-wassenberg: Good point, thanks for pointing that out. This is test-only, so I think we could simply wrap all usages in `#if HWY_CXX_LANG >= 201703L`.

jan-wassenberg:
I have learned how to patch copybara changes to fix the compile error we were seeing (inserting the missing hwy::). Hopefully import will happen soon :)

jan-wassenberg:
FYI the patch worked but we have an issue with MSVC which I am addressing in #2022 .

PR #1726, opened 2023-09-07
Added BitCastScalar operation

johnplatts:
Added BitCastScalar operation, which bit casts `val` to `To`.

Also added `HWY_BITCASTSCALAR_CONSTEXPR` macro, which expands to the `constexpr` keyword if the current compiler supports `__builtin_bit_cast` and is defined as an empty macro otherwise.


jan-wassenberg (approved):
Nice, thanks for adding this and updating the usages :)

PR #1733, opened 2023-09-09
Fixes to PPC integer arithmetic ops to avoid UB

johnplatts:
Updated PPC I8/I16/I32/I64 Neg implementation to do `Zero(d) - v` instead of `vec_neg(v.raw)` to avoid undefined behavior in the case where `v[i] == LimitsMin<T>()`.

GCC/Clang will still optimize `Zero(d) - v` for I32/U32 vectors down to a single `vnegw` instruction on PPC9/PPC10 if optimizations are enabled.

GCC/Clang will still optimize `Zero(d) - v` for I64/U64 vectors down to a single `vnegd` instruction on PPC9/PPC10 if optimizations are enabled.

Updated PPC I8/I16/I32/I64 Abs implementation to do `Max(v, Neg(v))` instead of `vec_abs(v.raw)` to avoid undefined behavior in the case where `v[i] == LimitsMin<T>()`.

`Max(v, Neg(v))` will compile down to the same instructions as `vec_abs(v.raw)` for signed integer vectors on PPC if optimizations are enabled.

Updated the implementations of PPC Add/Sub/Mul/ShiftLeftSame/RotateRight operations to BitCast signed integer vectors to unsigned integer vectors to avoid undefined behavior.

The changes made to the Add/Sub/Mul/ShiftLeftSame/RotateRight/Neg/Abs operations in this pull request fixes UBSan errors that were previously reported on PPC9/PPC10 when UBSan is enabled.

The changes made to the Add/Sub/Mul/ShiftLeftSame/RotateRight/Neg/Abs operations on PPC in this pull request also fix codegen issues with GCC 12/GCC 13.

### DIFF
--- a/hwy/ops/ppc_vsx-inl.h
+++ b/hwy/ops/ppc_vsx-inl.h
@@ -476,8 +476,16 @@ HWY_API Vec128<T, N> operator^(Vec128<T, N> a, Vec128<T, N> b) {
 
 // ------------------------------ Neg
 
-template <typename T, size_t N, HWY_IF_NOT_SPECIAL_FLOAT(T)>
-HWY_INLINE Vec128<T, N> Neg(Vec128<T, N> v) {
+template <typename T, size_t N, HWY_IF_SIGNED(T)>
+HWY_API Vec128<T, N> Neg(Vec128<T, N> v) {
+  // If T is an signed integer type, use Zero(d) - v instead of vec_neg to
+  // avoid undefined behavior in the case where v[i] == LimitsMin<T>()
+  const DFromV<decltype(v)> d;
+  return Zero(d) - v;
+}
+
+template <typename T, size_t N, HWY_IF_FLOAT3264(T)>
+HWY_API Vec128<T, N> Neg(Vec128<T, N> v) {
   return Vec128<T, N>{vec_neg(v.raw)};
 }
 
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Nice to fix sanitizer issues. I'm surprised the compiler is so aggressive about allowing this to break codegen in vector code; in my experience, code is not very strict about add/shift UB.
Does this fix any of the HWY_BROKEN_PPC10 or similar?

johnplatts:
> Nice to fix sanitizer issues. I'm surprised the compiler is so aggressive about allowing this to break codegen in vector code; in my experience, code is not very strict about add/shift UB. Does this fix any of the HWY_BROKEN_PPC10 or similar?

The GCC/Clang implementations of the SSE2/SSE4.1/AVX2/AVX512 _mm_add/_mm_sub/_mm_mullo intrinsics actually bitcast a __m128i/__m256i/__m512i vector to an unsigned integer vector type.

The Clang implementations of the wasm_i8x16_add/sub, wasm_i16x8_add/sub/mul, wasm_i32x4_add/sub/mul, and wasm_i64x2_add/sub/mul intrinsics actually bitcast a v128_t vector to an unsigned integer vector type.

jan-wassenberg:
Ah, that explains the difference, thanks :)

Issue #1727, opened 2023-09-08
Add hwy::IsInteger and hwy::IsFixedSizeInteger type traits


johnplatts:
`hwy::IsInteger<T>()` returns true if and only if T is any of the following types:
`bool`, `char`, `signed char`, `unsigned char`, `short`, `unsigned short`, `int`, `unsigned int`, `long`, `unsigned long`, `long long`, `unsigned long long`, `int8_t`, `uint8_t`, `int16_t`, `uint16_t`, `int32_t`, `uint32_t`, `int64_t`, `uint64_t`, `char8_t`, `char16_t`, `char32_t`, `wchar_t`, `intptr_t`, `uintptr_t`, `size_t`, `ptrdiff_t`

`hwy::IsFixedSizeInteger<T>()` returns true if and only if T is any of the following types:
`int8_t`, `uint8_t`, `int16_t`, `uint16_t`, `int32_t`, `uint32_t`, `int64_t`, `uint64_t`

`hwy::IsFixedSizeInteger<T>()` only includes the integer types that are valid for a Highway vector lane whereas `hwy::Integer<T>()` includes some additional integer types.

There are some use cases for a separate `hwy::Integer<T>()` that returns true for some additional integer types, including enabling operations such as `PopCount`, `Num0BitsBelowLS1Bit`, and `Num0BitsAboveMS1Bit` for additional integer types other than those for which `hwy::IsFixedSizeInteger<T>()` would return true.

jan-wassenberg:
These sound like useful traits to add 👍 Would you like to add them?
I wonder about the name: char8_t and even char (by definition size 1) are also fixed-size. Perhaps the real distinction is whether we support/allow them as lane types. Then maybe IsIntegerLaneType or IsSupportedIntegerType?

johnplatts:
> These sound like useful traits to add 👍 Would you like to add them? I wonder about the name: char8_t and even char (by definition size 1) are also fixed-size. Perhaps the real distinction is whether we support/allow them as lane types. Then maybe IsIntegerLaneType or IsSupportedIntegerType?

IsIntegerLaneType is a better name for integer types that are supported as Highway lane types.

PR #1734, opened 2023-09-09
Added IsInteger, IsIntegerLaneType, RemoveVolatile, and RemoveCvRef

johnplatts:
Added the following type traits:
- `IsInteger<T>()` - returns true if `T` is one of the following types and returns false otherwise:
  `bool`, `char`, `signed char`, `unsigned char`, `short`, `unsigned short`, `int`, `unsigned int`, `long`, `unsigned long`, `long long`, `unsigned long long`, `int8_t`, `uint8_t`, `int16_t`, `uint16_t`, `int32_t`, `uint32_t`, `int64_t`, `uint64_t`, `char8_t`, `char16_t`, `char32_t`, `wchar_t`, `intptr_t`, `uintptr_t`, `size_t`, `ptrdiff_t`
- `IsIntegerLaneType<T>()` - returns true if `T` is one of the following types and returns false otherwise:
  `int8_t`, `uint8_t`, `int16_t`, `uint16_t`, `int32_t`, `uint32_t`, `int64_t`, `uint64_t`
- `RemoveVolatile<T>` - removes the `volatile` qualifier off of `T`
- `RemoveCvRef<T>` - equivalent to `RemoveConst<RemoveVolatile<RemoveRef<T>>>`

Resolves issue #1727.

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -894,6 +894,7 @@ using If = typename IfT<Condition, Then, Else>::type;
   hwy::EnableIf<hwy::IsFloat<T>() || hwy::IsSpecialFloat<T>()>* = nullptr
 #define HWY_IF_NOT_FLOAT_NOR_SPECIAL(T) \
   hwy::EnableIf<!hwy::IsFloat<T>() && !hwy::IsSpecialFloat<T>()>* = nullptr
+#define HWY_IF_INTEGER(T) hwy::EnableIf<hwy::IsInteger<T>()>* = nullptr
 
 #define HWY_IF_T_SIZE(T, bytes) hwy::EnableIf<sizeof(T) == (bytes)>* = nullptr
 #define HWY_IF_NOT_T_SIZE(T, bytes) \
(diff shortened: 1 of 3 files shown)


hwy/base.h
```diff
@@ -1174,14 +1299,14 @@ constexpr bool IsSigned<bfloat16_t>() {
 // Largest/smallest representable integer values.
 template <typename T>
 HWY_API constexpr T LimitsMax() {
-  static_assert(!IsFloat<T>(), "Only for integer types");
-  using TU = MakeUnsigned<T>;
-  return static_cast<T>(IsSigned<T>() ? (static_cast<TU>(~0ull) >> 1)
-                                      : static_cast<TU>(~0ull));
+  static_assert(IsInteger<T>(), "Only for integer types");
+  using TU = UnsignedFromSize<sizeof(T)>;
```
jan-wassenberg, line 1303: Should we change MakeUnsigned to this implementation?

jan-wassenberg (approved):
Very nice, thanks for adding this and the extra checks :)

hwy/base.h
johnplatts: The reason why `UnsignedFromSize<sizeof(T)>` is used instead of `MakeSigned<T>` is that it is possible for `!IsSame<size_t, uint32_t>() && !IsSame<size_t, uint64_t>()` to be true on some platforms and the current implementation of `hwy::MakeSigned<T>` will give compiler errors if `!IsIntegerLaneType<T>() && !IsFloat<T>() && !IsSpecialFloat<T>()` is true.

jan-wassenberg: I understand why you are using UnsignedFromSize, this makes sense. Let me rephrase: what if we changed the implementation of Make(Un)Signed to what you are doing here, then we could still use Make(Un)Signed here, because it is a bit shorter/more clear?

PR #1743, opened 2023-09-13
Add an option to opt-out of HWY_RISCV

### LINKED ISSUE #1740: riscv64/LTO: lto1: fatal error: target specific builtin not available
Seems like default flag: ` -march=rv64gcv1p0` is conflicting with `-flto=auto`. Will need to dig further.

ref:
* https://buildd.debian.org/status/fetch.php?pkg=highway&arch=riscv64&ver=1.0.7-3&stamp=1694507044&raw=0

malaterre:
Fixes #1740


jan-wassenberg (approved):
👍

Issue #1739, opened 2023-09-12
ppc64el/LTO: Illegal instruction


malaterre:
I cannot use LTO flag on GCC-13 (default POWER8 setup).

Current state of investigation is:

```
 % gdb obj-*/./tests/math_test
GNU gdb (Debian 13.2-1) 13.2
Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "powerpc64le-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from obj-powerpc64le-linux-gnu/./tests/math_test...
(gdb) r
Starting program: /home/malat/new/highway-1.0.7/obj-powerpc64le-linux-gnu/tests/math_test
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/powerpc64le-linux-gnu/libthread_db.so.1".

Program received signal SIGILL, Illegal instruction.
hwy::N_PPC10::BitCast<float, int>(int const&) [clone .constprop.0] (in=@0x7ffffffff094: 1065353215) at ./hwy/contrib/math/math_test.cc:77
77      }
(gdb) display/i $pc
1: x/i $pc
=> 0x10000fcf0 <_ZN3hwy7N_PPC107BitCastIfiEET_RKT0_.constprop.0>:       xxspltidp vs1,1065353215
```

jan-wassenberg:
hm, xxspltidp is a PPC10 instruction that splats/broadcasts. This is reasonable codegen for initializing from a constant. I think the issue might instead be that QEMU is older and does not support PPC10? Version 7.2 is required.

jan-wassenberg:
Oh, I think I understand what might be happening. It appears the compiler has chosen the PPC10 static initializer and is running that at startup, before we even check the CPU. That would explain why your patch works: it moves init into the function and avoids static initializers. Let's go with your approach then :) I'll also document the problem later.

PR #1744, opened 2023-09-13
Workaround issue on ppc64el with LTO

malaterre:
Related to #1739


jan-wassenberg (commented):
hm, this seems workable. But maybe a simpler fix would be to make these variables static constexpr?

BTW while we're at it, we could replace occurrences of BitCast with BitCastScalar, and then remove the BitCast function definition from this file. Please let me know if you'd like to do that, otherwise I can.

jan-wassenberg:
I am not very familiar with PPC codegen. I suppose it is plausible that the compiler might realize in the LTO case it can just splat from one scalar, whereas in the non-LTO case it might either be pre-broadcasting the scalar to a vector in memory, or generating older-style code that does not use the new splat instruction.

PR #1757, opened 2023-09-15
Fix compilation on armhf/clang

malaterre:


jan-wassenberg (approved):
huh, I am surprised LLVM does not support this. Thanks for fixing it :)

jan-wassenberg:
Right, I meant at least supporting the flag. It would be nice if clang and GCC were more compatible.

PR #1765, opened 2023-09-19
Narrow include in algo-inl.h

benjaminmordaunt:
At the moment, algo-inl.h pulls in the test util header which in turn depends on the Google Test library. It seems however that only aligned allocator support is required. Narrow the include to permit using this header in contexts which do not form part of testing, or where a dependency on Google Test is not desired.


jan-wassenberg (approved):
Thanks, sounds very reasonable :) I have verified this is the only non-test code that includes test_util.

Issue #1756, opened 2023-09-15
Adding full-vector interleave lower/upper operations for >= 32-byte vectors


johnplatts:
While Highway has blockwise InterleaveLower and InterleaveUpper operations that operate on 128-bit or smaller blocks, there are use cases for full-vector interleave lower/upper operations for vectors with more than 32 bytes (including unit tests).

A full-vector interleave lower operation is equivalent to the following (even for vectors that are larger than 16 bytes):
`a[0]`, `b[0]`, `a[1]`, `b[1]`, ... , `a[N/2 - 2]`, `b[N/2 - 2]`, `a[N/2 - 1]`, `b[N/2 - 1]`

A full-vector interleave upper operation is equivalent to the following (even for vectors that are larger than 16 bytes):
`a[N/2]`, `b[N/2]`, `a[N/2 + 1]`, `b[N/2 + 1]`, ..., `a[N - 2]`, `b[N - 2]`, `a[N - 1]`, `b[N - 1]`

jan-wassenberg:
Thanks @johnplatts for proposing this. I agree it can be useful, and even efficient on AVX-512 with `permutex2var`. Also on RVV the cross-block op would be faster because it can for <64 bit types zero-extend, slide, and OR.

Let's think about naming. Some of the within-block operations have Block in their name, but unfortunately not Zip/InterleaveLower. Maybe InterleaveAllLower? 'Full' is probably best kept for distinguishing between full and partial, not block vs whole. Or InterleaveWholeLower? (It seems useful to have the Lower/Upper at the end.) What do you think?

johnplatts:
> Thanks @johnplatts for proposing this. I agree it can be useful, and even efficient on AVX-512 with `permutex2var`. Also on RVV the cross-block op would be faster because it can for <64 bit types zero-extend, slide, and OR.
> 
> Let's think about naming. Some of the within-block operations have Block in their name, but unfortunately not Zip/InterleaveLower. Maybe InterleaveAllLower? 'Full' is probably best kept for distinguishing between full and partial, not block vs whole. Or InterleaveWholeLower? (It seems useful to have the Lower/Upper at the end.) What do you think?

InterleaveWholeLower and InterleaveWholeUpper sound good for the interleave operations that operate on the lower half or upper half of the source vectors.

jan-wassenberg:
Sounds good to me as well :) Would you like to add them?

PR #1766, opened 2023-09-20
Added InterleaveWholeLower and InterleaveWholeUpper ops

johnplatts:
Added the InterleaveWholeLower and InterleaveWholeUpper operations, which operate on the whole lower halves or whole upper halves of the vectors.

Resolves issue #1756.


jan-wassenberg (approved):
Very nice, thanks for adding this!

PR #1768, opened 2023-09-21
Fixes to SVE F32->U64 PromoteEvenTo and TestPromoteOddEvenTo

johnplatts:
Fixed SVE F32->U64 PromoteEvenTo implementation to call svcvt_u64_f32_x instead of svcvt_s64_f32_x.

Also updated TestPromoteOddEvenTo to test F32->I64 and F32->U64 PromoteEvenTo/PromoteOddTo.


jan-wassenberg (approved):
👍

PR #1769, opened 2023-09-21
Updated AddWithWraparound to fix warnings in reduction_test

johnplatts:
Changed the type of the third parameter of hwy::AddWithWraparound from `size_t` to `T2` to fix compilation warnings in TestSumOfLanes in hwy/tests/reduction_test.cc.


jan-wassenberg (approved):
Thank you for fixing this warning :)

PR #1776, opened 2023-09-24
Added IfNegativeThenNegOrUndefIfZero operation

johnplatts:
The CondNegateOrZero operation was added as SSSE3/SSE4/AVX2/AVX3 have PSIGNB/PSIGNW/PSIGND instructions that can carry out the `IfNegativeThenElse(b, Neg(a), IfThenZeroElse(Eq(b, Zero(d)), a))` operation for I8/I16/I32 vectors that are 32 bytes or smaller.

AVX3 can also more efficiently carry out the `b[i] < 0 ? (-a[i]) : (a[i])` operation for I8/I16/I32/I64 vectors using `MaskedSubOr(a, MaskFromVec(b), Zero(d), a)`.

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -850,6 +850,12 @@ Special functions for signed types:
     <code>V **IfNegativeThenElse**(V v, V yes, V no)</code>: returns `v[i] < 0 ?
     yes[i] : no[i]`. This may be more efficient than `IfThenElse(Lt..)`.
 
+*   `V`: `{i,f}` \
+    <code>V **CondNegateOrZero**(V a, V b)</code>: returns `b[i] < 0 ? (-a[i]) :
+    ((b[i] == 0) ? 0 : a[i])`. `CondNegateOrZero(a, b)` is more efficient than
+    `IfNegativeThenElse(b, Neg(a), IfThenZeroElse(Eq(b, Zero(d)), a))` for
+    I8/I16/I32 vectors on SSSE3/SSE4/AVX2/AVX3.
+
 ### Masks
 
 Let `M` denote a mask capable of storing a logical true/false for each lane (the
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -851,10 +851,14 @@ Special functions for signed types:
     yes[i] : no[i]`. This may be more efficient than `IfThenElse(Lt..)`.
 
 *   `V`: `{i,f}` \
-    <code>V **CondNegateOrZero**(V a, V b)</code>: returns `b[i] < 0 ? (-a[i]) :
-    ((b[i] == 0) ? 0 : a[i])`. `CondNegateOrZero(a, b)` is more efficient than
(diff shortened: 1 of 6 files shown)


jan-wassenberg (changes requested):
I like your idea of supporting `sign_epi*`. Am concerned, though, that by exactly matching its semantics, we force other platforms to go to more effort than might be required for the user's code, and introduce performance cliffs. I have not yet seen a use-case for conditionally zeroing if b=0. Also, sign(x, x) is already covered by Abs(x).

What does seem very useful is `if b<0, negate`. This can help shoehorn two signed values into the unsigned+signed pair expected by x86 maddubs and also SumOfMulQuadAccumulate for i8->i32. By skipping the OrZero part, it's also cheaper to emulate on other platforms. We could then name it `IfNegativeNeg()` or similar to match existing functions. What do you think?

johnplatts:
> What does seem very useful is `if b<0, negate`. This can help shoehorn two signed values into the unsigned+signed pair expected by x86 maddubs and also SumOfMulQuadAccumulate for i8->i32. By skipping the OrZero part, it's also cheaper to emulate on other platforms. We could then name it `IfNegativeNeg()` or similar to match existing functions. What do you think?

I like the idea of a NegateIfNegative(a, b) operation that has implementation-defined results if `b[i] == 0` is true and returns `(b[i] < 0) ? (-a[i]) : a[i])` if `b[i] != 0` is true.

Some of the users of IfNegativeNeg would expect IfNegativeNeg(a, b) to have the same semantics as `IfNegativeThenElse(b, Neg(b), a)` on all targets with IfNegativeNeg returning `a[i]` in the `b[i] == 0` case.

`IfNegativeThenElse(b, Neg(b), a)` is equivalent to `MaskedSubOr(a, b, Zero(d), a)` for integer vectors on all targets, but `MaskedSubOr(a, b, Zero(d), a)` is more efficient than `IfNegativeThenElse(b, Neg(b), a)` on some targets such as AVX3.

There are use cases for both a NegateIfNegative(a, b) operation that has implementation-defined results for the `b[i] == 0` case and a IfNegativeNeg(a, b) operation that returns `a[i]` if `b[i] == 0` case on all targets.

jan-wassenberg:
Ah, good point about the implementation-defined. For the `i*i` product using `u*i` instructions, we'd first take abs(a), then 'multiply' the sign of a into b via `flipped_sign_b = NegIfNegative(b, a)`, then abs(a)*flipped_sign_b. If a=0, then the product is anyway going to be zero and implementation-defined is OK. The use-case is time-critical enough that I doubt we'd want to go with the more general abs_epi8+blendv_epi8 instead.

Still, I wonder if we can make the behavior explicit in the name so the difference vs. IfNegativeNeg (which I agree is also useful) is more clear - perhaps IfNegativeThenNegUndefinedIfZero?

Let's also consider the argument order as a potential stumbling block. It's probably fine (or at least consistent) with the longer IfNegative.. names, but a concern if we call it NegIfNegative (which is which?).

jan-wassenberg (approved):
Thanks for renaming the op :)

PR #1781, opened 2023-09-26
Added optimized implementations of I8/U8 SumsOf4 for AVX3

johnplatts:
Updated U8 SumsOf4 to use _mm_maskz_dbsad_epu8/_mm256_maskz_dbsad_epu8/_mm512_maskz_dbsad_epu8 on AVX3 targets.

Also added an optimized implementation of I8 SumsOf4 on AVX3 that does the following:
- Adjusts the source values of the source I8 vector to be in the 0..255 range, bitcasted to an uint8_t vector
- Sums the adjusted source values by doing an U8 SumsOf4 operation
- Adds -512 to each i32 lane of the result of the U8 SumsOf4 operation (bitcasted to an I32 vector) to account for the adjustment that was made to the source values


jan-wassenberg (approved):
Very nice, thanks for adding this optimization :D

PR #1783, opened 2023-09-26
Docs: add libvips to list of users

kleisauke:
See: https://github.com/libvips/libvips/pull/3618


jan-wassenberg (approved):
Nice, thanks for adding to the list :)

PR #1786, opened 2023-09-27
Fix compilation on x86+AVX2

malaterre:
Fixes compilation error:

hwy/ops/x86_512-inl.h: In function 'hwy::N_AVX3::MFromD<D> hwy::N_AVX3::FirstN(D, size_t)': hwy/ops/x86_512-inl.h:755:25: error: 'T' was not declared in this scope
  755 |   return detail::FirstN<T>(n);
      |                         ^


jan-wassenberg (approved):
Thank you for fixing this!

PR #1788, opened 2023-09-28
Made fixes to I8 SumsOf8AbsDiff

johnplatts:
Updated I8/U8 implementation of SumsOf8AbsDiff in generic_ops-inl.h to fix a bug with I8 SumsOf8AbsDiff as the result of the I8 AbsDiff operation needs to be bitcasted to an U8 vector.

Also enabled TestSumsOf8AbsDiff for int8_t vectors that are >= 64 bits.


jan-wassenberg (approved):
👍

PR #1795, opened 2023-10-05
Fix UB case with signed overflow, prefer unsigned

malaterre:
Link: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=110643
Suggested-by: Andrew Pinski <pinskia@gcc.gnu.org>


jan-wassenberg (approved):
Thank you for fixing, and to Andrew Pinski for suggesting the fix :)
I will also apply this to the other implementations after we merge.

PR #1798, opened 2023-10-05
Detect signed arithemtic overflow in GCC github/CI

malaterre:
See:
* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=110643


jan-wassenberg (approved):
Nice, thanks for adding this.

Issue #1706, opened 2023-08-31
Adding VecFromValues and Dup128VecFromValues operations


johnplatts:
I would like to add the VecFromValues and Dup128VecFromValues operations to Google Highway as VecFromValues and Dup128VecFromValues have better codegen than using a temporary array followed by Load or LoadDup128 on some targets.

VecFromValues is equivalent to _mm_set_epi8, _mm_set_epi16, _mm_set_epi32, _mm_set_epi64, _mm_set_ps, and _mm_set_pd on SSE2/SSSE3/SSE4/AVX2/AVX3.

Dup128VecFromValues is equivalent to VecFromValues followed by BroadcastBlock or svdupq_n_s8, svdupq_n_s16, svdupq_n_s32, svdupq_n_s64, svdupq_n_u8, svdupq_n_u16, svdupq_n_u32, svdupq_n_u64, svdupq_n_f16, svdupq_n_f32, svdupq_n_f64 on SVE.

VecFromValues/Dup128VecFromValues can be implemented for PPC/WASM/NEON using GCC/Clang vector extensions when compiled with GCC or Clang.

Here are generic implementations for VecFromValues and Dup128VecFromValues:
```
template <class D, class... TVal>
HWY_API VFromD<D> VecFromValues(D d, TVal&&... vals) {
#if HWY_HAVE_SCALABLE
  constexpr size_t kArraySize = HWY_MAX(sizeof...(TVal), 1);
#else
  constexpr size_t kArraySize = HWY_MAX(sizeof...(TVal), HWY_MAX_LANES_D(D));
#endif

  HWY_ALIGN const TFromD<D> src_values[kArraySize] = {
      static_cast<TFromD<D>>(static_cast<TVal&&>(vals))...};
#if HWY_HAVE_SCALABLE
  return LoadN(d, src_values, kArraySize);
#else
  return Load(d, src_values);
#endif
}

template <class D, class... TVal>
HWY_API VFromD<D> Dup128VecFromValues(D d, TVal&&... vals) {
  constexpr size_t kMaxLanesToLoad =
      HWY_MIN(HWY_MAX_LANES_D(D), 16 / sizeof(TFromD<D>));
  constexpr size_t kArraySize = HWY_MAX(sizeof...(TVal), kMaxLanesToLoad);

  alignas(16) const TFromD<D> src_values[kMaxLanesToLoad] = {
      static_cast<TFromD<D>>(static_cast<TVal&&>(vals))...};
[...]

PR #1801, opened 2023-10-06
Added Dup128VecFromValues operation

johnplatts:
Resolves issue #1706

Dup128VecFromValues was added as Dup128FromValues is more efficient than LoadDup128 on some targets.

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -412,6 +412,26 @@ example `Lt` instead of `operator<`.
     lane has index 0. This is useful in tests for detecting lane-crossing bugs.
 *   <code>V **SignBit**(D, T)</code>: returns N-lane vector with all lanes set
     to a value whose representation has only the most-significant bit set.
+*   `D`: `{u,i}8` \
+    <code>V **Dup128VecFromValues**(D d, T t0, T t1, T t2, T t3, T t4, T t5,
+    T t6, T t7, T t8, T t9, T t10, T t11, T t12, T t13, T t14, T t15)</code>:
+    Creates a vector with the following values, broadcasted to each 16-byte
+    block if `Lanes(d) >= 16` is true:
+    `{t0, t1, t2, t3, t4, t5, t6, t7, t8, t9, t10, t11, t12, t13, t14, t15}`
+*   `D`: `{u,i,f,bf}16` \
+    <code>V **Dup128VecFromValues**(D d, T t0, T t1, T t2, T t3, T t4, T t5,
+    T t6, T t7)</code>:
+    Creates a vector with the following values, broadcasted to each 16-byte
+    block if `Lanes(d) >= 8` is true:
+    `{t0, t1, t2, t3, t4, t5, t6, t7}`
+*   `D`: `{u,i,f}32` \
+    <code>V **Dup128VecFromValues**(D d, T t0, T t1, T t2, T t3)</code>:
+    Creates a vector with the following values, broadcasted to each 16-byte
+    block if `Lanes(d) >= 4` is true: `{t0, t1, t2, t3}`
+*   `D`: `{u,i,f}64` \
+    <code>V **Dup128VecFromValues**(D d, T t0, T t1)</code>:
+    Creates a vector with the following values, broadcasted to each 16-byte
+    block if `Lanes(d) >= 2` is true: `{t0, t1}`
 
 ### Getting/setting lanes
 
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -412,26 +412,15 @@ example `Lt` instead of `operator<`.
     lane has index 0. This is useful in tests for detecting lane-crossing bugs.
 *   <code>V **SignBit**(D, T)</code>: returns N-lane vector with all lanes set
     to a value whose representation has only the most-significant bit set.
-*   `D`: `{u,i}8` \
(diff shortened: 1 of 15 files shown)


g3doc/quick_reference.md
```diff
@@ -412,6 +412,26 @@ example `Lt` instead of `operator<`.
     lane has index 0. This is useful in tests for detecting lane-crossing bugs.
 *   <code>V **SignBit**(D, T)</code>: returns N-lane vector with all lanes set
     to a value whose representation has only the most-significant bit set.
+*   `D`: `{u,i}8` \
+    <code>V **Dup128VecFromValues**(D d, T t0, T t1, T t2, T t3, T t4, T t5,
+    T t6, T t7, T t8, T t9, T t10, T t11, T t12, T t13, T t14, T t15)</code>:
+    Creates a vector with the following values, broadcasted to each 16-byte
```
jan-wassenberg, line 418: Would you mind updating the documentation so that there is a single op, for all types?
We could express the varying argument count as T t0, .., T tK where K is 16/sizeof(T) - 1.

Would also be good to explain how replication works: t0 goes into the least-significant lane of each 128-bit block.

hwy/ops/x86_128-inl.h
```diff
@@ -394,6 +394,102 @@ HWY_API VFromD<D> ResizeBitCast(D d, FromV v) {
...
+
+template <class D, HWY_IF_UI8_D(D), HWY_IF_V_SIZE_LE_D(D, 16)>
+HWY_API VFromD<D> Dup128VecFromValues(D /*d*/, TFromD<D> t0, TFromD<D> t1,
+                                      TFromD<D> t2, TFromD<D> t3, TFromD<D> t4,
+                                      TFromD<D> t5, TFromD<D> t6, TFromD<D> t7,
+                                      TFromD<D> t8, TFromD<D> t9, TFromD<D> t10,
+                                      TFromD<D> t11, TFromD<D> t12,
+                                      TFromD<D> t13, TFromD<D> t14,
+                                      TFromD<D> t15) {
+  return VFromD<D>{_mm_set_epi8(
```
jan-wassenberg, line 407: I'm curious if there is any reason not to use setr_epi8? No need to change it to that now.

jan-wassenberg (changes requested):
Very nice implementation, thanks for adding this!
Let's update the docs a bit. Would you also like to add Dup128VecFromValues to the list of ops that unconditionally support f16/bf16? It's after "Note that minimal support" in quick_reference.

hwy/ops/x86_128-inl.h
johnplatts: Updated x86 implementation to use _mm_setr intrinsics where available, but there are a few cases where _mm_set intrinsics needs to be used as there are no _mm_setr_epi64x, _mm512_setr_epi8, or _mm512_setr_epi16 intrinsics available on x86.

g3doc/quick_reference.md
johnplatts: Updated Dup128VecFromValues documentation

hwy/ops/x86_128-inl.h
jan-wassenberg: Nice. Huh, I'm surprised that those are missing. But Highway shields users from these inconsistencies, so all good.

jan-wassenberg (approved):
Thanks for making these changes :)

jan-wassenberg:
CI is failing with:
```
emu128-inl.h:169: error: aggregate initialization of type 'const VFromD<Simd<signed char, 16, 0>>' (aka 'const Vec128<signed char, Simd<signed char, 16, 0>::kPrivateLanes>') with user-declared constructors is incompatible with C++20
  169 |   const VFromD<D> result = {
      |                            ^
  170 |       {t0, t1, t2, t3, t4, t5, t6, t7, t8, t9, t10, t11, t12, t13, t14, t15}};
```
Is there a better way to do this than actually writing result.raw[0] = t0; etc?

jan-wassenberg (approved):

johnplatts:
> Is there a better way to do this than actually writing result.raw[0] = t0; etc?

`result.raw[0] = t0;`, etc. probably is the best way to implement this.

hwy/ops/arm_sve-inl.h
```diff
@@ -521,6 +521,102 @@ HWY_API VFromD<D> ResizeBitCast(D d, FromV v) {
...
+
+template <class D, HWY_IF_F32_D(D)>
+HWY_API svfloat32_t Dup128VecFromValues(D /*d*/, TFromD<D> t0, TFromD<D> t1,
+                                        TFromD<D> t2, TFromD<D> t3) {
+  return svdupq_n_f32(t0, t1, t2, t3);
+}
+
+template <class D, HWY_IF_I64_D(D)>
+HWY_API svint64_t Dup128VecFromValues(D /*d*/, TFromD<D> t0, TFromD<D> t1) {
+  return svdupq_n_s64(t0, t1, t2, t3);
```
jan-wassenberg, line 607: CI noticed that we have extra t2, t3 here which do not exist :) Sorry I didn't catch this during review.

jan-wassenberg: (Also in the two functions below.)

johnplatts: Pushed the fix to SVE I64/U64/F64 Dup128VecFromValues

jan-wassenberg (approved):

PR #1802, opened 2023-10-06
Updated IsInf implementation in generic_ops-inl.h

johnplatts:
Updated IsInf implementation in generic_ops-inl.h to cast MaxExponentField&lt;T&gt;() to MakeUnsigned&lt;T&gt; to fix compilation errors.


jan-wassenberg (approved):
Thank you for fixing this :)

PR #1803, opened 2023-10-06
Hint compiler in case of armv7 + -mno-unaligned-access

malaterre:


jan-wassenberg:
Can you help me understand the goal here? __builtin_assume_aligned's argument is in bytes, so it seems we are promising 16 byte alignment, which is usually not the case.
Also, CopyBytes aka __builtin_memcpy does not require alignment, right?

jan-wassenberg (approved):
I see, thanks. Yes, makes sense to assume f16/bf16 are natively aligned unless someone is calling CopyBytes directly.

jan-wassenberg (approved):

PR #1806, opened 2023-10-07
Added new scalar float and integer ops in hwy/base.h

johnplatts:
Added the following new operations in hwy/base.h:
- RemoveCvRef&lt;T&gt; ScalarAbs(T val) - returns absolute value of floating point or integer
- bool ScalarIsNaN(T val) - returns true if `val` is NaN and false otherwise
- bool ScalarIsInf(T val) - returns true if `val` is infinity and false otherwise
- bool ScalarIsFinite(T val) - returns true if `val` is a finite value and false otherwise
- RemoveCvRef&lt;T&gt; ScalarCopySign(T magn, T sign) - returns the number with the magnitude of `magn` and sign of `sign`
- bool ScalarSignBit(T val) - returns the sign bit of `val`

All 6 of the above operations are supported for the `float`, `double`, `hwy::float16_t` and `hwy::bfloat16_t` types.

ScalarAbs and ScalarIsFinite are also supported also for integer types (including unsigned integer types), with ScalarAbs simply returning `val` if `val` is an unsigned integer (as unsigned integers are always non-negative) and ScalarIsFinite always returning true for integers.

Also updated hwy/ops/emu128_inl.h to use ScalarAbs, ScalarIsInf, ScalarIsNaN, and ScalarSignBit instead of std::abs/std::fabs, std::isinf, std::isnan, and std::signbit.

Also updated hwy/ops/emu128_inl.h to use sqrt implementation from hwy/ops/scalar_inl.h if HWY_NO_LIBCXX is defined.

Also updated Abs implementation in hwy/ops/scalar-inl.h to use ScalarAbs.

Also updated hwy/ops/scalar-inl.h to use ScalarAbs and ScalarSignBit instead of detail::Abs and detail::SignBit.

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -1574,6 +1574,20 @@ constexpr bool IsSigned<bfloat16_t>() {
   return true;
 }
 
+template <typename T, bool = IsInteger<T>() && !IsIntegerLaneType<T>()>
+struct MakeLaneTypeIfIntegerT {
+  using type = T;
+};
+
+template <typename T>
+struct MakeLaneTypeIfIntegerT<T, true> {
+  using type = hwy::If<IsSigned<T>(), SignedFromSize<sizeof(T)>,
+                       UnsignedFromSize<sizeof(T)>>;
+};
+
+template <typename T>
+using MakeLaneTypeIfInteger = typename MakeLaneTypeIfIntegerT<T>::type;
+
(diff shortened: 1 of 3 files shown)


jan-wassenberg (approved):
Nice, thanks for adding these. It is helpful for compatibility to avoid std::.

PR #1807, opened 2023-10-07
Made enhancements to x86 F16/F32/F64/I64 Abs

johnplatts:
Made the following enhancements to x86 Abs:
- Made F16/F32/F64 Abs generic for all vector lengths on SSE2/SSSE3/SSE4/AVX2/AVX3
- Made I64 Abs generic for all vector lengths on SSE2/SSSE3/SSE4/AVX2
- Reimplemented SSE2/SSSE3/SSE4/AVX2 I64 Abs to use IfNegativeThenElse


jan-wassenberg (approved):
Nice improvement, thanks, one less instruction and always good to reuse more code 👍

PR #1808, opened 2023-10-08
Made fixes to PPC float to int ConvertTo

johnplatts:
Made changes to PPC F32->U32, F64->I64, and F64->U64 ConvertTo routines to prevent GCC/Clang from incorrectly optimizing out the vec_ctu, vec_cts, or __builtin_convertvector operation if the values of any lanes of `v[i]` are outside of the range of the type being converted to.

Also updated the PPC F64->I64 and F64->U64 ConvertTo routines to use __builtin_convertvector on Clang to ensure that values are correctly converted as Clang F64->I64 vec_cts converts values using xvcvdpsxws instead of xvcvdpsxds if `__XL_COMPAT_ALTIVEC__` is defined and as Clang F64->I64 vec_ctu converts values using xvcvdpuxws instead of xvcvdpuxds if `__XL_COMPAT_ALTIVEC__` is defined.

Here is a link to a Compiler Explorer snippet that shows constant vectors with out-of-range values being incorrectly converted from F64->I64 or F64->U64 when compiled for little-endian POWER8 with Clang: https://godbolt.org/z/d5e8ExfcY

### DIFF
--- a/hwy/ops/ppc_vsx-inl.h
+++ b/hwy/ops/ppc_vsx-inl.h
@@ -3745,10 +3745,9 @@ HWY_API VFromD<D> ConvertTo(D /* tag */,
 }
 
 // Truncates (rounds toward zero).
-template <class D, typename FromT, HWY_IF_SIGNED_D(D), HWY_IF_FLOAT(FromT),
-          HWY_IF_T_SIZE_D(D, sizeof(FromT))>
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
👍 Nice that we can defeat these compiler footguns :)

PR #1810, opened 2023-10-09
Add Clang 13/14/15 and GCC 11/12 to GitHub workflow

johnplatts:
Here is an updated version of .github/workflows/build_test.yml that adds Clang 13/14/15 and GCC 11/12 to the Build/Test workflow.

The clang-13, clang-14, clang-15, g++-11, and g++-12 Build/Test steps require Ubuntu 22.04 as the clang-13, clang-14, clang-15, g++-11, and g++-12 packages are not available in the Ubuntu 20.04 repository.

Still need to use Ubuntu 20.04 to compile with Clang 6/7/8/9/10 and GCC 8 as Ubuntu 22.04 has removed the Clang 6/7/8/9/10 and GCC 8 packages.

It is possible to update .github/workflows/build_test.yml to have the Clang 11/12 and GCC 9/10 build/test steps to use Ubuntu 22.04 instead of Ubuntu 20.04.

Is it too much to add the extra Clang/GCC compilers that are available in Ubuntu 22.04 to the GitHub workflow?

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -94,6 +94,58 @@ jobs:
           cmake --build out
           ctest --test-dir out
 
+  cmake_ubuntu_2204:
+    name: Build and test ${{ matrix.name }}
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        include:
+          - name: Clang-13
+            extra_deps: clang-13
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Sounds good. So far we're meeting the CI deadlines, let's see if that still holds after these.

PR #1811, opened 2023-10-09
Fix SVE LoadDup128 compilation error if HWY_SVE_HAVE_BFLOAT16 is 1

johnplatts:
Added missing detail::LoadDupFull128 for BF16 vectors to hwy/ops/arm_sve-inl.h if HWY_SVE_HAVE_BFLOAT16 is 1.


jan-wassenberg (approved):
👍

PR #1821, opened 2023-10-13
Introduce a macro for reinterpret_cast special case

malaterre:
In some case we know it is acceptable to increase alignement. Hint the compiler when doing so.

Solve typical warning messages:

highway/hwy/ops/ppc_vsx-inl.h:619:35: error: cast from 'const long unsigned int* __restrict__' to 'const LoadRaw*' {aka 'const __vector(2) long unsigned int*'} increases required alignment of target type [-Werror=cast-align]
  619 |   const LoadRaw* HWY_RESTRICT p = reinterpret_cast<const LoadRaw*>(aligned);
      |                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -146,6 +146,10 @@ namespace hwy {
 #define HWY_ASSUME_ALIGNED(ptr, align) (ptr) /* not supported */
 #endif
 
+// Special case to increases required alignment
+#define HWY_RCAST_ALIGNED(type, ptr) \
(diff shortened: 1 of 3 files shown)


jan-wassenberg (approved):
Nice, thanks. I have verified Load8Bytes is only called for aligned tables.

PR #1845, opened 2023-10-30
Fixed compilation error in hwy/contrib/matvec/matvec-inl.h

johnplatts:
Remove unused VF typedef in hwy/contrib/matvec/matvec-inl.h to fix compilation errors in build/test


jan-wassenberg (approved):
Thanks for fixing :) FYI I am out of office and will be fully back on Wed morning.

PR #1846, opened 2023-10-30
Made fixes to loops in hwy/contrib/algo/copy-inl.h, find-inl.h, and transform-inl.h

johnplatts:
Replaced `idx + N <= count` with  `idx <= count - N` in copy-inl.h, find-inl.h, transform-inl.h in hwy/contrib/algo as it is possible for `idx + N` to overflow if `idx > LimitsMax<size_t>() - N` is true, whereas `idx <= count - N` is guaranteed to not overflow if `count >= N` is true.

Also added `if (count >= N)` checks to ensure that the `for (; idx <= count - N; idx += N)` loops in copy-inl.h, find-inl.h, transform-inl.h are only executed if `count >= N` is true.

The changes to the loops in copy-inl.h, find-inl.h, and transform-inl.h also work around `-Waggressive-loop-optimizations` warnings that were encountered when compiling hwy/contrib/matvec/matvec_test.cc on PPC with GCC.

### DIFF
--- a/hwy/contrib/algo/copy-inl.h
+++ b/hwy/contrib/algo/copy-inl.h
@@ -40,8 +40,10 @@ void Fill(D d, T value, size_t count, T* HWY_RESTRICT to) {
   const Vec<D> v = Set(d, value);
 
   size_t idx = 0;
-  for (; idx + N <= count; idx += N) {
-    StoreU(v, d, to + idx);
+  if (count >= N) {
+    for (; idx <= count - N; idx += N) {
+      StoreU(v, d, to + idx);
+    }
   }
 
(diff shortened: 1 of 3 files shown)


johnplatts:
> Thanks @johnplatts. Are you happy with the codegen for the loop, in particular whether the compiler avoids another test/branch before the for loop? If not, should we change the loops into do/while?

GCC 5.1 and later, Clang 3.2 and later, and MSVC 19.14 and later will generate the same code for a do-while loop when optimizations are enabled as demonstrated in a snippet over at https://godbolt.org/z/f1Mqx39n9.

jan-wassenberg (approved):
Thanks for confirming, that's great :)

PR #1899, opened 2023-12-14
Fix compilation errors with HWY_SCALAR target

johnplatts:
Replaced `Vec1<T>(0)` with `Vec1<T>(ConvertScalarTo<T>(0))` in the IfThenElseZero, IfThenZeroElse, and ZeroIfNegative implementations in hwy/ops/scalar-inl.h to fix compilation errors with HWY_SCALAR target.

Also updated Zero function in hwy/ops/scalar-inl.h to return `Vec1<T>(ConvertScalarTo<T>(0))`.


jan-wassenberg (approved):
Nice, thank you :)

Issue #1895, opened 2023-12-11
Adding support for IBM/Z14 and onwards


seiko2plus:
### Brief

IBM/Z has introduced SIMD extensions comparable to those in PowerPC64's `VSX`, starting with the IBM/Z13 a.k.a `VX`.
Nonetheless, implementing support for the `Z13` is not applicable, this is because Highway assumes the support of single-precision operations and any blind conversions involving from/to double-precision might lead to decreased performance compared to scalar operations.
Therfore, it is more appropriate to initiate support from the `IBM/Z14` onwards.

### Motivation

Recently, there have been discussions in Numpy advocating for the abandonment of its own SIMD framework(universal intrinsics) and replacing it with Googls's Highway. NumPy provides SIMD support for IBMZ starting from `IBM/Z13`, any future move to Highway will require the support of this architecture.

### References
- [GCC build options](https://gcc.gnu.org/onlinedocs/gcc/S_002f390-and-zSeries-Options.html)
- [NumPy's Implmentation](https://github.com/numpy/numpy/tree/main/numpy/_core/src/common/simd/vec) supports both `VSX` and `VX` under the same path.

johnplatts:
Implementing the HWY_Z13 target requires using the `+`, `-`, `*`, and `/` operators available with GCC/Clang vector extensions instead of the vec_add, vec_sub, vec_mul, and vec_div intrinsics that are available on Altivec/VSX.

The HWY_Z13 target also needs to avoid using the POWER-specific inline assembly that is there in hwy/ops/ppc_vsx.h to work around suboptimal GCC/Clang code generation and GCC/Clang compiler bugs on PPC8/PPC9/PPC10.

The Z13 architecture does support the __vector double type and __vector double arithmetic ops, and a Compiler Explorer snippet that works on both Z13 and POWER7 can be found at https://godbolt.org/z/xWG6MdYcf.

It is also possible to re-implement the PPC8/PPC9/PPC10/WASM targets (and the SSE2/SSSE3/SSE4/AVX2/AVX3/NEON targets when compiled with GCC or Clang) to use GCC/Clang vector extensions for the Load, LoadU, Store, StoreU, Add, Sub, Mul, Div, Set, Dup128VecFromValues, And, AndNot, Or, Xor, Not, BitCast, and ResizeBitCast operations.

[...]

PR #1906, opened 2023-12-19
Added HWY_Z14 and HWY_Z15 targets

johnplatts:
Resolves issue #1895.

The HWY_Z14 and HWY_Z15 targets compile successfully with GCC 11, GCC 12, Clang 16, and Clang 17. However, some of the tests fail on the HWY_Z14 and HWY_Z15 targets when compiled with Clang 16 or Clang 17 with optimizations enabled.

The HWCAP_S390_VXRS_EXT2 bit (which is HWCAP bit for the Z15 vector instructions) is not set by the result of the `getauxval(AT_HWCAP)` call on QEMU 8.1.3 on Ubuntu 22.04, even when the Z15 vector instructions are enabled.

Here is the cmake command line to build only the HWY_Z14 target:
```CC=s390x-linux-gnu-gcc-11 CXX=s390x-linux-gnu-g++-11 cmake .. -DCMAKE_C_COMPILER_TARGET="s390x-linux-gnu" -DCMAKE_CXX_COMPILER_TARGET="s390x-linux-gnu" -DCMAKE_C_FLAGS='-march=z14 -mzvector' -DCMAKE_CXX_FLAGS='-march=z14 -mzvector -DHWY_DISABLED_TARGETS="(HWY_SCALAR|HWY_EMU128|HWY_Z15)" -DHWY_COMPILE_ONLY_STATIC=1' -DCMAKE_CROSSCOMPILING=true -DCMAKE_CROSSCOMPILING_EMULATOR="/usr/bin/qemu-s390x;-L;/usr/s390x-linux-gnu"```

Here is the cmake command line to build only the HWY_Z15 target (need to add the `-DTOOLCHAIN_MISS_SYS_AUXV_H=1` option to CMAKE_CXX_FLAGS to disable HWCAP detection to work around a QEMU bug):
```CC=s390x-linux-gnu-gcc-11 CXX=s390x-linux-gnu-g++-11 cmake .. -DCMAKE_C_COMPILER_TARGET="s390x-linux-gnu" -DCMAKE_CXX_COMPILER_TARGET="s390x-linux-gnu" -DCMAKE_C_FLAGS='-march=z15 -mzvector' -DCMAKE_CXX_FLAGS='-march=z15 -mzvector -DHWY_DISABLED_TARGETS="(HWY_SCALAR|HWY_EMU128|HWY_Z14)" -DHWY_COMPILE_ONLY_STATIC=1' -DCMAKE_CROSSCOMPILING=true -DCMAKE_CROSSCOMPILING_EMULATOR="/usr/bin/qemu-s390x;-cpu;max,vxeh2=on;-L;/usr/s390x-linux-gnu"```

The Z14/Z15 SIMD instruction set also lacks instructions for I8/U8/I16/U16 SaturatedAdd/SaturatedSub, F32 ApproximateReciprocal, and F32 ApproximateReciprocalSqrt that most other SIMD instruction sets have.

### CODE REVIEW

File: hwy/ops/ppc_vsx-inl.h [hunk 1]
```diff
@@ -3731,44 +3923,85 @@ HWY_API Vec64<uint32_t> DemoteTo(D /* tag */, Vec128<double> v) {
...
 }
 
+}  // namespace detail
+#endif  // HWY_S390X_HAVE_Z14
+
 template <class D, HWY_IF_V_SIZE_D(D, 4), HWY_IF_F32_D(D)>
-HWY_API Vec32<float> DemoteTo(D /* tag */, Vec64<int64_t> v) {
+HWY_API Vec32<float> DemoteTo(D df32, Vec64<int64_t> v) {
+#if HWY_S390X_HAVE_Z14
+  return DemoteTo(df32, detail::ConvToF64WithRoundToOdd(v));
```
jan-wassenberg, line 3954: I'm curious why round to odd?

jan-wassenberg (approved):
Wow! I am super impressed. Thank you @johnplatts for implementing this already :-o
Nice that Z intrinsics have operators. Curious that they lack saturated arithmetic and rsqrt, but your emulations of them look good.

hwy/ops/ppc_vsx-inl.h
johnplatts: Doing an I64->F64 or U64->F32 conversion using round to odd ensures that the I64->F32 or U64->F32 DemoteTo conversion produces the same result as `static_cast<float>(v[i])`.

A round-to-odd integer to floating-point conversion is equivalent to a truncating integer to floating-point conversion followed by a bitwise OR of the bit representation by 1 if the truncated floating-point value is not equal to the original value.

The SSE2/SSSE3/SSE4/AVX2/A64 NEON/WASM UI64->F32 DemoteTo implementations first do a round-to-odd conversion of `v[i]` to a F64 followed by a demotion of the rounded-to-odd F64 conversion result to float to ensure that SSE2/SSSE3/SSE4/AVX2/A64 NEON/WASM UI64->F32 matches the behavior of AVX3/PPC/RVV/SVE UI64->F32 DemoteTo.

A round-to-odd I64/U64->F64 conversion avoids the double-rounding that can occur with `static_cast<float>(static_cast<double>(v[i]))` or `DemoteTo(df32, ConvertTo(df64, v))`.

Here is the round-to-odd I64->F64 conversion snippet from the SSE2/SSSE3/SSE4/AVX2 I64->F32 DemoteTo implementation in hwy/ops/x86_128-inl.h:
```
  const auto k2p64_63 = Set(df64, 27670116110564327424.0);
  const auto f64_hi52 =
      Xor(BitCast(df64, ShiftRight<12>(BitCast(du64, v))), k2p64_63) - k2p64_63;
  const auto f64_lo12 =
      PromoteTo(df64, BitCast(di32, And(TruncateTo(du32, BitCast(du64, v)),
                                        Set(du32, uint32_t{0x00000FFF}))));

  const auto f64_sum = f64_hi52 + f64_lo12;
  const auto f64_carry = (f64_hi52 - f64_sum) + f64_lo12;

  const auto f64_sum_is_inexact =
      ShiftRight<63>(BitCast(du64, VecFromMask(df64, f64_carry != Zero(df64))));
  const auto f64_bits_decrement =
      And(ShiftRight<63>(BitCast(du64, Xor(f64_sum, f64_carry))),
          f64_sum_is_inexact);

  const auto adj_f64_val = BitCast(
      df64,
      Or(BitCast(du64, f64_sum) - f64_bits_decrement, f64_sum_is_inexact));
```

Z14 can carry out the UI64->F64 round-to-odd conversion using a single vcdgb or vcdlgb instruction, whereas the SSE2 UI64->F64 round-to-odd conversion using the above algorithm requires 18 instructions.

jan-wassenberg: I see, thank you. That's handy :)

jan-wassenberg (approved):

johnplatts:
There is a fix for the U16/I16/F16/BF16 Expand operation on big-endian targets (including Z14/Z15) in pull request #1929.

PR #1907, opened 2023-12-19
GCC AVX2/AVX3 compilation fixes for F16 Load/Store

johnplatts:
Fixes compile errors for F16 Load/LoadU/Store/StoreU on AVX2/AVX3 with GCC


jan-wassenberg (approved):
Thank you for fixing these :)

PR #1911, opened 2023-12-20
Made fixes to PreventElision on PPC

johnplatts:
Updated PreventElision implementation in hwy/base.h to work around a bug that causes TestBroadcast to fail when compiled for PPC with Clang.


jan-wassenberg (approved):
👍

PR #1919, opened 2024-01-01
Made fixes to SCALAR, EMU128, NEON, and PPC float to int conversions

johnplatts:
Updated SCALAR/EMU128 F->UI PromoteTo/ConvertTo/DemoteTo to ensure that undefined behavior does not occur if the source vector contains any NaN values.

On PPC8/PPC9/PPC10, also updated F->UI PromoteTo/ConvertTo/DemoteTo operations that use VSX F->UI conversion instructions to zero out any NaN source values unless HWY_DISABLE_PPC_VSX_QEMU_F2I_WORKAROUND is defined to work around a bug in QEMU 7/8 where a NaN value in the source floating-point vector can cause subsequent lanes to be incorrectly converted to an integer.

Also updated NEON F->UI PromoteTo/ConvertTo/DemoteTo operations to use inline assembly when compiled for ArmV7 with Clang or when compiled for AArch64 with Clang 11 or earlier to avoid undefined behavior if any of the source values are outside of the range of the result type.

### DIFF
--- a/hwy/tests/convert_test.cc
+++ b/hwy/tests/convert_test.cc
@@ -21,6 +21,7 @@
 #define HWY_TARGET_INCLUDE "tests/convert_test.cc"
 #include "hwy/foreach_target.h"  // IWYU pragma: keep
 #include "hwy/highway.h"
+#include "hwy/nanobenchmark.h"
 #include "hwy/tests/test_util-inl.h"
 
 HWY_BEFORE_NAMESPACE();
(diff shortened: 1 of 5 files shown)


hwy/tests/convert_test.cc
```diff
@@ -832,6 +832,227 @@ HWY_NOINLINE void TestAllFloatFromUint() {
...
+    using TTo_I = MakeSigned<TTo>;
+
+    constexpr TF kMinOutOfRangePosVal =
+        static_cast<TF>((-static_cast<TF>(LimitsMin<TTo_I>())) *
+                        static_cast<TF>(IsSigned<TTo>() ? 1 : 2));
+    static_assert(kMinOutOfRangePosVal > static_cast<TF>(0),
+                  "kMinOutOfRangePosVal > 0 must be true");
+
+    TU scalar_non_elided_zero = static_cast<TU>(0);
+#if HWY_COMPILER_MSVC
```
jan-wassenberg, line 909: Consider instead using Unpredictable1() from nanobenchmark.h? (Then subtract one from that to get zero)

hwy/tests/convert_test.cc
```diff
@@ -832,6 +832,227 @@ HWY_NOINLINE void TestAllFloatFromUint() {
...
+    const auto neg_inf = Neg(pos_inf);
+
+    VerifyNonFiniteF2I(df, pos_nan, __FILE__, __LINE__);
+    VerifyNonFiniteF2I(df, neg_nan, __FILE__, __LINE__);
+    VerifyNonFiniteF2I(df, pos_inf, __FILE__, __LINE__);
+    VerifyNonFiniteF2I(df, neg_inf, __FILE__, __LINE__);
+
+    TI non_elided_one = static_cast<TI>(1);
+#if HWY_COMPILER_MSVC
+    non_elided_one = *reinterpret_cast<const volatile TI*>(&non_elided_one);
```
jan-wassenberg, line 977: Here also Unpredictable1, and also below?

jan-wassenberg (commented):
Happy new year! Nice, thank you for making these improvements :)

hwy/tests/convert_test.cc
johnplatts: Already replaced the above code with Unpredictable1()

hwy/tests/convert_test.cc
johnplatts: Already replaced the above code with `Unpredictable1() - 1`

jan-wassenberg (approved):
Nice, thanks for making the change :)

PR #1928, opened 2024-01-09
Added integer MaskedDivOr and MaskedModOr ops plus unsigned Div bug fix

johnplatts:
Added implementation of MaskedDivOr for integer vectors on AVX3/RVV/SVE, and added MaskedModOr operation on all targets.

Also added TestAllIntegerDivMod in hwy/tests/masked_arithmetic_test.cc that tests the behavior of MaskedDivOr and MaskedModOr for integer vector types.

Also fixed bug in default implementation of U8/U16/U32/U64 Div in hwy/ops/generic_ops-inl.h.


jan-wassenberg (approved):
Nice to also support MaskedDivOr for integers, thanks for adding it 👍

PR #1929, opened 2024-01-09
Fixed endianness bug in U16/I16/F16/BF16 Expand in generic_ops-inl.h

johnplatts:
Fixed endianness bug in U16/I16/F16/BF16 Expand implementation in generic_ops-inl.h.

The U16/I16/F16/BF16 Expand implementation in generic_ops-inl.h now has correct behavior on both big-endian targets and little-endian targets with the fix in this pull request.


jan-wassenberg (approved):
👍

PR #1935, opened 2024-01-12
Split Abort definition from targets.cc

Mousius:
This allows compilation without the dynamic dispatch machinery and gives an example of the interface for the `Abort` function should you wish to write your own.


jan-wassenberg (changes requested):
Nice, thanks for doing this already :)
Would you mind also adding this new abort.cc to the list of files in hwy.gni, which is the build system used by Chrome?

Mousius:
I've added the implementation of the hook here as well @jan-wassenberg, and some tests to check it works as I think we discussed 😸

jan-wassenberg (approved):
Very nice, thanks for also adding the handler :)

jan-wassenberg:
I'm sorry this one has not landed yet. The internal CI failed the merge here because of the formatting of the BUILD file. Unfortunately the change has already been garbage-collected so I do not see the error. It is often whitespace type, indentation or sort order. I don't see the problem here; when adding the test, the list is already unsorted, but as far as the tooling is concerned the list of tests is just an arbitrary list and I don't think we enforce sort order.

We could try to merge again, restarting..

Mousius:
@jan-wassenberg hmm, maybe it just got confused. I also saw some unrelated errors so I've rebased, so it should be a clean diff now 😸

jan-wassenberg (approved):
Thanks for updating. Let's try again :)

jan-wassenberg:
OK, I see now some of the CI failures:
```
abort.cc:44: error: missing '#include <sanitizer/common_interface_defs.h>'; '__sanitizer_print_stack_trace' must be declared before it is used
   44 |   __sanitizer_print_stack_trace();
```

I think the following has to move from targets.cc to abort.cc, and come after including hwy/base.h:
```
#if HWY_IS_ASAN || HWY_IS_MSAN || HWY_IS_TSAN
#include "sanitizer/common_interface_defs.h"  // __sanitizer_print_stack_trace
#endif
```

Mousius:
> OK, I see now some of the CI failures:
> 
> ```
> abort.cc:44: error: missing '#include <sanitizer/common_interface_defs.h>'; '__sanitizer_print_stack_trace' must be declared before it is used
>    44 |   __sanitizer_print_stack_trace();
> ```
> 
> I think the following has to move from targets.cc to abort.cc, and come after including hwy/base.h:
> 
> ```
> #if HWY_IS_ASAN || HWY_IS_MSAN || HWY_IS_TSAN
> #include "sanitizer/common_interface_defs.h"  // __sanitizer_print_stack_trace
> #endif
> ```

Yikes, good catch! I've pushed a patch 😸

Mousius:
Confused by:
```
ERROR: no such package 'third_party/highway': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
```

Staring intently at the `BUILD` changes reveals nothing that could break this.

jan-wassenberg (approved):

jan-wassenberg:
Nice, rerunning the CI.
Don't worry about the third_party error message, this seems to have been caused by my earlier change (https://github.com/google/highway/actions/runs/7986902373/job/21808245834), not your current one.

Mousius:
@jan-wassenberg I used `GTEST_HAS_DEATH_TEST`, which should be more robust. Did it fail again? If so, I'll add the specific WASM check as well.

jan-wassenberg:
Nice, that is a better solution than what I proposed, and works fine. We are almost there, just one more issue from msvc:
```
z:\build\work\f31a378485b250bd0d597f1bc48169615cc9\google3\third_party\highway\hwy\abort_test.cc:16
Death test: ::hwy::Abort("z:\\build\\work\\f31a378485b250bd0d597f1bc48169615cc9\\google3\\third_party\\highway\\hwy\\abort_test.cc", 16, "Test %s","Abort")
    Result: died but not with expected error.
  Expected: contains regular expression "Abort at z:\\build\\work\\f31a378485b250bd0d597f1bc48169615cc9\\google3\\third_party\\highway\\hwy\\abort_test.cc:16: Test Abort"
Actual msg:
[  DEATH   ] E0000 00:00:1708522008.549367      37 re2.cc:243] Error parsing 'Abort at z:\build\work\f31a378485b250bd0d597f1bc48169615cc9\google3\third_party\highway\hwy\abort_te...': invalid escape sequence: \g
[  DEATH   ] Abort at z:\build\work\f31a378485b250bd0d597f1bc48169615cc9\google3\third_party\highway\hwy\abort_test.cc:16: Test Abort
[  DEATH   ] 
```
It is not clear to me why __FILE__ should contain double slashes like this :/ One possible solution could be to extract only the filename from __FILE__ both in Abort and the test, what do you think?

jan-wassenberg (approved):

hwy/abort_test.cc
```diff
@@ -0,0 +1,63 @@
+// Copyright 2024 Arm Limited and/or its affiliates <open-source-office@arm.com>
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "hwy/abort.h"
+
+#include "gtest/gtest.h"
```
jan-wassenberg, line 7: Please change to "hwy/tests/hwy_gtest.h" - our internal MSVC is no longer able to build with GTest.

jan-wassenberg:
Thanks @Mousius  for updating! I've commented with a small change required for CI to pass.

jan-wassenberg (dismissed):
👍

jan-wassenberg:
Apologies, missed something the CI found. Our gtest wrapper does not support ASSERT_EQ in abort_test. Would you mind changing to HWY_ASSERT_EQ?

Mousius:
> Apologies, missed something the CI found. Our gtest wrapper does not support ASSERT_EQ in abort_test. Would you mind changing to HWY_ASSERT_EQ?

No worries, easy fix 😸

Mousius:
I spoke too soon 😿 looks like this is going to be trickier when comparing pointer types 🤔

jan-wassenberg:
Ah, right. The simplest fix would be to just HWY_ASSERT(expected == actual), because we're not very likely to want to print the actual value in case of failure.

jan-wassenberg:
Thanks for making the change. Rerunning CI :)

hwy/abort_test.cc
```diff
@@ -0,0 +1,65 @@
...
+
+  HWY_ASSERT(SetAbortFunc(FirstHandler) == nullptr);
+  HWY_ASSERT(GetAbortFunc() == FirstHandler);
+  HWY_ASSERT(SetAbortFunc(SecondHandler) == FirstHandler);
+  HWY_ASSERT(GetAbortFunc() == SecondHandler);
+  HWY_ASSERT(SetAbortFunc(nullptr) == SecondHandler);
+  HWY_ASSERT(GetAbortFunc() == nullptr);
+}
+
+}  // namespace hwy
```
jan-wassenberg, line 65: The new GTest replacement requires a `HWY_TEST_MAIN();` at the end of this file :)

jan-wassenberg:
Looks like non-MSVC compilers are good so far, MSVC requires an extra line at the end of the test, added comment :)

Mousius:
> Looks like non-MSVC compilers are good so far, MSVC requires an extra line at the end of the test, added comment :)

Argh, sorry, I did look at the other test files but missed that line 😿

jan-wassenberg:
No worries :)

PR #1936, opened 2024-01-14
Fixed EMU128 InterleaveLower with D parameter

johnplatts:
Updated `InterleaveLower(D, VFromD<D>, VFromD<D>)` overload in hwy/ops/emu128-inl.h to fix compilation errors in the case where `D().Pow2() != 0` is true.


jan-wassenberg (approved):
Thank you for fixing this :) I believe ZipLower/Upper in that file also have the same DFromV issue, would you like to change those as well or should I?

PR #1946, opened 2024-01-18
Reimplemented F32->F16 DemoteTo in generic_ops-inl.h

johnplatts:
The implementation of F32->F16 DemoteTo implementation in hwy/ops/generic_ops-inl.h in this pull request is much more efficient than the current F32->F16 DemoteTo implementation in hwy/ops/generic_ops-inl.h on some targets, including SSE2/SSSE3/SSE4.

A snippet showing the difference in codegen between the two F32->F16 DemoteTo implementations can be found at https://godbolt.org/z/snqa49Wj6.

The updated F32->F16 implementation in this pull request also uses a F32 Add operation to shift the mantissa bits down, which also rounds the mantissa to the nearest value (instead of simply truncating the mantissa).

Shifting the mantissa bits down using an F32 Add operation also avoids the overhead that is there with Shl/Shr on SSE2/SSSE3/SSE4, which lack instructions for per-lane variable shifts.

### DIFF
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -2853,37 +2853,143 @@ HWY_API VFromD<D> PromoteTo(D df32, VFromD<Rebind<float16_t, D>> v) {
 
 template <class D, HWY_IF_F16_D(D)>
 HWY_API VFromD<D> DemoteTo(D df16, VFromD<Rebind<float, D>> v) {
-  const RebindToUnsigned<decltype(df16)> du16;
-  const Rebind<uint32_t, decltype(df16)> du32;
-  const RebindToSigned<decltype(du32)> di32;
-  using VU32 = VFromD<decltype(du32)>;
-  using VI32 = VFromD<decltype(di32)>;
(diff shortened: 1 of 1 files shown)


hwy/ops/generic_ops-inl.h
```diff
@@ -2853,37 +2853,143 @@ HWY_API VFromD<D> PromoteTo(D df32, VFromD<Rebind<float16_t, D>> v) {
...
+  // mantissa bits (including the implicit 1 bit) to the left of the
+  // F32 mantissa bits in rounded_val[i] since 23 - 10 is equal to 13
+
+  // The biased exponent of round_incr[i] needs to be at least 126 as
+  // (-14) + 13 + 127 is equal to 126
+
+  // We also want to biased exponent of round_incr[i] to be less than or equal
+  // to 255 (which is equal to MaxExponentField<float>())
+
+  // The biased F64 exponent of round_incr is equal to
```
jan-wassenberg, line 2876: F64 -> F32?

jan-wassenberg (approved):
Wow, that's about half the code size. Nice improvement, thank you :D

PR #1948, opened 2024-01-19
Enable F16 for AVX3_SPR when compiled with GCC

johnplatts:
This pull request enables F16 support on the AVX3_SPR target when compiled with GCC.

This pull request also includes fixes to some of the operations in x86_128-inl.h, x86_256-inl.h, and x86_512-inl.h to allow the AVX3_SPR target to successfully compile with F16 support enabled.

There is still a compilation warning in hwy/tests/reduction_test.cc that needs to be fixed when AVX3_SPR is compiled with F16 support enabled.

All of the Highway unit tests pass on the AVX3_SPR target with F16 support enabled with the changes in this pull request when compiled with GCC 13.2.0.

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -1038,6 +1038,19 @@ HWY_API HWY_BITCASTSCALAR_CONSTEXPR To BitCastScalar(const From& val) {
 #endif
 #endif  // HWY_HAVE_SCALAR_F16_OPERATORS
 
+namespace detail {
+
+template <class T, class TVal = RemoveCvRef<T>>
+struct NativeSpecialFloatToWrapperT {
+  using type = T;
+};
+
+template <class T>
+using NativeSpecialFloatToWrapper =
+    typename NativeSpecialFloatToWrapperT<T>::type;
+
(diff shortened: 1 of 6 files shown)


jan-wassenberg (approved):
Nice, this is great to see. F16 has required quite a few changes to SVE and the tests recently, thank you for also updating the x86 code!

jan-wassenberg (changes requested):
I see that we have some remaining [compile errors](https://github.com/google/highway/actions/runs/7584511366/job/20658325273).
Likely the fix is to change any `static_cast<T>`, where T could be f16, to `ConvertScalarTo`. Would you mind updating the code before we merge?

jan-wassenberg:
Thanks for making the changes :) I think there are a few [more issues](https://github.com/google/highway/actions/runs/7587383738/job/20667614422), including unroller_test somehow using TypeTag on the native float type, instead of the hwy::float16_t type which we specialize for.

jan-wassenberg (approved):
Great :) Thanks for updating ScalarAbs etc.

PR #1952, opened 2024-01-23
Added mask slide up/down ops

johnplatts:
The SlideMask1Up, SlideMask1Down, SlideMaskUpLanes, and SlideMaskDownLanes operations were added as it is more efficient to slide a mask up or down by doing a left or right shift of `m.raw` on AVX3.


hwy/ops/x86_256-inl.h
```diff
@@ -1074,6 +1074,28 @@ HWY_API MFromD<D> UpperHalfOfMask(D /*d*/, MFromD<Twice<D>> m) {
...
+#if HWY_COMPILER_HAS_MASK_INTRINSICS
+  return MFromD<D>{
+      static_cast<RawM>(_kshiftli_mask32(static_cast<__mmask32>(m.raw), 1))};
+#else
+  return MFromD<D>{static_cast<RawM>(static_cast<uint32_t>(m.raw) << 1)};
+#endif
+}
+
+template <class D, HWY_IF_LANES_D(D, 32)>
+HWY_API MFromD<D> SlideMask1Down(D /*d*/, MFromD<D> m) {
```
jan-wassenberg, line 1089: Do you think we should specialize for 1 byte (32-bit mask) vs larger lanes (16-bit mask)?
Also in x86-128 and x86-512?

jan-wassenberg (commented):
Nice, looks useful, thanks for adding :) One question:

PR #1957, opened 2024-01-26
Make MaskFalse available to non-test code on all targets

johnplatts:
This pull request makes the `MaskFalse(d)` op available to non-test code on all targets.

Also reimplemented the `MaskFalse(d)` op on targets other than RVV.

Also added the TestAllMaskFalse test that checks that all of the bits of `MaskFalse(d)` are false, including any bits past the first `Lanes(d)` lanes of `MaskFalse(d)`.

### DIFF
--- a/hwy/tests/mask_test.cc
+++ b/hwy/tests/mask_test.cc
@@ -28,6 +28,79 @@ namespace hwy {
 namespace HWY_NAMESPACE {
 
 // All types.
+struct TestMaskFalse {
+  template <typename T, class D>
+  HWY_NOINLINE void operator()(T /*unused*/, D d) {
+#if HWY_HAVE_SCALABLE || HWY_TARGET == HWY_SVE_256 || \
+    HWY_TARGET == HWY_SVE2_128 || HWY_TARGET == HWY_SCALAR
+    const DFromV<Vec<D>> d2;
+#else
+    constexpr size_t kMinD2Lanes =
+        ((HWY_TARGET == HWY_NEON || HWY_TARGET == HWY_NEON_WITHOUT_AES) ? 8
+                                                                        : 16) /
+        sizeof(T);
+    const FixedTag<T, HWY_MAX(HWY_MAX_LANES_D(D), kMinD2Lanes)> d2;
+#endif
+
+    static_assert(d2.MaxBytes() >= d.MaxBytes(),
+                  "d2.MaxBytes() >= d.MaxBytes() should be true");
+
+    const size_t N = Lanes(d);
+    const size_t N2 = Lanes(d2);
+    HWY_ASSERT(N2 >= N);
+
+    auto expected1 = AllocateAligned<T>(N);
+    auto expected2 = AllocateAligned<T>(N2);
+    auto actual1 = AllocateAligned<T>(N);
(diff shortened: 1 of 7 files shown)


hwy/tests/mask_test.cc
```diff
@@ -28,6 +28,79 @@ namespace hwy {
...
+    ZeroBytes(expected2.get(), sizeof(T) * N2);
+
+    memset(actual1.get(), 0xFF, sizeof(T) * N);
+    memset(actual2.get(), 0xFF, sizeof(T) * N2);
+
+    // If possible, use a compiler memory barrier to prevent the compiler from
+    // reordering the above ZeroBytes and memset operations with the Store
+    // operations below
+
+#if HWY_COMPILER_MSVC && !defined(HWY_NO_LIBCXX)
```
jan-wassenberg, line 68: We do have HWY_FENCE for this (compiler, not CPU) but I'm not sure about its implementation and efficacy. IIRC clang has ignored an asm with memory clobber before.
For use in a test, we could consider adding a HWY_MEMORY_BARRIER or MemoryBarrier to cache_control.h that actually generates a barrier instruction, preferably in a way the compiler understands so that it also serves as a compiler fence?

jan-wassenberg (approved):
Nice 👍 Here's a suggestion for the fencing, let me know if you'd like to merge the PR as-is.

hwy/tests/mask_test.cc
johnplatts: It is okay to merge this pull request as-is.

We want to ensure that the initialization of `actual1` and `actual2` to all ones by `memset` in TestMaskFalse is not optimized out by the compiler and that the compiler emits the instructions for initializing `actual1` and `actual2` prior to the instructions that store `VecFromMask(d, MaskFalse(d))` to `actual1` and `ResizeBitCast(d2, VecFromMask(d, MaskFalse(d)))` to `actual2`.

A compiler memory barrier that prevents `memset(actual1.get(), 0xFF, sizeof(T) * N)` from being optimized out or reordered with subsequent stores to `actual1` is sufficient in TestMaskFalse as all of the accesses to `actual1` and `actual2` in TestMaskFalse are on the same thread.

I did some testing with a similar snippet over at https://godbolt.org/z/oaWj7sKf4 with various GCC and Clang versions, and the `asm volatile("" ::: "memory")` barrier in that snippet did prevent the `memset(actual1.get(), 0xFF, sizeof(T) * N)` from being optimized out (even at the `-O3` optimization level) since `actual1` was allocated with `AllocateAligned<T>(N)`.

I also did some testing with a similar snippet over at https://godbolt.org/z/3ETsETbcd that used native SIMD intrinsics instead of the Google Highway library with various MSVC versions, and the `std::atomic_signal_fence(std::memory_order_seq_cst))` barrier did prevent the `memset(actual1.get(), 0xFF, sizeof(T) * N)` from being optimized out with MSVC (even at the `-Ox` optimization level) since `actual1` was allocated with `AllocateAligned<T>(N)`.

jan-wassenberg: Thanks for investigating. Looks like the atomic_signal_fence also works for clang, so let's keep the base.h as is and we'll merge this PR.

jan-wassenberg (approved):

PR #1958, opened 2024-01-27
Fixed PrintTargets in list_targets.cc

johnplatts:
Updated PrintTargets in list_targets.cc to avoid undefined behavior if the the parameter has the sign bit set as it is possible for HWY_BROKEN_TARGETS and HWY_DISABLED_TARGETS to have the sign bit set if the HWY_BROKEN_TARGETS or HWY_DISABLED_TARGETS macros are defined by the user prior to including hwy/highway.h.


jan-wassenberg (approved):
Nice catch, thanks :)

PR #1959, opened 2024-01-27
Fix PPC10 Div/Mod inline assembly

johnplatts:
Fixes to PPC10 Div/Mod inline assembly to fix compilation errors with Clang


jan-wassenberg (approved):
👍

PR #1960, opened 2024-01-27
Fixed PPC InsertLane

johnplatts:
Replaced hwy::CopySameSize with BitCastScalar in PPC InsertLane op to fix compilation error.


jan-wassenberg (approved):
👍

PR #1961, opened 2024-01-27
Fix operator / in wasm_256-inl.h

johnplatts:
Updated operator / in wasm_256-inl.h to fix compilation error for 256-bit integer vectors.


jan-wassenberg (approved):
Thanks for fixing :)

PR #1969, opened 2024-01-31
Added AddSub, MulAddSub, integer MulSub, and integer NegMulSub ops

johnplatts:
Added the AddSub and MulAddSub operations for both integer and floating point vectors.

The AddSub operation returns `a[i] - b[i]` in the even lanes and `a[i] + b[i]` in the odd lanes.

The MulAddSub operation returns `mul[i] * x[i] - sub_or_add[i]` in the even lanes and `mul[i] * x[i] + sub_or_add[i]` in the odd lanes.

Also added the MulSub and NegMulSub ops for integer vectors.


jan-wassenberg (approved):
Nice :)

PR #1970, opened 2024-01-31
Reformatted TestIntegerDiv in arithmetic_test.cc

johnplatts:
Reformatted TestIntegerDiv in hwy/tests/arithmetic_test.cc using clang-format


jan-wassenberg (approved):
👍

PR #1978, opened 2024-02-09
Added F16->F64 PromoteTo and F64->F16 DemoteTo ops

johnplatts:
Added F64->F16 DemoteTo op as AVX3_SPR/PPC9/PPC10/SVE have F64->F16 demotion instructions and to ensure that the F64->F16 demotion is correctly rounded on all targets as simply doing a `DemoteTo(df16, DemoteTo(Rebind<float, decltype(df16)>(), v))` might result in an incorrectly rounded value.

Added F16->F64 PromoteTo op as AVX3_SPR/SVE have instructions for F16->F64 promotion.


jan-wassenberg (approved):
Nice :)

PR #1979, opened 2024-02-11
Fix FindUnit constructor in unroller_test.cc

johnplatts:
Updated FindUnit constructor in unroller_test.cc, which fixes compilation errors with GCC 11 in C++20 mode.


jan-wassenberg (approved):
Thanks for fixing :)

PR #1983, opened 2024-02-12
Updated GitHub workflow to check that Highway successfully compiles in C++11/14/17/20 mode

johnplatts:
Updated GitHub workflow to check that Highway successfully compiles in C++11/14/17/20 mode.


jan-wassenberg (approved):
Nice idea, thank you :) Hopefully the test time is not too long such that our internal CI considers it to have timed out, let's see.

PR #2003, opened 2024-02-28
Fixes to x86 target detection

johnplatts:
The following OS's are known to preserve the lower 128 bits of XMM registers across context switches on x86 CPU's with SSE support, even in 32-bit mode:
- Windows 2000 (released on 2/17/2000) or later
- Linux 2.4.0 (released on 1/4/2001) or later
- FreeBSD 4.4 (released on 9/20/2001) or later
- NetBSD 1.6 (released on  9/14/2002) or later
- OpenBSD 3.5 (released on 5/1/2004) or later
- Mac OS X 10.4.4 (released on 1/10/2006) or later
- Solaris 9 4/04 (released in 2004) or later
- UnixWare 7 Release 7.1.1 (released in 2000) or later

The lower 128 bits of XMM0-XMM15 are guaranteed to be preserved across context switches on x86_64, even on x86_64 CPU's that don't support the XSAVE instruction.

Updated x86 DetectTargets() to only clear the AVX2/AVX3 bits if the CPU or OS doesn't support XSAVE as most of the x86 OS's released in the last 19 years correctly preserve the lower 128 bits of XMM registers across context switches on x86 CPU's that have SSE support, even if the CPU or OS doesn't support the XSAVE instruction.

x86_32 Highway builds that are configured with `-DHWY_CMAKE_SSE2=ON` also assume that the CPU supports SSE2 instructions and the OS correctly preserves the lower 128 bits of XMM registers across context switches, even on CPU's that do not support the XSAVE instruction.

### DIFF
--- a/hwy/targets.cc
+++ b/hwy/targets.cc
@@ -289,32 +289,36 @@ int64_t DetectTargets() {
   }
 #endif
 
-  // Clear bits if the CPU or OS does not support XSAVE - otherwise, registers
-  // are not preserved across context switches.
+  // Clear AVX2/AVX3 bits if the CPU or OS does not support XSAVE - otherwise,
+  // YMM/ZMM registers are not preserved across context switches.
+
+  // The lower 128 bits of XMM0-XMM15 are guaranteed to be preserved across
+  // context switches on x86_64
+
+  // The following OS's are known to preserve the lower 128 bits of XMM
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Thanks for fixing :) Oops, this previously disabled SSE2, sorry about that.

Issue #2011, opened 2024-03-05
bit_pack-inl.h is missing from CMakeLists.txt


niranjan-nilakantan:
bit_pack-inl.h  is missing from the top-level CMakeLists.txt file here.

```
if (HWY_ENABLE_CONTRIB)
# Glob all the traits so we don't need to modify this file when adding
# additional special cases.
file(GLOB HWY_CONTRIB_SOURCES "hwy/contrib/sort/vqsort_*.cc")
list(APPEND HWY_CONTRIB_SOURCES
    hwy/contrib/dot/dot-inl.h
    hwy/contrib/image/image.cc
    hwy/contrib/image/image.h
    hwy/contrib/math/math-inl.h
    hwy/contrib/matvec/matvec-inl.h
        hwy/contrib/random/random-inl.h
    hwy/contrib/sort/order.h
    hwy/contrib/sort/shared-inl.h
    hwy/contrib/sort/sorting_networks-inl.h
    hwy/contrib/sort/traits-inl.h
    hwy/contrib/sort/traits128-inl.h
    hwy/contrib/sort/vqsort-inl.h
    hwy/contrib/sort/vqsort.cc
    hwy/contrib/sort/vqsort.h
    hwy/contrib/thread_pool/futex.h
    hwy/contrib/thread_pool/thread_pool.h
    hwy/contrib/algo/copy-inl.h
    hwy/contrib/algo/find-inl.h
    hwy/contrib/algo/transform-inl.h
    hwy/contrib/unroller/unroller-inl.h
)
endif()  # HWY_ENABLE_CONTRIB
```

As a result, it is skipped from the *cmake install* step.

PR #2012, opened 2024-03-05
Add bit_pack-inl.h to CMakeLists.txt

niranjan-nilakantan:
Add the corresponding test as well.
Test with `run_tests.bat`

Verify with `cmake --install . --prefix=install` in the build_win directory.

Fix warnings as errors for gcc and clang using explicit casts.

#2011


jan-wassenberg (approved):
Thanks for adding and fixing the warnings!

PR #2013, opened 2024-03-06
Added ScalarShr operation

johnplatts:
Added ScalarShr operation, which guarantees that signed right shifts are arithmetic right shifts.

Also updated EMU128/SCALAR Shr, ShiftRight, ShiftRightSame, and BroadcastSignBit ops to use ScalarShr.

GCC also guarantees that `(val >> shift_amt)` is an arithmetic right shift if `val` is a signed integer according to https://gcc.gnu.org/onlinedocs/gcc-4.8.0/gcc/Integers-implementation.html.

MSVC also guarantees that `(val >> shift_amt)` is an arithmetic right shift if `val` is a signed integer according to https://learn.microsoft.com/en-us/cpp/cpp/left-shift-and-right-shift-operators-input-and-output?view=msvc-140.

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -2434,6 +2434,46 @@ constexpr inline size_t RoundDownTo(size_t what, size_t align) {
   return what - (what % align);
 }
 
+namespace detail {
+
+// T is unsigned or T is signed and (val >> shift_amt) is an arithmetic right
+// shift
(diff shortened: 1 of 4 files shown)


jan-wassenberg (approved):
Nice simplification 👍

PR #2017, opened 2024-03-07
Fixes to RVV CappedLanes

johnplatts:
Section 6.3 of the RVV specification, which can be found at https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc, allows the __riscv_vsetvl_e* intrinsics to return a value of less than `cap` but at least `(cap + 1) / 2` if all of the following are true:
- `cap >= 5`
- `(cap & (cap - 1)) != 0` (in other words, `cap` is not 0 or a power of 2)
- `cap > Lanes(ScalableTag<TFromD<decltype(d)>, d.Pow2()>())`
- `cap < 2 * Lanes(ScalableTag<TFromD<decltype(d)>, d.Pow2()>())`

We want `CappedLanes(d, cap)` to return the same result as `HWY_MIN(d, Lanes(d))` as the as `LoadN(d, p, cap)` operation expects to load exactly `HWY_MIN(cap, Lanes(d))` lanes and as the `StoreN(v, d, p, cap)` operation expects to store exactly `HWY_MIN(cap, Lanes(d))` lanes, even on RVV implementations where vsetvl returns a result that is less than `Lanes(d)` but greater than or equal to `(cap + 1) / 2` in the case where `cap >= 5 && cap > Lanes(d) && cap < 2 * Lanes(d)`.

This pull request fixes CappedLanes to ensure that `Lanes(d)` is always returned in the case where `cap > Lanes(d)`, even on RVV implementations whose vsetvl instruction returns a value less than `Lanes(d)` in the case where `cap >= 5 && cap > Lanes(d) && cap < 2 * Lanes(d)`.

### DIFF
--- a/hwy/ops/rvv-inl.h
+++ b/hwy/ops/rvv-inl.h
@@ -438,7 +438,9 @@ HWY_RVV_FOREACH(HWY_SPECIALIZE, _, _, _ALL)
 // WARNING: we want to query VLMAX/sizeof(T), but this may actually change VL!
 
 #if HWY_COMPILER_GCC && !HWY_IS_DEBUG_BUILD
-#define HWY_RVV_CAPPED_LANES_OPTO(BASE, SEW, LMUL)                             \
+// HWY_RVV_CAPPED_LANES_SPECIAL_CASES provides some additional optimizations
+// to CappedLanes in non-debug builds
+#define HWY_RVV_CAPPED_LANES_SPECIAL_CASES(BASE, SEW, LMUL)                    \
   if (__builtin_constant_p(cap >= kMaxLanes) && (cap >= kMaxLanes)) {          \
     /* If cap is known to be greater than or equal to MaxLanes(d), */          \
     /* HWY_MIN(cap, Lanes(d)) will be equal to Lanes(d) */                     \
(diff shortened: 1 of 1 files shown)


hwy/ops/rvv-inl.h
```diff
@@ -436,22 +436,76 @@ HWY_RVV_FOREACH(HWY_SPECIALIZE, _, _, _ALL)
...
-  }                                                                           \
-  template <size_t N>                                                         \
-  HWY_API size_t Capped##NAME(HWY_RVV_D(BASE, SEW, N, SHIFT) d, size_t cap) { \
-    /* If no cap, avoid the HWY_MIN. */                                       \
-    return detail::IsFull(d)                                                  \
-               ? __riscv_vsetvl_e##SEW##LMUL(cap)                             \
-               : __riscv_vsetvl_e##SEW##LMUL(HWY_MIN(cap, MaxLanes(d)));      \
+
+#if HWY_COMPILER_GCC && !HWY_IS_DEBUG_BUILD
+#define HWY_RVV_CAPPED_LANES_OPTO(BASE, SEW, LMUL)                             \
```
jan-wassenberg, line 441: Can you help me understand the OPTO naming?

hwy/ops/rvv-inl.h
```diff
@@ -436,22 +436,76 @@ HWY_RVV_FOREACH(HWY_SPECIALIZE, _, _, _ALL)
...
+    /* If cap is known to be greater than or equal to MaxLanes(d), */          \
+    /* HWY_MIN(cap, Lanes(d)) will be equal to Lanes(d) */                     \
+    return Lanes(d);                                                           \
+  }                                                                            \
+                                                                               \
+  if ((__builtin_constant_p((cap & (cap - 1)) == 0) &&                         \
+       ((cap & (cap - 1)) == 0)) ||                                            \
+      (__builtin_constant_p(cap <= HWY_MAX(kMinLanesPerFullVec, 4)) &&         \
+       (cap <= HWY_MAX(kMinLanesPerFullVec, 4)))) {                            \
+    /* If cap is known to be a power of 2 or cap is known to be less than */   \
```
jan-wassenberg, line 452: Let's expand on this comment a bit. VLEN is a pow2. VLMAX is that times a pow2 so also a pow2.
So if cap is a pow2, then either vl=cap if cap <= VLMAX, or vl = VLMAX if cap > VLMAX due to rule 3, and in both cases vl is a pow2. If cap > VLMAX, then it is a loose upper bound and HWY_MIN has no effect. Otherwise, the vsetvl anyway returns cap and HWY_MIN with that is again cap.

What about the kMinLanesPerFullVec path?

hwy/ops/rvv-inl.h
```diff
@@ -1816,9 +1876,9 @@ HWY_API void StoreN(size_t count, VFromD<D> v, D d, TFromD<D>* HWY_RESTRICT p) {
 template <class D>
 HWY_API void StoreN(VFromD<D> v, D d, TFromD<D>* HWY_RESTRICT p,
```
jan-wassenberg, line 1877: Do we also want to apply capping to LoadN?

jan-wassenberg (commented):
Thanks for looking into this. Some questions:

hwy/ops/rvv-inl.h
johnplatts: `kMinLanesPerFullVec` is the smallest possible value of VLMAX of a vector of type `VFromD<decltype(d)>` (or the smallest value that would be returned by `Lanes(DFromV<VFromD<decltype(d)>>())`).

The HWY_RVV target currently requires support for both the RVV Zvl128b (VLEN of at least 128 bits) and Zve64d (support for F32/F64/I64/U64 vectors) extensions. The RISC-V Vector Extension for Application Processors described in Section 18.3 of the RVV specification also requires support for both the Zvl128b and Zve64d extensions.

If `cap <= kMinLanesPerFullVec` is true, then vsetvl is guaranteed to return `cap` as VLMAX is at least `kMinLanesPerFullVec`.

If `cap <= 4` is true, then vsetvl is guaranteed to return `(cap <= VLMAX) ? cap : VLMAX` as VLMAX is at least 2 on RISC-V CPUs that support the Zvl128b extension and as vsetvl is guaranteed to return 2 in the case where `(cap == 3 || cap == 4) && VLMAX == 2` under the requirements of Section 6.3 of the RISC-V specification since 4 is a power of 2 and `(3 + 1) / 2` is equal to 2.

hwy/ops/rvv-inl.h
johnplatts: The capping is already applied to the LoadN operation through `CappedLanes(d, num_lanes)` in the LoadN implementation.

hwy/ops/rvv-inl.h
johnplatts: The HWY_RVV_CAPPED_LANES_OPTO macro is there to implement some optimizations to CappedLanes in non-debug builds for the cases where (a) `cap` is known to be at least `MaxLanes(d)`, (b) `cap` is known to be either zero or a power of 2, and (c) `cap` is known to be less than or equal to `HWY_MAX(kMinLanesPerFullVec, 4)` (as described below).

jan-wassenberg: Ah, I had not seen that usage of 'opto' before. I suppose OPT could also be understood as option(al). Maybe FAST or _SPECIAL_CASES would be more clear?

hwy/ops/rvv-inl.h
jan-wassenberg: Ah, now I understand where the 5 comes from, thanks. It would be very helpful to have some of this explanation as comments in the code, because the commit message/discussion are less likely to be looked at. Would you like to expand the code comments a bit? Especially the "at least two" * 2 bit was not clear to me.

hwy/ops/rvv-inl.h
johnplatts: I have renamed HWY_RVV_CAPPED_LANES_OPTO to HWY_RVV_CAPPED_LANES_SPECIAL_CASES.

jan-wassenberg (approved):
Nice, thank you for expanding the comments :)

Issue #2018, opened 2024-03-11
Does the Highway have partial sort functionality?


enum-class:
If not, are there any plans to add it in the future ?

jan-wassenberg:
I assume for TopK? We do not have it, but it sounds useful.
Would you like to add a variant of Quicksort that just skips the right partition (if it is beyond K)?

enum-class:
Yes for TopK, I will do it

jan-wassenberg:
Thanks @enum-class for adding this :)

PR #2019, opened 2024-03-12
Add Partial Sort

enum-class:
Add Partial Sort https://github.com/google/highway/issues/2018
Fix minor issues in select algorithm #2024


hwy/contrib/sort/sort_test.cc
```diff
@@ -618,6 +618,40 @@ std::vector<Algo> AlgoForTest() {
...
+
+  std::partial_sort(s.begin(), s.begin() + RANGE, s.end(),
+                    std::greater<KeyType>{});
+  PartialSort(d, st, aligned.get(), num_keys, RANGE);
+
+  // Select(d, st, aligned.get(), num_keys, buf, RANGE);
+  // std::sort(aligned.get(), aligned.get() + RANGE, std::greater<KeyType>{});
+
+  for (size_t i = 0; i < RANGE; ++i)
+    HWY_ASSERT(std::fabs(aligned.get()[i] - s[i]) < 0.01);
```
jan-wassenberg, line 652: Nit: 0.01f is more efficient (avoids promoting the expression to double).

0.01 is also pretty large epsilon because the random inputs are in [1, 2), perhaps 1E-4f?

Finally, vqsort supports f16. If you want to support that in this test as well, we can `ConvertScalarTo<float>` all of the sorted values before subtracting/comparing.

jan-wassenberg:
Wow, that was very quick :) There was some great work in #1710 that added VQSelect and pretty thorough testing. Unfortunately there was a merge failure and we'd just need to rebase and then we could build on that for your partial sort.
Would you like to patch that PR into this PR?

enum-class:
@jan-wassenberg This is just a draft and still needs more refinement and testing. :)
#1710 looks perfect, perhaps we could kindly ask @zbjornson to update the pull request.
Then partial sort would be an easy task to build upon

jan-wassenberg:
I believe this is superseded by #2024, closing but please feel free to reopen if not :)

jan-wassenberg:
Oops, that PR does not yet contain the PartialSort defined here. Reopening :)

jan-wassenberg (approved):
Nice, great that you've added this 👍

PR #2024, opened 2024-03-14
Add VQSelect and HeapSelect

enum-class:
Implemented VQSelect by @zbjornson in #1710.
I've incorporated his patch with some refinements.
Also, added HeapSelect functionality.


hwy/contrib/sort/sort_test.cc
```diff
@@ -732,6 +733,96 @@ void TestAllSort() {
   }
 }
 
+std::vector<Algo> SelectAlgoForTest() {
+  return {
+#if VQSORT_ENABLED
+//    Algo::kVQSelect,
```
jan-wassenberg, line 739: Should we uncomment this?

hwy/contrib/sort/vqsort-inl.h
```diff
@@ -42,6 +42,12 @@
 #define VQSORT_PRINT 0
 #endif
 
+#if HWY_CXX_LANG >= 201703L
```
jan-wassenberg, line 45: Let's move this to base.h, for example near HWY_HAVE_CXX20_THREE_WAY_COMPARE.

jan-wassenberg (changes requested):
Super, thanks for updating and improving this patch! Two minor suggestions, then we're good to merge :) (The MSVC fix is in.)

jan-wassenberg (approved):
Nice, thanks for making the changes.

enum-class:
@jan-wassenberg Thanks for your comments, I have fixed issues.

PR #2026, opened 2024-03-14
Added IsNegative, IfNegativeThenElseZero, and IfNegativeThenZeroElse ops

johnplatts:
Added IsNegative op as `Lt(v, Zero(d))` is more efficient than `MaskFromVec(BroadcastSignBit(v))` on some targets whereas `MaskFromVec(BroadcastSignBit(v))` is more efficient than `Lt(v, Zero(d))` on some other targets.

Also added IfNegativeThenElseZero op as `IfThenElseZero(IsNegative(v), yes)` is more efficient than `IfNegativeThenElse(v, yes, Zero(d))` on some targets whereas `IfNegativeThenElse(v, yes, Zero(d))` is more efficient than `IfThenElseZero(IsNegative(v), yes)` on some other targets.

Also added IfNegativeThenZeroElse op as `IfThenZeroElse(IsNegative(v), no)` is more efficient than `IfNegativeThenElse(v, Zero(d), no)` on some targets whereas `IfNegativeThenElse(v, Zero(d), no)` is more efficient than `IfThenZeroElse(IsNegative(v), no)` on some other targets.

Also moved implementation of ZeroIfNegative for all targets to hwy/ops/generic_ops-inl.h.

Also updated TestZeroIfNegative to handle int8_t vectors and vectors that have more than 8192 lanes.

### DIFF
--- a/hwy/ops/x86_128-inl.h
+++ b/hwy/ops/x86_128-inl.h
@@ -870,6 +870,19 @@ HWY_API MFromD<D> MaskFalse(D /*d*/) {
   return MFromD<D>{static_cast<decltype(MFromD<D>().raw)>(0)};
 }
 
+// ------------------------------ IsNegative (MFromD)
+#ifdef HWY_NATIVE_IS_NEGATIVE
+#undef HWY_NATIVE_IS_NEGATIVE
+#else
+#define HWY_NATIVE_IS_NEGATIVE
+#endif
+
+// Generic for all vector lengths
+template <class V, HWY_IF_NOT_UNSIGNED_V(V)>
+HWY_API MFromD<DFromV<V>> IsNegative(V v) {
+  return MaskFromVec(v);
+}
+
 // ------------------------------ PromoteMaskTo (MFromD)
 
 #ifdef HWY_NATIVE_PROMOTE_MASK_TO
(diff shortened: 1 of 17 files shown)


hwy/ops/x86_128-inl.h
```diff
@@ -870,6 +870,19 @@ HWY_API MFromD<D> MaskFalse(D /*d*/) {
...
+#ifdef HWY_NATIVE_IS_NEGATIVE
+#undef HWY_NATIVE_IS_NEGATIVE
+#else
+#define HWY_NATIVE_IS_NEGATIVE
+#endif
+
+// Generic for all vector lengths
+template <class V, HWY_IF_NOT_UNSIGNED_V(V)>
+HWY_API MFromD<DFromV<V>> IsNegative(V v) {
+  return MaskFromVec(v);
```
jan-wassenberg, line 883: I was concerned about this one because it's not obvious from the diff that we are in the AVX3 section, where this is safe. I've verified that we are, so all good. But consider adding a comment that this works (only) for AVX3+, whereas on AVX2 we have to return 0xFF..FF in masks?

jan-wassenberg (approved):
Very nice as usual, looks useful. I always thought ZeroIfNegative was too specific. Thanks for adding, and removing the extra ZeroIfNegative calls for u64.

PR #2027, opened 2024-03-14
Added IsEitherNaN op

johnplatts:
Added IsEitherNaN op as SSE2/SSSE3/SSE4/AVX2/AVX3 can carry out `Or(IsNaN(a), IsNaN(b))` using a single VCMPPH/(V)CMPPS/(V)CMPPD instruction.


jan-wassenberg (approved):
Nice optimization, thanks for adding :)

PR #2028, opened 2024-03-14
Replaced remaining usages of isnan with ScalarIsNaN in scalar-inl.h

johnplatts:
Replaced remaining usages of isnan with ScalarIsNaN in scalar-inl.h.


jan-wassenberg (approved):
👍

PR #2030, opened 2024-03-18
Mention in `README.md` that sometimes MSVC would generate VEX-encoded SSE, but not always

### LINKED ISSUE #2029: Question: VEX-encoded SSE4 mentioned in `README.md`
It says:

> If intending to use the AVX2 target together with half-width vectors (e.g. for `PromoteTo`), it is also important to compile with `/arch:AVX2`. This seems to be the only way to generate VEX-encoded SSE4 instructions on MSVC. Otherwise, mixing VEX-encoded AVX2 instructions and non-VEX SSE4 may cause severe performance degradation. Unfortunately, the resulting binary will then require AVX2. 

I haven't encountered a case where MSVC does not emit VEX-encoded SSE what it is followed by AVX (unless compiling with optimizations disabled).

My experiments has shown that it applies to SSE4 either, though i didn't test every instruction https://godbolt.org/z/9PP4onW34

Maybe you have a specific instruction combination in mind (which should be reported as a bug probably), or this information is obsolete?

AlexGuteniev:
Resolves #2029

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,7 +322,7 @@ For MSVC, we recommend compiling with `/Gv` to allow non-inlined functions to
 pass vector arguments in registers. If intending to use the AVX2 target together
 with half-width vectors (e.g. for `PromoteTo`), it is also important to compile
 with `/arch:AVX2`. This seems to be the only way to reliably generate VEX-encoded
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Thanks for expanding this and adding the link :)

PR #2035, opened 2024-03-21
Made enhancements to unsigned to signed DemoteTo/ReorderDemote2To

johnplatts:
Unsigned to signed DemoteTo/ReorderDemote2To on EMU128/SCALAR/SVE/SVE2 now use target-specific implementations

Also updated U64->I8/I16/I32 DemoteTo/ReorderDemote2To to use target-specific implementations on SSE2/SSSE3/SSE4/AVX2


jan-wassenberg (approved):
👍

PR #2036, opened 2024-03-22
Fix for compilation error in matvec-inl.h with GCC 8 in C++17 mode

johnplatts:
Added `(void)add` to MatVecAddImpl to work around a compilation warning with GCC 8 in C++17 mode if `kAdd` is false.


jan-wassenberg (approved):
Thanks for fixing 👍

PR #2042, opened 2024-03-28
Added float to int PromoteInRangeTo/ConvertInRangeTo/DemoteInRangeTo ops

johnplatts:
The FastPromoteTo, FastConvertTo, FastDemoteTo, FastPromoteLowerTo, FastPromoteUpperTo, FastPromoteEvenTo, and FastPromoteOddTo ops were added to allow for more efficient conversions from floating point to integer vectors in the case where `v[i]` is known to be within the range of `TFromD<D>`.

The FastPromoteTo, FastConvertTo, and FastDemoteTo ops return an implementation-defined result in the case where `v[i]` is outside of the range of `TFromD<D>`.

The implementations of FastPromoteTo, FastConvertTo, and FastDemoteTo in this pull request also will convert values that are greater than or equal to `LimitsMin<TFromD<D>>() * 2` and less than `LimitsMax<TFromD<D>>() * 2 + 1` by either saturation, truncation, or wrapping around to `LimitsMin<TFromD<D>>()`, which allows FastPromoteTo or FastConvertTo to be used to implement `detail::IntDivUsingFloatDiv<kOrigLaneSize>(V a, V b)` for the case where `sizeof(TFromV<V>) >= kOrigLaneSize`.

The FastPromoteTo, FastConvertTo, and FastDemoteTo ops also allow the x86 implementations of float to signed integer PromoteTo, ConvertTo, and DemoteTo implementations to be simplified as the float to signed integer FastPromoteTo, FastConvertTo, and FastDemoteTo implementations on x86 are wrappers around SSE2/AVX2/AVX3/AVX3_DL intrinsics if supported by the target.


hwy/ops/arm_neon-inl.h
```diff
@@ -4489,6 +4489,31 @@ HWY_API VFromD<D> PromoteTo(D du64, VFromD<Rebind<float, D>> v) {
             lo32_or_mask);
 }
 
+#ifdef HWY_NATIVE_F32_TO_UI64_FAST_PROMOTE_TO
```
jan-wassenberg, line 4492: I'm curious why this one is native, and not in generic_ops?

jan-wassenberg (commented):
Thanks for adding this. I particularly like the f32->ui64 implementation and the extensive comments 👍 
I'm wondering about the naming. Thinking aloud:
- It's not clear to users what exactly Fast means, or when they should NOT use it?
- Native (just do what the platform gives us) doesn't say anything about how the platforms differ.
- a short name helps because we also have XPromoteUpperTo etc.
- how should this interact with "demote f64 < 2^52 to int", in case we later add an op for that?
- maybe we could call this one PromoteInRangeTo? Or something with Bounded?

johnplatts:
> * maybe we could call this one PromoteInRangeTo? Or something with Bounded?

Should I go ahead and rename FastPromoteTo to PromoteInRangeTo, FastDemoteTo to DemoteInRangeTo, and FastConvertTo to ConvertInRangeTo?

jan-wassenberg:
> > * maybe we could call this one PromoteInRangeTo? Or something with Bounded?
> 
> Should I go ahead and rename FastPromoteTo to PromoteInRangeTo, FastDemoteTo to DemoteInRangeTo, and FastConvertTo to ConvertInRangeTo?

Sounds good to me, please do :)

jan-wassenberg (approved):
Thanks for renaming :)

PR #2043, opened 2024-03-28
Fix compiler warning with Load/Store on PPC with GCC

johnplatts:
Updated the 16-byte vector Load/Store ops on PPC to suppress the ignoring attributes warning that is generated by the `HWY_RCAST_ALIGNED(const LoadRaw*, aligned)` and `HWY_RCAST_ALIGNED(StoreRaw*, aligned)` expressions with GCC.


jan-wassenberg (approved):
👍

PR #2063, opened 2024-04-08
Use __ARM_FEATURE_FMA to detect FMA support

Mousius:
Currently `__ARM_VFPV4__` is used, which I can't see defined in GCC, though `__ARM_FEATURE_FMA` is available: https://github.com/gcc-mirror/gcc/blob/97d5cd8740384dbce5a83080916388f80d8976dd/gcc/config/arm/arm-c.cc#L361C31-L361C48 And is targetting the same as `__ARM_VFPV4__`:
https://github.com/gcc-mirror/gcc/blob/97d5cd8740384dbce5a83080916388f80d8976dd/gcc/config/arm/arm.h#L211

### DIFF
--- a/hwy/ops/arm_neon-inl.h
+++ b/hwy/ops/arm_neon-inl.h
@@ -2579,7 +2579,7 @@ HWY_API Vec128<T, N> NegMulAdd(Vec128<T, N> mul, Vec128<T, N> x,
 
 namespace detail {
 
-#if defined(__ARM_FEATURE_FMA) || defined(__ARM_VFPV4__) || HWY_ARCH_ARM_A64
+#if HWY_NATIVE_FMA
 // Wrappers for changing argument order to what intrinsics expect.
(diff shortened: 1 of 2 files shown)


Mousius:
@jan-wassenberg I wasn't sure, but I think this should be wrapped up in `HWY_NATIVE_FMA` from `set_macros-inl.h`? 

I'm also aware there's some logic here in `detect_targets.h`, which looks at all these flags, but I'm unsure of the history.

jan-wassenberg (commented):
Thanks for updating this. Good point about the set_macros. I think it would make sense to update the definition of NATIVE_FMA, to ensure that also benefits from this change, then we could change the occurrence in arm_neon to use HWY_NATIVE_FMA instead, what do you think?

Mousius:
@jan-wassenberg agree and updated 😸 I'm also running a local NumPy build to double-check it works as intended.

Mousius:
Local NumPy check showed this worked as intended 😸

PR #2067, opened 2024-04-09
Made fix to F32 to F16 DemoteTo in generic_ops-inl.h

johnplatts:
Fixed the F32 to F16 DemoteTo implementation in generic_ops-inl.h to prevent out-of-range finite values from being incorrectly converted to NaN values.

Also updated the ForDemoteVectors test to check that the F32->F16, F64->F16, and F64->F32 DemoteTo ops correctly convert out-of-range finite values to positive infinity or negative infinity.

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -1394,8 +1394,22 @@ HWY_API HWY_F16_CONSTEXPR float16_t F16FromF32(float f32) {
   // 1[01] + 10 =  1[11]
   // 1[10] + 10 = C0[00]  (round up toward even with C=1 carry out)
   // 1[11] + 10 = C0[01]  (round up toward even with C=1 carry out)
-  const uint32_t odd_bit = (mantissa32 >> 13) & 1;
-  const uint32_t rounded = mantissa32 + odd_bit + 0xFFF;
+
+  // If |f32| >= 2^-24, f16_ulp_bit_idx is the index of the F32 mantissa bit
+  // that will be shifted down into the ULP bit of the rounded down F16 result
+
+  // The biased F32 exponent of 2^-14 (the smallest positive normal F16 value)
+  // is 113, and bit 13 of the F32 mantissa will be shifted down to into the ULP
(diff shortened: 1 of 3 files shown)


jan-wassenberg (dismissed):
👍

jan-wassenberg:
We are seeing CI failures:
```
[ RUN      ] HwyDemoteTestGroup/HwyDemoteTest.TestAllDemoteToFloat/SSE4
f16x2 expect [0+ ->]:
  3.5167E-06,0.0000E+00,
f16x2 actual [0+ ->]:
  3.5763E-06,0.0000E+00,
Abort at demote_test.cc:182: SSE4, f16x2 lane 0 mismatch: expected '3.5167E-06', got '3.5763E-06'.
```
Unfortunately the input is random so hard to tell which one is correct. But it does look like rounding is involved, because
the mantissas are [1.]111 and [1.]11011.

johnplatts:
> We are seeing CI failures:
> 
> ```
> [ RUN      ] HwyDemoteTestGroup/HwyDemoteTest.TestAllDemoteToFloat/SSE4
> f16x2 expect [0+ ->]:
>   3.5167E-06,0.0000E+00,
> f16x2 actual [0+ ->]:
>   3.5763E-06,0.0000E+00,
> Abort at demote_test.cc:182: SSE4, f16x2 lane 0 mismatch: expected '3.5167E-06', got '3.5763E-06'.
> ```
> 
> Unfortunately the input is random so hard to tell which one is correct. But it does look like rounding is involved, because the mantissas are [1.]111 and [1.]11011.

I have found a bug in F16FromF32 that is causing the conversion of F32 values that are within the range of denormal F16 values to be incorrectly rounded.

jan-wassenberg (approved):
Thanks for fixing :)

johnplatts:
What is the status of merging this pull request?

jan-wassenberg:
Thanks for checking. It's ready to go and I have followed up internally.

Issue #2062, opened 2024-04-05
ReorderWidenMulAccumulate with guaranteed lanes order and sum


awxkee:
It would be very nice when `ReorderWidenMulAccumulate` or similar api that will do always guaranteed widen `a[i] * b[i] + c[i]` because it is base operation when you're trying to apply transformation matrix in integer space. For ex. whole `libyuv` done using `vmlal_u16` or it's similars on other than NEON platforms. At the moment it is necessary to check ReorderWidenMulAccumulate on platforms I know and write tons of if like this one `#if (HWY_ARCH_ARM_A64 && (HWY_TARGET == HWY_NEON`. Of course it is possible to write generic backed by MulAdd but benefits will be lost.

jan-wassenberg:
Hi, thanks for reaching out. Does WidenMulPairwiseAdd work for your use case? That's just missing the accumulate part which is an extra add.
If not, any thoughts on how to bridge the difference between SVE and NEON, where the former handles odd/even lanes, and the latter upper/lower halves? This is why we relaxed the interface to allow reordering.

awxkee:
Hi. No WidenMulPairwiseAdd doesn't do that required. I'm using ReorderWidenMulAccumulate because it is based ( in NEON version ) on vmlal where [vmlal](https://developer.arm.com/architectures/instruction-sets/intrinsics/vmlal_u16) function wide lower/upper part and multiplies it on corresponding vectors. There is not deal with even/odd lanes so I think it will be easy to implement generic on platforms where it is not supported.

I don't tested this, however idea is simple.
```c++
template<typename D, HWY_IF_U16_D(D), HWY_IF_LANES_D(D, 8), typename VW = Vec<RepartitionToWide<D>>>
void WidenMul(D d, Vec<D> v, VW hi, VW lo, VW &hiSum, VW& loSum) {
  const VW dw;
  loSum = MulAdd(PromoteLowerTo(dw, v), lo, loSum);
  hiSum = MulAdd(PromoteUpperTo(dw, v), hi, hiSum);
}
```

Especially for your question about deal with even/odd lanes it can be achieved by bit masking, however it is not needed in this case I believe.

```c++
template<typename D, HWY_IF_I16_D(D), HWY_IF_LANES_D(D, 8), typename VH = Vec<Half<D>>>
VH GetOdd(D d, Vec<D> v) {
  const Half<D> dh;
  const Rebind<int32_t, decltype(dh)> du32;
[...]

PR #2071, opened 2024-04-09
Added `WidenMulAccumulate`

awxkee:
Allows to use wide mul and add with predictable lanes as per #2062

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -771,6 +771,11 @@ All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
     `sum[i] + a[4*i+3]*b[4*i+3] + a[4*i+2]*b[4*i+2] + a[4*i+1]*b[4*i+1] +
     a[4*i+0]*b[4*i+0]`
 
+*   `VW`: `{u,i}{16,32,64}`: `V`: `{u,i}{8,16,32}`, `D`, 
+    `VW`: `Vec<RepartitionToWide<DFromV<V>>`:\
+    `VW WidenMulAccumulate(D, V, V, VW low, VW& high)`: widens `a` and `b`, \
+    then mul lower and higher and adds low and high `a[i]*b[i] + c[i]`
+
 #### Fused multiply-add
 
 When implemented using special instructions, these functions are more precise
--- a/hwy/ops/arm_neon-inl.h
+++ b/hwy/ops/arm_neon-inl.h
@@ -6486,6 +6486,320 @@ HWY_API VFromD<D> SlideDownLanes(D d, VFromD<D> v, size_t amt) {
   return detail::SlideDownLanes(v, amt);
 }
 
+// ------------------------------- WidenHighMulAdd
+
+#ifdef HWY_NATIVE_WIDEN_HIGH_MUL_ADD
+#undef HWY_NATIVE_WIDEN_HIGH_MUL_ADD
+#else
+#define HWY_NATIVE_WIDEN_HIGH_MUL_ADD
+#endif
+
+template<class D, HWY_IF_U32_D(D)>
+HWY_API Vec128<uint64_t> WidenHighMulAdd(D /* tag */, Vec128<uint32_t> mul,
+                                         Vec128<uint32_t> x, Vec128<uint64_t> add) {
+#if HWY_ARCH_ARM_A64
+  return Vec128<uint64_t>(vmlal_high_u32(add.raw, mul.raw, x.raw));
+#else
+  const Full64<uint32_t> dh;
+  return Vec128<uint64_t>(
+      vmlal_u32(add.raw, UpperHalf(dh, mul).raw, UpperHalf(dh, x).raw));
+#endif
+}
+
+template<class D, HWY_IF_U32_D(D)>
+HWY_API Vec64<uint64_t> WidenHighMulAdd(D /* tag */, Vec64<uint32_t> mul,
+                                        Vec64<uint32_t> x, Vec64<uint64_t> add) {
+  Vec128<uint64_t> mulResult = Vec128<uint64_t>(vmull_u32(mul.raw, x.raw));
+  const RepartitionToWide<D> d64;
+  return UpperHalf(d64, mulResult) + add;
+}
+
+template<class D, HWY_IF_I32_D(D)>
+HWY_API Vec128<int64_t> WidenHighMulAdd(D /* tag */, Vec128<int32_t> mul,
+                                        Vec128<int32_t> x, Vec128<int64_t> add) {
+#if HWY_ARCH_ARM_A64
+  return Vec128<int64_t>(vmlal_high_s32(add.raw, mul.raw, x.raw));
+#else
+  const Full64<int32_t> dh;
+  return Vec128<int64_t>(
+      vmlal_s32(add.raw, UpperHalf(dh, mul).raw, UpperHalf(dh, x).raw));
+#endif
+}
+
+template<class D, HWY_IF_I32_D(D)>
+HWY_API Vec64<int64_t> WidenHighMulAdd(D /* tag */, Vec64<int32_t> mul,
+                                       Vec64<int32_t> x, Vec64<int64_t> add) {
+  Vec128<int64_t> mulResult = Vec128<int64_t>(vmull_s32(mul.raw, x.raw));
+  const RepartitionToWide<D> d64;
+  return UpperHalf(d64, mulResult) + add;
+}
+
+template<class D, HWY_IF_I16_D(D)>
+HWY_API Vec128<int32_t> WidenHighMulAdd(D /* tag */, Vec128<int16_t> mul,
+                                        Vec128<int16_t> x, Vec128<int32_t> add) {
+#if HWY_ARCH_ARM_A64
+  return Vec128<int32_t>(vmlal_high_s16(add.raw, mul.raw, x.raw));
+#else
+  const Full64<int16_t> dh;
+  return Vec128<int32_t>(
+      vmlal_s16(add.raw, UpperHalf(dh, mul).raw, UpperHalf(dh, x).raw));
+#endif
+}
+
+template<class D, HWY_IF_I16_D(D)>
+HWY_API Vec64<int32_t> WidenHighMulAdd(D d, Vec64<int16_t> mul,
+                                       Vec64<int16_t> x, Vec64<int32_t> add) {
+  Vec128<int32_t> widen = Vec128<int32_t>(vmull_s16(mul.raw, x.raw));
+  const RepartitionToWide<decltype(d)> d32;
+  Vec64<int32_t> hi = UpperHalf(d32, widen);
+  return hi + add;
+}
+
+template<class D, HWY_IF_I16_D(D)>
+HWY_API Vec32<int32_t> WidenHighMulAdd(D d, Vec32<int16_t> mul,
+                                       Vec32<int16_t> x, Vec32<int32_t> add) {
+  Vec128<int32_t> widen = Vec128<int32_t>(vmull_s16(mul.raw, x.raw));
+  const RepartitionToWide<decltype(d)> d32;
+  Vec32<int32_t> hi = UpperHalf(d32, Vec64<int32_t>(vget_high_s32(widen.raw)));
+  return hi + add;
+}
+
+template<class D, HWY_IF_U16_D(D)>
+HWY_API Vec128<uint32_t> WidenHighMulAdd(D /* tag */, Vec128<uint16_t> mul,
+                                         Vec128<uint16_t> x, Vec128<uint32_t> add) {
+#if HWY_ARCH_ARM_A64
+  return Vec128<uint32_t>(vmlal_high_u16(add.raw, mul.raw, x.raw));
+#else
+  const Full64<uint16_t> dh;
+  return Vec128<uint32_t>(
+      vmlal_u16(add.raw, UpperHalf(dh, mul).raw, UpperHalf(dh, x).raw));
+#endif
+}
+
+template<class D, HWY_IF_U16_D(D)>
+HWY_API Vec64<uint32_t> WidenHighMulAdd(D d, Vec64<uint16_t> mul,
+                                        Vec64<uint16_t> x, Vec64<uint32_t> add) {
+  Vec128<uint32_t> widen = Vec128<uint32_t>(vmull_u16(mul.raw, x.raw));
+  const RepartitionToWide<decltype(d)> d32;
+  Vec64<uint32_t> hi = UpperHalf(d32, widen);
+  return hi + add;
+}
+
+template<class D, HWY_IF_U16_D(D)>
+HWY_API Vec32<uint32_t> WidenHighMulAdd(D d, Vec32<uint16_t> mul,
+                                        Vec32<uint16_t> x, Vec32<uint32_t> add) {
+  Vec128<uint32_t> widen = Vec128<uint32_t>(vmull_u16(mul.raw, x.raw));
+  const RepartitionToWide<decltype(d)> d32;
+  Vec32<uint32_t> hi = UpperHalf(d32, Vec64<uint32_t>(vget_high_s32(widen.raw)));
+  return hi + add;
+}
+
+template<class D, HWY_IF_U8_D(D)>
+HWY_API Vec128<uint16_t> WidenHighMulAdd(D /* tag */, Vec128<uint8_t> mul,
+                                         Vec128<uint8_t> x, Vec128<uint16_t> add) {
+#if HWY_ARCH_ARM_A64
+  return Vec128<uint16_t>(vmlal_high_u8(add.raw, mul.raw, x.raw));
+#else
+  const Full64<uint8_t> dh;
+  return Vec128<uint16_t>(
+      vmlal_u8(add.raw, UpperHalf(dh, mul).raw, UpperHalf(dh, x).raw));
+#endif
+}
+
+template<class D, HWY_IF_U8_D(D)>
+HWY_API Vec64<uint16_t> WidenHighMulAdd(D d, Vec64<uint8_t> mul,
+                                        Vec64<uint8_t> x, Vec64<uint16_t> add) {
+  Vec128<uint16_t> widen = Vec128<uint16_t>(vmull_u8(mul.raw, x.raw));
+  const RepartitionToWide<decltype(d)> d16;
+  Vec64<uint16_t> hi = UpperHalf(d16, widen);
+  return hi + add;
+}
+
+template<class D, HWY_IF_U8_D(D)>
+HWY_API Vec32<uint16_t> WidenHighMulAdd(D d, Vec32<uint8_t> mul,
+                                        Vec32<uint8_t> x, Vec32<uint16_t> add) {
+  Vec128<uint16_t> widen = Vec128<uint16_t>(vmull_u8(mul.raw, x.raw));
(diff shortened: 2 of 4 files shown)


g3doc/quick_reference.md
```diff
@@ -786,6 +786,11 @@ All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
     `sum[i] + a[4*i+3]*b[4*i+3] + a[4*i+2]*b[4*i+2] + a[4*i+1]*b[4*i+1] +
     a[4*i+0]*b[4*i+0]`
 
+*   `VW`: `{u,i}{16,32,64}`: `V`: `{u,i}{8,16,32}`, `D`,
```
jan-wassenberg, line 789: Let's change this to just mention the constraints on V, because D and VW are derived from V
`V`: `{u,i}{8,16,32}`, `D = RepartitionToWide<DFromV<V>`, `VW = Vec<D>`.

g3doc/quick_reference.md
```diff
@@ -786,6 +786,11 @@ All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
     `sum[i] + a[4*i+3]*b[4*i+3] + a[4*i+2]*b[4*i+2] + a[4*i+1]*b[4*i+1] +
     a[4*i+0]*b[4*i+0]`
 
+*   `VW`: `{u,i}{16,32,64}`: `V`: `{u,i}{8,16,32}`, `D`, 
+    `VW`: `Vec<RepartitionToWide<DFromV<V>>`:\
+    `VW WidenMulAccumulate(D, V, V, VW low, VW& high)`: widens `a` and `b`, \
```
jan-wassenberg, line 791: Let's add a and b to the parameters. How about something like this for the documentation?

widens `a` and `b`, multiplies them together, then adds them to the concatenated vectors `high:low`. Returns the lower half of the result, and sets `high` to the upper half.

hwy/ops/arm_neon-inl.h
```diff
@@ -6484,6 +6484,320 @@ HWY_API VFromD<D> SlideDownLanes(D d, VFromD<D> v, size_t amt) {
...
+// ------------------------------- WidenHighMulAdd
+
+#ifdef HWY_NATIVE_WIDEN_HIGH_MUL_ADD
+#undef HWY_NATIVE_WIDEN_HIGH_MUL_ADD
+#else
+#define HWY_NATIVE_WIDEN_HIGH_MUL_ADD
+#endif
+
+template<class D, HWY_IF_U32_D(D)>
+HWY_API Vec128<uint64_t> WidenHighMulAdd(D /* tag */, Vec128<uint32_t> mul,
```
jan-wassenberg, line 6496: By convention, D describes the return type. Would you mind changing all the functions? Here toIF_U64. And to make this clear + code more generic, we can write the return type as VFromD<D>.

Another important bit is to support partial vectors. Hard-coding Vec128<uint32_t> means we can only handle full vectors. All that's required is to make the arguments depend on D. For example, we can set `VN = VFromD<RepartitionToNarrow<D>>` and use that as the arg type.

hwy/ops/generic_ops-inl.h
```diff
@@ -5134,6 +5134,55 @@ HWY_API Vec512<T> operator%(Vec512<T> a, Vec512<T> b) {
...
+#undef HWY_NATIVE_WIDEN_HIGH_MUL_ADD
+#else
+#define HWY_NATIVE_WIDEN_HIGH_MUL_ADD
+#endif
+
+template<class D, HWY_IF_INTEGER(TFromD<D>),
+         class DW = RepartitionToWide<D>, class VW = VFromD<DW>>
+HWY_API VW WidenHighMulAdd(D /* tag */, VFromD<D> mul, VFromD<D> x, VW add) {
+  const Half<D> dh;
+  return WidenMulAdd(dh, UpperHalf(dh, mul), UpperHalf(dh, x), add);
```
jan-wassenberg, line 5164: Recommend using PromoteUpperTo (more efficient on SVE) rather than delegating to WidenMulAdd.

hwy/ops/generic_ops-inl.h
```diff
@@ -5134,6 +5134,55 @@ HWY_API Vec512<T> operator%(Vec512<T> a, Vec512<T> b) {
...
+#if (defined(HWY_NATIVE_WIDEN_MUL_ACCUMULATE) == defined(HWY_TARGET_TOGGLE))
+#ifdef HWY_NATIVE_WIDEN_MUL_ACCUMULATE
+#undef HWY_NATIVE_WIDEN_MUL_ACCUMULATE
+#else
+#define HWY_NATIVE_WIDEN_MUL_ACCUMULATE
+#endif
+
+template<class D, HWY_IF_INTEGER(TFromD<D>), HWY_IF_LANES_GT_D(D, 1),
+         class DW = RepartitionToWide<D>, typename VW = VFromD<DW>>
+HWY_API VW WidenMulAccumulate(D d, VFromD<D> mul, VFromD<D> x, VW low, VW& high) {
```
jan-wassenberg, line 5178: I think we are only intending to expose this one to users? If so, let's move the Widen(High)MulAdd into namespace detail.

jan-wassenberg (changes requested):
Thanks for sending the pull request 👍

hwy/ops/generic_ops-inl.h
awxkee: Done

hwy/ops/arm_neon-inl.h
awxkee: Done. By the way, it is not clear enough to me how does partial vectors in highway.

I changed here to MulAdd, because it fails tests for sometimes for upper half, sometimes for lower. I didn't get where it is has to be where vec length = 1.
```c++
template<class D, HWY_IF_F16(TFromD<D>), HWY_IF_LANES_D(D, 2),
    class DW = RepartitionToWide<D>>
HWY_API VFromD<DW> WidenHighMulAdd(D /* tag */, VFromD<D> mul,
                                   VFromD<D> x, VFromD<DW> add) {
  const DW dw;
  auto ho = Vec64<float32_t>(vfmlal_high_f16(add.raw, mul.raw, x.raw));
  return UpperHalf(dw, ho);
//  return MulAdd(add, PromoteUpperTo(dw, mul), PromoteUpperTo(dw, x));
}
```

g3doc/quick_reference.md
awxkee: Thanks for suggestion! Added

awxkee:
Thank you for you attenrtion. I'll try to add f16 accumulation also.

awxkee:
I limited f16 accumulation only to NEON at the moment. There are compilation issues on AVX, not sure I'm able to handle AVX issues. If you with I may remove f16.

hwy/ops/arm_neon-inl.h
```diff
@@ -6484,6 +6484,406 @@ HWY_API VFromD<D> SlideDownLanes(D d, VFromD<D> v, size_t amt) {
...
+
+#ifdef HWY_NATIVE_WIDEN_HIGH_MUL_ADD
+#undef HWY_NATIVE_WIDEN_HIGH_MUL_ADD
+#else
+#define HWY_NATIVE_WIDEN_HIGH_MUL_ADD
+#endif
+
+namespace detail {
+
+template<class D, HWY_IF_U32(TFromD<D>), HWY_IF_LANES_GT_D(D, 2),
```
jan-wassenberg, line 6497: Nice you've got the SFINAE to handle (disable) the case of N=1 inputs :)
We can simplify a bit to `HWY_IF_U32_D(D)`.
One remaining issue: there is a consistent convention that D describes the return type. Would you mind changing all functions to 
```
template<class D, HWY_IF_Uxx(D), HWY_IF_LANES_GT_D(D, 2),
class DN = RepartitionToNarrow<D>>
HWY_API VFromD<D> WidenHighMulAdd(D /* tag */, VFromD<DN> mul,
                                   VFromD<DN> x, VFromD<D> add) {
```
? So D is the wide, and DN is the narrow input.

hwy/ops/arm_neon-inl.h
```diff
@@ -6484,6 +6484,406 @@ HWY_API VFromD<D> SlideDownLanes(D d, VFromD<D> v, size_t amt) {
...
+
+template<class D, HWY_IF_INTEGER(TFromD<D>), HWY_IF_LANES_GT_D(D, 1),
+    class DW = RepartitionToWide<D>, typename VW = VFromD<DW>>
+HWY_API VW WidenMulAccumulate(D d, VFromD<D> mul, VFromD<D> x, VW low, VW& high) {
+  const Half<decltype(d)> dh;
+  high = detail::WidenHighMulAdd(d, mul, x, high);
+  return detail::WidenMulAdd(dh, LowerHalf(dh, mul), LowerHalf(dh, x), low);
+}
+
+#if HWY_NEON_HAVE_F16C
```
jan-wassenberg, line 6870: This one is only for the f16 conversion instruction. For full support of f16 operations, we check `HWY_HAVE_FLOAT16`.

hwy/tests/widen_mul_test.cc
```diff
@@ -493,6 +493,105 @@ HWY_NOINLINE void TestAllReorderWidenMulAccumulate() {
...
+    auto delta_w = AllocateAligned<TW>(NN);
+    for (size_t p = 0; p < NN; ++p) {
+      // Workaround for incorrect Clang wasm codegen: re-initialize the entire
+      // array rather than zero-initialize once and then toggle lane p.
+      for (size_t i = 0; i < NN; ++i) {
+        delta_w[i] = static_cast<TW>(i == p);
+      }
+      const VW delta0 = Load(dw, delta_w.get());
+      const VW delta1 = Load(dw, delta_w.get() + NN / 2);
+      const VN delta = Combine(dn, DemoteTo(dnh, Add(delta0, delta1)), DemoteTo(dnh, Sub(delta0, delta1)));
```
jan-wassenberg, line 531: I think we could just use OrderedDemote2To?

jan-wassenberg (changes requested):
Nice, no worries about f16, we can leave that as NEON-only for now. You can consider adding a line in g3doc/op_wishlist.md for "f16 WidenMulAccumulate on non-NEON" :)

hwy/ops/arm_neon-inl.h
awxkee: Got it. I'll make changes

hwy/tests/widen_mul_test.cc
awxkee: OrderedDemote2To is not supported for f16. So that the reason I changed to this one. I can create another test for f16 if required

jan-wassenberg: Ah, makes sense. Let's leave as-is :)

hwy/ops/arm_neon-inl.h
jan-wassenberg: Is your question what to do if Lanes()==1? I believe this does not happen because ForShrinkableVectors's kPow2=1 should be ensuring the minimum lane count is 2.

hwy/ops/arm_neon-inl.h
```diff
@@ -6484,6 +6484,388 @@ HWY_API VFromD<D> SlideDownLanes(D d, VFromD<D> v, size_t amt) {
...
+  return Vec128<int32_t>(vmlal_high_s16(add.raw, mul.raw, x.raw));
+#else
+  const Full64<int16_t> dh;
+  return Vec128<int32_t>(
+      vmlal_s16(add.raw, UpperHalf(dh, mul).raw, UpperHalf(dh, x).raw));
+#endif
+}
+
+template<class D, HWY_IF_I32_D(D), class DN = RepartitionToNarrow<D>,
+         HWY_IF_LANES_LE_D(DN, 4), HWY_IF_LANES_GT_D(DN, 2)>
```
jan-wassenberg, line 6553: For lanes == 4, you can use HWY_IF_LANES_D(DN, 4).

hwy/ops/arm_neon-inl.h
```diff
@@ -6484,6 +6484,388 @@ HWY_API VFromD<D> SlideDownLanes(D d, VFromD<D> v, size_t amt) {
...
+  return Vec128<uint32_t>(vmlal_high_u16(add.raw, mul.raw, x.raw));
+#else
+  const Full64<uint16_t> dh;
+  return Vec128<uint32_t>(
+      vmlal_u16(add.raw, UpperHalf(dh, mul).raw, UpperHalf(dh, x).raw));
+#endif
+}
+
+template<class D, HWY_IF_U32_D(D), class DN = RepartitionToNarrow<D>,
+         HWY_IF_LANES_LE_D(DN, 4), HWY_IF_LANES_GT_D(DN, 2)>
```
jan-wassenberg, line 6584: Here also.

hwy/ops/arm_neon-inl.h
```diff
@@ -6484,6 +6484,388 @@ HWY_API VFromD<D> SlideDownLanes(D d, VFromD<D> v, size_t amt) {
...
+  return Vec128<uint16_t>(vmlal_high_u8(add.raw, mul.raw, x.raw));
+#else
+  const Full64<uint8_t> dh;
+  return Vec128<uint16_t>(
+      vmlal_u8(add.raw, UpperHalf(dh, mul).raw, UpperHalf(dh, x).raw));
+#endif
+}
+
+template<class D, HWY_IF_U16_D(D), class DN = RepartitionToNarrow<D>,
+         HWY_IF_LANES_LE_D(DN, 8), HWY_IF_LANES_GT_D(DN, 4)>
```
jan-wassenberg, line 6615: And all subsequent :)

hwy/ops/arm_neon-inl.h
```diff
@@ -6484,6 +6484,388 @@ HWY_API VFromD<D> SlideDownLanes(D d, VFromD<D> v, size_t amt) {
...
+#if HWY_HAVE_FLOAT16
+template<class D, HWY_IF_F32_D(D), HWY_IF_LANES_D(D, 4),
+         class DN = RepartitionToNarrow<D>>
+HWY_API VFromD<D> WidenHighMulAdd(D /* tag */, VFromD<DN> mul,
+                                  VFromD<DN> x, VFromD<D> add) {
+  return VFromD<D>(vfmlalq_high_f16(add.raw, mul.raw, x.raw));
+}
+
+template<class D, HWY_IF_F32_D(D),
+         HWY_IF_LANES_LE_D(D, 3), HWY_IF_LANES_GT_D(D, 1),
```
jan-wassenberg, line 6674: Lanes are always a power of two, so we can fuse these to == 2.

jan-wassenberg (changes requested):
Thanks for making the change! One final tweak to the SFINAE bits and we're good to go :)

hwy/ops/arm_neon-inl.h
awxkee: It happens when returning vector `VFromD<DW>` is 1

awxkee: It was quite surprising when tests were failed for UpperHalf and LowerHalf :)

jan-wassenberg (dismissed):
Nice 👍

awxkee:
Thank you for your help!

hwy/ops/arm_neon-inl.h
```diff
@@ -6484,6 +6484,387 @@ HWY_API VFromD<D> SlideDownLanes(D d, VFromD<D> v, size_t amt) {
...
+#endif
+
+template<class D, HWY_IF_INTEGER(TFromD<D>), class DN = RepartitionToNarrow<D>>
+HWY_API VFromD<D> WidenMulAccumulate(D d, VFromD<DN> mul, VFromD<DN> x,
+                                     VFromD<D> low, VFromD<D>& high) {
+  high = detail::WidenHighMulAdd(d, mul, x, high);
+  return detail::WidenMulAdd(d, LowerHalf(mul), LowerHalf(x), low);
+}
+
+#if HWY_HAVE_FLOAT16
```
jan-wassenberg, line 6851: We have a CI build error in Arm that I don't fully understand. It's that generic_ops line 5163 is redefining the f16 WidenMulAccumulate.
```
hwy/ops/generic_ops-inl.h:5163: error: template parameter redefines default argument
 5163 | template<class D, HWY_IF_F32_D(D), class DN = RepartitionToNarrow<D>>
                        |                   ^
hwy/ops/arm_neon-inl.h:6859: note: previous default template argument defined here
 6859 | template<class D, HWY_IF_F32(TFromD<D>), class DN = RepartitionToNarrow<D>>
```
I think the issue is once we decide that we're going to have a native f16 version, it has to be unconditional.

So arm_neon would do the ifdef/undef/endif and define its f16 overload unconditionally, and *inside* that function we'd check HWY_HAVE_FLOAT16 and either call detail::, or copy the generic implementation.

Does that make sense?
BTW are you testing with GCC on Arm? We're using clang.

awxkee: Hm, ok. I'll make that change. 
I'm testing with Apple Clang with F16 support :)

awxkee: I think better will be define `WidenMulAccumulate` only for NEON, checking inside the function might be not necessary because if device have no support f16 there still will be no implementation for MulAdd

jan-wassenberg: :) It might depend on the build flags. I think our environment does not assume that we have f16.

I think there's still a small change required: when we have a NATIVE_ flag, it's important to always flip that, or never - it must not change between targets. The _TOGGLE flag we are comparing against flips after each target/compile pass, and the NATIVE_ has to match that.

In short, let's always flip the NATIVE_F16 flag, ie move that outside the F16 check. Then, generic_ops will never define the F16 version as desired (either we have no f16 at all, or arm_neon already defined it).

jan-wassenberg:
We seem to have hit a race condition, this can unfortunately happen when quickly pushing multiple PR updates. Attempting to straighten this out in the internal merge infra.

jan-wassenberg:
We've hopefully cleared the logjam now.

jan-wassenberg:
CI reports:
```
error: always_inline function 'vfmlalq_high_f16' requires target feature 'fp16fml', but would be inlined into function 'WidenHighMulAdd' that is compiled without support for 'fp16fml'
 6670 |   return VFromD<D>(vfmlalq_high_f16(add.raw, mul.raw, x.raw));
```
Looks like this is optional and requires extra flags. This kind of fragmentation makes our life harder.
I think what's required is to add a new target as proposed in #2076, I've now added fp16fml to consider there.

For now, how about we comment out (or `#if 0 && ..`) the f16 part?

awxkee:
> CI reports:
> 
> ```
> error: always_inline function 'vfmlalq_high_f16' requires target feature 'fp16fml', but would be inlined into function 'WidenHighMulAdd' that is compiled without support for 'fp16fml'
>  6670 |   return VFromD<D>(vfmlalq_high_f16(add.raw, mul.raw, x.raw));
> ```
> 
> Looks like this is optional and requires extra flags. This kind of fragmentation makes our life harder. I think what's required is to add a new target as proposed in #2076, I've now added fp16fml to consider there.
> 
> For now, how about we comment out (or `#if 0 && ..`) the f16 part?

Ok. I'll comment it out. As my best knoweldge with mobile ARM support F16 is very low ( no one android device support f16, no iPhones ) and only M1, M2, M3 apple processors do support F16 it can might be easier just completely drop this feature. However might be the latest AWS Graviton do support this.

jan-wassenberg:
Yes, we'll check what's available when adding the new target :)

PR #2073, opened 2024-04-10
Replace HWY_ARCH_RVV with HWY_ARCH_RISCV

luyahan:
RVV isn't  arch name. We should use RISCV

### DIFF
--- a/hwy/abort.cc
+++ b/hwy/abort.cc
@@ -60,7 +60,7 @@ HWY_DLLEXPORT HWY_NORETURN void HWY_FORMAT(3, 4)
   fflush(stderr);
 
 // Now terminate the program:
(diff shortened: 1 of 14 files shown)


jan-wassenberg (changes requested):
I agree with you and have regretted this naming. Thanks for making the change. For compatibility with existing code, please add an extra `#define HWY_ARCH_RVV HWY_ARCH_RISCV`, with comment that it is deprecated.

luyahan:
> I agree with you and have regretted this naming. Thanks for making the change. For compatibility with existing code, please add an extra `#define HWY_ARCH_RVV HWY_ARCH_RISCV`, with comment that it is deprecated.

Done

PR #2075, opened 2024-04-10
Minor docs update

Ryo-not-rio:
Add data types for `SetTableIndices` and made it clearer on how to load indices for `TableLookUpBytes`

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -1928,16 +1928,16 @@ their operands into independently processed 128-bit *blocks*.
 
 All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
 
-*   `V`: `{u,i}` \
+*   `V`, `VI`: `{u,i}` \
     <code>VI **TableLookupBytes**(V bytes, VI indices)</code>: returns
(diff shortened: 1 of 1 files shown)


g3doc/quick_reference.md
```diff
@@ -1928,16 +1928,16 @@ their operands into independently processed 128-bit *blocks*.
 
 All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
 
-*   `V`: `{u,i}` \
+*   `V`, `VI`: `{u,i}` \
     <code>VI **TableLookupBytes**(V bytes, VI indices)</code>: returns
     `bytes[indices[i]]`. Uses byte lanes regardless of the actual vector types.
     Results are implementation-defined if `indices[i] < 0` or `indices[i] >=
     HWY_MIN(Lanes(DFromV<V>()), 16)`. `VI` are integers, possibly of a different
```
jan-wassenberg, line 1935: How about something like "`VI` are normal vectors with integer lane type possibly differing from the lanes of `V`. Note that TableLookup*Lanes* has an unspecified, potentially non-vector type type for the indices."?

jan-wassenberg (changes requested):
Thanks for clarifying :) Some suggested updates:

PR #2079, opened 2024-04-11
Optimized `Lt128` operator for RVV

lsrcz:
This pull request adds an optimized implementation of the `Lt128` operator for RVV targets. The new implementation is synthesized using a program synthesizer.

The main computations use LMUL 1/8, which is usually more efficient than vector groups (LMUL > 1) and can outperform full vector registers (LMUL = 1) on some microarchitectures.

### DIFF
--- a/hwy/ops/rvv-inl.h
+++ b/hwy/ops/rvv-inl.h
@@ -5772,6 +5772,40 @@ HWY_API VW RearrangeToOddPlusEven(const VW sum0, const VW sum1) {
 }
 
 // ------------------------------ Lt128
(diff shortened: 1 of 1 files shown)


lsrcz:
The compilation result: https://lt128.godbolt.org/z/xEK6v4f6f.

jan-wassenberg (approved):
I'm sorry it has taken so long, we have now recovered from the crashing RVV toolchain and reactivated tests.

This is a great idea to use the mask bits directly, congrats on finding it via the synthesizer!

PR #2083, opened 2024-04-11
Made fixes to AVX3 detection on macOS

johnplatts:
Updated DetectTargets() to fix a bug that can occur on macOS 12.1 or earlier on x86_64 CPU's that have AVX3 support.

There is a bug on macOS 12.1 or earlier on x86_64 CPU's with AVX3 support where ZMM16-ZMM31, the upper 256 bits of ZMM0-ZMM15, and K0-K7 are not always properly preserved across context switches, and this bug is described at https://community.intel.com/t5/Software-Tuning-Performance/MacOS-Darwin-kernel-bug-clobbers-AVX-512-opmask-register-state/m-p/1327259, https://github.com/golang/go/issues/49233, and https://github.com/simdutf/simdutf/pull/236.

On macOS, it is also possible for bits 5, 6, and 7 of XCR0 to be zero on x86_64 CPU's that support AVX3 before the first AVX512 instruction is executed by the current thread as macOS only preserves ZMM16-ZMM31, the upper 256 bits of ZMM0-ZMM15, and K0-K7 across context switches on threads that have executed an AVX512 instruction.

DetectTargets() has been updated to check for AVX3 support on macOS by doing `IsMacOs12_2OrLater() && HasCpuFeature("hw.optional.avx512f")` instead of checking that bits 5, 6, and 7 of XCR0 are set to avoid false negative results and inconsistent behavior on x86_64 CPU's that support AVX3 on macOS and to disable AVX3 targets on macOS versions earlier than 12.2.

HasCpuFeature is a wrapper around the macOS sysctlbyname API function that returns true if a particular CPU feature is supported and returns false if a CPU feature is not supported.

### DIFF
--- a/hwy/targets.cc
+++ b/hwy/targets.cc
@@ -44,6 +44,11 @@
 
 #endif  // HWY_ARCH_*
 
+#ifdef __APPLE__
+#include <sys/sysctl.h>
+#include <sys/utsname.h>
+#endif  // __APPLE__
+
 namespace hwy {
 namespace {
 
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Thanks @johnplatts for finding and fixing this issue :)

PR #2084, opened 2024-04-12
Enable atomic on riscv

luyahan:
test pass on clang 17.0.6 and gcc13.2


jan-wassenberg (approved):
Thanks! FYI this was for Godbolt, the RVV environment there previously lacked atomic.

PR #2087, opened 2024-04-12
Improve generic implementations of GatherIndexN and ScatterIndexN

Mousius:
Whilst benchmarking https://github.com/numpy/numpy/pull/25781 I realised these were not translated well by the compiler. These implementations make the compiler work less by explicitly using `ExtractLane` and `InsertLane`.

With these, the NumPy benchmarks show `sin`/`cos` not regressing on AArch64 after migrating to Highway :smile_cat:

### DIFF
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -2679,6 +2679,25 @@ HWY_API void MaskedScatterIndex(VFromD<D> v, MFromD<D> m, D d,
   }
 }
 
+template <class D, typename T = TFromD<D>>
+HWY_API void ScatterIndexN(VFromD<D> v, D d, T* HWY_RESTRICT base,
+                           VFromD<RebindToSigned<D>> index,
+                           const size_t max_lanes_to_store) {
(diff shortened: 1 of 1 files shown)


jan-wassenberg (commented):
Thanks, I agree this is a win for SSE4/NEON.
For modern ISAs (AVX3, SVE, RVV) I'd think the Masked versions are actually faster.
Even with the Gather microcoding on Intel, I'd expect it's at least a win on AVX3_SPR and AVX3_ZEN4.

We could `#if HWY_TARGET ==` here, or centralize that into something like HWY_NATIVE_SCATTER(implies also gather) that we set in set_macros, as we do for HWY_NATIVE_FMA. What do you think?

Mousius:
@jan-wassenberg I think I got this right, I've added it to the existing `#if` block

jan-wassenberg (approved):
Ah, that works too. It might make sense to later decouple "native gather" from "it's actually efficient" (especially on Intel), but this is a good first step and already improves things. Thanks for updating the change!

PR #2099, opened 2024-04-20
Fix HWY_NEON compilation errors with GCC 12 and earlier and Clang 12 and earlier

johnplatts:
Updated HWY_TARGET string for the HWY_NEON target for GCC 12 and earlier and Clang 12 and earlier as GCC 12 and earlier and Clang 12 and earlier require "+crypto" instead of "+aes" in the target string to enable the AArch64 AES intrinsics.


jan-wassenberg (approved):
Ah, that explains why we've had some back and forth here. Not ideal that the compilers have changed their behavior :/ Thanks for updating this!

PR #2100, opened 2024-04-20
Fixed CallSinCos compilation error on Arm

johnplatts:
Fixed compilation error with CallSinCos on Arm by replacing `VecArg<V>&` with `V&` since `VecArg<V>&` is a typedef for `const V&` on some platforms.


jan-wassenberg (approved):
Thank you for fixing :)

PR #2109, opened 2024-04-25
Fix the documentation for the type signature of SignBit

lsrcz:
This pull request fixes the documentation for the type signature of SignBit and removes the extra `T` argument.


jan-wassenberg (approved):
Thanks for fixing!

PR #2110, opened 2024-04-26
Add  HWY_ATTAINABLE_RISCV on RISCV

luyahan:


hwy/detect_targets.h
```diff
@@ -621,6 +621,10 @@
 #define HWY_ATTAINABLE_S390X 0
 #endif
 
+#if HWY_ARCH_RISCV && HWY_HAVE_RUNTIME_DISPATCH
+#define HWY_ATTAINABLE_RISCV (HWY_RVV)
```
luyahan, line 625:
```suggestion
#define HWY_ATTAINABLE_RISCV (HWY_BASELINE_RVV)
```

jan-wassenberg (commented):
Can you help me understand the goal here?

The intention of "attainable" is "whatever the compiler is able to compile, independent of compiler flags". I understand that this might not fully work yet for RVV without compiler flags. But the ultimate goal is to have this be `HWY_RVV` as it currently is.

Do you think we should disable HWY_HAVE_RUNTIME_DISPATCH again for RISC-V?

luyahan:
> Can you help me understand the goal here?
> 
> The intention of "attainable" is "whatever the compiler is able to compile, independent of compiler flags". I understand that this might not fully work yet for RVV without compiler flags. But the ultimate goal is to have this be `HWY_RVV` as it currently is.
> 
> Do you think we should disable HWY_HAVE_RUNTIME_DISPATCH again for RISC-V?

When I use foreach_targets.h to build code, it can't build target for RVV on  RISC-V due to HWY_TARGETS doesn't been set HWY_RVV bit.
```
#define HWY_TARGETS \
  (HWY_ATTAINABLE_TARGETS & ((HWY_STATIC_TARGET - 1LL) | HWY_STATIC_TARGET))
```

And  HWY_BASELINE_RVV been set only when `defined(__riscv_v_intrinsic) && __riscv_v_intrinsic >= 11000`
```
// RVV requires intrinsics 0.11 or later, see #1156.
#if HWY_ARCH_RISCV && defined(__riscv_v_intrinsic) && __riscv_v_intrinsic >= 11000
#define HWY_BASELINE_RVV HWY_RVV
#else
#define HWY_
```

jan-wassenberg:
I think we're almost there. It is good you have added the
```
#elif HWY_ARCH_RVV
#define HWY_ATTAINABLE_TARGETS
```
I think now all that's required is to change 
`#define HWY_ATTAINABLE_RISCV (HWY_BASELINE_RVV)` to
`#define HWY_ATTAINABLE_RISCV (HWY_RVV)`.

PR #2115, opened 2024-04-27
Added Exp2 op

johnplatts:
Added implementation of Exp2 to math-inl.h.


jan-wassenberg (approved):
Nice, thanks for adding :) I'll update op_wishlish.md.

PR #2116, opened 2024-04-28
Add GatherIndexNOr

luyahan:


hwy/ops/generic_ops-inl.h
```diff
@@ -2808,13 +2808,35 @@ HWY_API VFromD<D> GatherIndexN(D d, const T* HWY_RESTRICT base,
   }
   return v;
 }
+
+template <class D, typename T = TFromD<D>>
+HWY_API VFromD<D> GatherIndexN(VFromD<D> no, D d, const T* HWY_RESTRICT base,
```
jan-wassenberg, line 2813: I think this function is also intended to be named GatherIndexNOr?

hwy/ops/generic_ops-inl.h
```diff
@@ -2808,13 +2808,35 @@ HWY_API VFromD<D> GatherIndexN(D d, const T* HWY_RESTRICT base,
...
+                               VFromD<RebindToSigned<D>> index,
+                               const size_t max_lanes_to_load) {
+  const RebindToSigned<D> di;
+  using TI = TFromD<decltype(di)>;
+  static_assert(sizeof(T) == sizeof(TI), "Index/lane size must match");
+
+  VFromD<D> v = no;
+  for (size_t i = 0; i < MaxLanes(d); ++i) {
```
jan-wassenberg, line 2821: Maybe `i < HWY_MIN(MaxLanes(d), max_lanes_to_load)`? Or break after the if condition is false?

jan-wassenberg (changes requested):

jan-wassenberg (approved):
👍

PR #2117, opened 2024-04-29
Updated SSE2/SSSE3 Trunc/Ceil/Floor to use ConvertInRangeTo

johnplatts:
Updated SSE2/SSSE3 Trunc/Ceil/Floor ops to do the float to int conversion using ConvertInRangeTo instead of ConvertTo since the SSE2/SSSE3 Trunc/Ceil/Floor ops have an IfThenElse that will simply return `v[i]` if `v[i]` is outside of the range of `MakeSigned<T>`.


jan-wassenberg (approved):
👍

Issue #2125, opened 2024-04-30
tests fail on riscv64 ***Exception: Illegal on Milk-V Pioneer


ncopa:
We have recently enabled the tests in Alpine Linux. There is a significant number of tests that fails:

```
The following tests FAILED:
	  1 - CopyTestGroup/CopyTest.TestAllFill/RVV  # GetParam() = 137438953472 (Failed)
	  3 - CopyTestGroup/CopyTest.TestAllCopy/RVV  # GetParam() = 137438953472 (Failed)
	  5 - CopyTestGroup/CopyTest.TestAllCopyIf/RVV  # GetParam() = 137438953472 (ILLEGAL)
	  7 - FindTestGroup/FindTest.TestAllFind/RVV  # GetParam() = 137438953472 (Failed)
	  9 - FindTestGroup/FindTest.TestAllFindIf/RVV  # GetParam() = 137438953472 (Failed)
	 11 - TransformTestGroup/TransformTest.TestAllGenerate/RVV  # GetParam() = 137438953472 (Failed)
	 13 - TransformTestGroup/TransformTest.TestAllTransform/RVV  # GetParam() = 137438953472 (ILLEGAL)
	 15 - TransformTestGroup/TransformTest.TestAllTransform1/RVV  # GetParam() = 137438953472 (ILLEGAL)
	 17 - TransformTestGroup/TransformTest.TestAllTransform2/RVV  # GetParam() = 137438953472 (ILLEGAL)
	 19 - TransformTestGroup/TransformTest.TestAllReplace/RVV  # GetParam() = 137438953472 (ILLEGAL)
	 50 - HighwayTestGroup/HighwayTest.TestAllCapped/RVV  # GetParam() = 137438953472 (ILLEGAL)
	 52 - HighwayTestGroup/HighwayTest.TestAllMaxLanes/RVV  # GetParam() = 137438953472 (Failed)
	 54 - HighwayTestGroup/HighwayTest.TestAllSet/RVV  # GetParam() = 137438953472 (Failed)
	 56 - HighwayTestGroup/HighwayTest.TestAllOverflow/RVV  # GetParam() = 137438953472 (Failed)
	 58 - HighwayTestGroup/HighwayTest.TestAllClamp/RVV  # GetParam() = 137438953472 (Failed)
	 60 - HighwayTestGroup/HighwayTest.TestAllSignBit/RVV  # GetParam() = 137438953472 (Failed)
[...]

PR #2127, opened 2024-04-30
Fix RVV target detection

johnplatts:
Fixes issue #2125

Updated hwy/targets.cc to check that tail agnostic mode, mask agnostic mode, and that the length of a vuint8m1_t vector is at least 16 bytes on RVV.


jan-wassenberg (approved):
Nice, thank you for implementing the fix!

PR #2128, opened 2024-04-30
Fix for RVV DetectTargets inline asm with Clang

johnplatts:
Added `.option arch, +v` to inline assembly in RVV DetectTargets to fix compilation errors with Clang on RISC-V if the `-march=rv64gcv1p0` or `-march=rv32gcv1p0` option is not specified.


jan-wassenberg (approved):
👍

jan-wassenberg:
(I am following up internally on getting this merged.)

PR #2135, opened 2024-05-03
Added BitShuffle op

johnplatts:
The BitShuffle op was added as AVX3_DL/PPC8/PPC9/PPC10/Z14 have instructions that can extract the bits at specified indices from each lane of an U64 or I64 vector.

The BitShuffle op can be used to speed up the Pack operation in hwy/contrib/bit_pack/bit_pack-inl.h on AVX3_DL/PPC/Z14.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -474,6 +474,7 @@ HWY_TESTS = [
     ("hwy/", "highway_test"),
     ("hwy/", "targets_test"),
     ("hwy/tests/", "arithmetic_test"),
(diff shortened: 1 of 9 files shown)


jan-wassenberg (approved):
Wow! I am impressed that it is feasible to emulate vpshufbitqmb and work around its mask return value :) Nice work, thanks for adding it!

FYI we have an internal use case for 32/64 bit packing which we intend to implement soon.
It will likely use a generic templated implementation, but it should be possible to override.

PR #2136, opened 2024-05-03
Fix Clang compiler warnings in topology.cc

johnplatts:
Added casts to size_t, uint16_t, or int in hwy/contrib/thread_pool/topology.cc to fix compiler warnings


jan-wassenberg (approved):
Thanks for fixing these :)

jan-wassenberg:
I followed up internally on merging this.

PR #2137, opened 2024-05-03
Workaround for GCC warnings with CPU_SET/CPU_ISSET in topology.cc

johnplatts:
Added workaround for the -Wsign-conversion warning that is triggered by `CPU_ISSET(static_cast<int>(lp), &set)` and `CPU_SET(static_cast<int>(lp), &set)` when topology.cc is compiled with GCC by suppressing the -Wsign-conversion warnings that GCC emits with `CPU_ISSET(static_cast<int>(lp), &set)` and `CPU_SET(static_cast<int>(lp), &set)` on Linux and FreeBSD.

The documentation of the CPU_ISSET and CPU_SET macros expect the first argument to CPU_ISSET and CPU_SET to be an int (according to https://www.gnu.org/software/libc/manual/html_node/CPU-Affinity.html), but casting `lp` to an int causes a -Wsign-conversion compiler warning when compiled with GCC.


jan-wassenberg (approved):

PR #2138, opened 2024-05-04
convert most of core to header only

boxerab:
if `per_target.cc` becomes header only, then library cannot be built as there are no longer any .cc files

Here is the error:

```
CMake Error: Cannot determine link language for target "hwy".
CMake Error: CMake can not determine linker language for target: hwy
CMake Error in CMakeLists.txt:
  Exporting the target "hwy" is not allowed since its linker language cannot
  be determined
```


CMakeLists.txt
```diff
@@ -179,13 +179,14 @@ if (NOT HWY_CMAKE_HEADER_ONLY)
...
     hwy/aligned_allocator.cc
     hwy/nanobenchmark.cc
-    hwy/per_target.cc
     hwy/print.cc
     hwy/targets.cc
     hwy/timer.cc
   )
 endif()
 
+list(APPEND HWY_SOURCES hwy/per_target.cc)
```
jan-wassenberg, line 188: As an alternative, https://stackoverflow.com/questions/11801186/cmake-unable-to-determine-linker-language-with-c suggests either setting the language directly, or making Highway an "INTERFACE" library.

hwy/abort.cc
```diff
@@ -20,6 +20,10 @@
 namespace hwy {
 
 namespace {
+
+#ifdef HWY_HEADER_ONLY
+inline
```
jan-wassenberg, line 25: I think for header files, we want static inline just to be sure/safe.

jan-wassenberg (changes requested):

CMakeLists.txt
boxerab: Thanks. The issue is only that we still try to build a library. so we need at least one translation unit.
Not worth worrying about at this stage

hwy/abort.cc
boxerab: Doesn't anonymous namespace take the place of static in this case ?

boxerab: Probably a good idea, but I don't think that's issue with the compile error. Compile currently chokes on
`mask_slide_test.cc`. so it is able to compile a number of other tests successfully. When you have a chance, take a look as I can't make head or tail of the error.

CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ if(POLICY CMP0128)
   cmake_policy(SET CMP0128 NEW)
 endif()
 
-project(hwy VERSION 1.2.0)  # Keep in sync with highway.h version
+project(hwy VERSION 1.2.0 LANGUAGES CXX)  # Keep in sync with highway.h version
```
jan-wassenberg, line 39: I think the ppc builder is complaining about this. If we specify it manually, it should also include C because the try_compile seems to create a .c file.
But is this really necessary? The stackoverflow link below has another way (per-library property) to set the linker language.

hwy/abort.cc
jan-wassenberg: Interesting. What's the error? I'm not seeing this in the build logs, currently they are failing at CMake already due to the CXX project setting.

CMakeLists.txt
boxerab: yes, not necessary

hwy/abort.cc
boxerab: Only one test fails, with these kinds of errors:

```
/home//src/highway/hwy/tests/mask_mem_test.cc:297:7: error: redefinition of ‘class hwy::N_SCALAR::TestStoreMaskBits’
  297 | class TestStoreMaskBits {
      |       ^~~~~~~~~~~~~~~~~
In file included from /home//src/highway/hwy/foreach_target.h:327,
                 from /home//src/highway/hwy/tests/mask_mem_test.cc:27:
/home//src/highway/hwy/tests/mask_mem_test.cc:297:7: note: previous definition of ‘class hwy::N_SCALAR::TestStoreMaskBits’
  297 | class TestStoreMaskBits {
      |       ^~~~~~~~~~~~~~~~~
/home//src/highway/hwy/tests/mask_mem_test.cc:377:19: error: redefinition of ‘void hwy::N_SCALAR::TestAllStoreMaskBits()’
  377 | HWY_NOINLINE void TestAllStoreMaskBits() {
      |                   ^~~~~~~~~~~~~~~~~~~~
In file included from /home//src/highway/hwy/foreach_target.h:327,
                 from /home//src/highway/hwy/tests/mask_mem_test.cc:27:
/home//src/highway/hwy/tests/mask_mem_test.cc:377:19: note: ‘void hwy::N_SCALAR::TestAllStoreMaskBits()’ previously defined here
  377 | HWY_NOINLINE void TestAllStoreMaskBits() {
      |                   ^~~~~~~~~~~~~~~~~~~~
In file included from /home//src/highway/hwy/targets.cc:23,
                 from /home//src/highway/hwy/targets.h:345,
                 from /home//src/highway/hwy/highway.h:22,
                 from /home//src/highway/hwy/timer-inl.h:26,
                 from /home//src/highway/hwy/timer.cc:25,
                 from /home//src/highway/hwy/timer.h:67,
                 from /home//src/highway/hwy/nanobenchmark.h:52,
                 from /home//src/highway/hwy/tests/mask_mem_test.cc:23:
/home//src/highway/hwy/highway.h:226:38: error: ‘N_SSSE3’ has not been declared
  226 | #define HWY_CHOOSE_SSSE3(FUNC_NAME) &N_SSSE3::FUNC_NAME
  ```

boxerab: Unfortunately, error persists with this fix.

jan-wassenberg: hm, that's mystifying. Is that happening for all of the struct/class in this file?
TestStoreMaskBits is the only class, not that that should make a difference.

That class name is not repeated in other files. I'm not seeing anything different about this file compared to others.

boxerab: Thanks, indeed it is strange, there must be something else about this test that is different from all the others. Can you try on your system with the header only flag enabled ?

jan-wassenberg: I've currently got lots going on. Would anyone else be interested in having a look?

PR #2152, opened 2024-05-10
Added support for dynamic dispatch for macOS/iOS/iPadOS on AArch64

johnplatts:
Added support for dynamic dispatch on macOS/iOS/iPadOS on AArch64 as some Apple Silicon CPU's have support for the ARM BF16 extension.


jan-wassenberg (approved):
Thanks for adding! It's great that we now support/detect Apple M2, I think that can handle NEON_BF16.

jan-wassenberg:
FYI internal iOS tests are failing due to svqxtnb_u16 apparently lacking an sve2 attribute. I am investigating.

jan-wassenberg:
@johnplatts would you please rebase this one?

jan-wassenberg (approved):

jan-wassenberg:
Thanks for rebasing! Unfortunately internal CI is failing with a compiler crash for convert_test, but possibly also for all other targets.

```
clang: error: clang frontend command failed due to signal (use -v to see invocation)
Apple clang version 15.0.0 (clang-1500.3.9.4)
Target: arm64-apple-ios17.4
```
I think that corresponds to what we call HWY_COMPILER_CLANG=1600. Are you able to repro this? Should we disable dynamic dispatch, or more likely NEON_BF16, until even more recent compilers? FYI #2170 raises the minimum to 1600.

jan-wassenberg (approved):
Thanks for rebasing. Let's try again, but we might have to raise the minimum compiler for NEON_BF16 in case this still fails.

jan-wassenberg:
Indeed still breaking. I am patching the PR to disable runtime dispatch for Clang < 17, and fix the detection - Apple Clang 15.3 was mistakenly found to be 17 when it should be 16.

jan-wassenberg:
I have patched this PR as mentioned :)

PR #2153, opened 2024-05-10
Made C++11/C++14 compilation fixes to Unroller in bit_pack-inl.h

johnplatts:
Fixed compiler errors in bit_pack-inl.h in C++11/C++14 mode by using the NextUnroller typedef in Unroller::pack and Unroller::unpack, which prevents unnecessary template instantiation in the case where C++17 if constexpr support is not available.


jan-wassenberg (approved):
Thanks for fixing the build - that's a clever solution :)

PR #2154, opened 2024-05-13
Fix build failed on riscv without rvv

luyahan:


jan-wassenberg:
What we are trying to achieve is to compile two codepaths, one with just scalar code, and one with RVV. RVV would only be used if the CPU actually supports it.

In your setting, does the compiler refuse to compile RVV, perhaps because it is configured without it?
If so, I think the better fix would be to disable runtime dispatch entirely because it does not make sense. We will soon allow -DHWY_HAVE_RUNTIME_DISPATCH=0 for that.

johnplatts:
I have made another change in pull request #2159 that should fix compilation errors with GCC 13 or later on RISC-V if targets.cc is compiled without the -march=rv64gcv1p0 or -march=rv32gcv1p0 option.

Issue #2140, opened 2024-05-07
Different test results using Clang when enabling Debug or not on targer RVV


wychlw:
When using clang@d70267fb with highway@e9a2799, cmake with `-DCMAKE_BUILD_TYPE=Debug` option gives different test results between without this option.

Test results without `Debug` option:
```text
99% tests passed, 4 tests failed out of 684

Total Test time (real) =   7.99 sec

The following tests FAILED:
        249 - HwyDemoteTestGroup/HwyDemoteTest.TestAllDemoteToFloat/RVV  # GetParam() = 137438953472 (Failed)
        283 - HwyFloatTestGroup/HwyFloatTest.TestAllCeil/RVV  # GetParam() = 137438953472 (Failed)
        285 - HwyFloatTestGroup/HwyFloatTest.TestAllFloor/RVV  # GetParam() = 137438953472 (Failed)
        651 - SortTestGroup/SortTest.TestAllPartition/RVV  # GetParam() = 137438953472 `(Failed)`
Errors while running CTest
```

Test results with `Debug` option:
```text
98% tests passed, 12 tests failed out of 684

Total Test time (real) =  25.01 sec

The following tests FAILED:
        249 - HwyDemoteTestGroup/HwyDemoteTest.TestAllDemoteToFloat/RVV  # GetParam() = 137438953472 (Failed)
        571 - MatVecTestGroup/MatVecTest.TestAllMatVecBF16/RVV  # GetParam() = 137438953472 (Failed)
        645 - SortTestGroup/SortTest.TestAllFloatInf/RVV  # GetParam() = 137438953472 (Failed)
        651 - SortTestGroup/SortTest.TestAllPartition/RVV  # GetParam() = 137438953472 (Failed)
        655 - SortTestGroup/SortTest.TestAllSort/RVV  # GetParam() = 137438953472 (Failed)
        656 - SortTestGroup/SortTest.TestAllSort/EMU128  # GetParam() = 2305843009213693952 (Failed)
        657 - SortTestGroup/SortTest.TestAllSelect/RVV  # GetParam() = 137438953472 (Failed)
        658 - SortTestGroup/SortTest.TestAllSelect/EMU128  # GetParam() = 2305843009213693952 (Failed)
        659 - SortTestGroup/SortTest.TestAllPartialSort/RVV  # GetParam() = 137438953472 (Failed)
        660 - SortTestGroup/SortTest.TestAllPartialSort/EMU128  # GetParam() = 2305843009213693952 (Failed)
[...]

PR #2164, opened 2024-05-13
Made fixes to RVV Ceil, Floor, and F->F DemoteTo

johnplatts:
Resolves issue #2140.

Updated Ceil and Floor on RVV to use detail::CeilInt and detail::FloorInt (which are wrappers around the RVV __riscv_vfcvt_x_f_v_i*_rm intrinsic) with GCC 14 and later and Clang 17 to fix reordering bugs with changing floating-point rounding mode using inline assembly with Clang 16 and later.

Also changed implementation of RVV Ceil and Floor with GCC 13 and earlier and Clang 16 and earlier to avoid changing the floating-point rounding mode using inline assembly, which fixes bugs with RVV Ceil/Floor with Clang 16 and later and fixes test failures on RVV with Clang 16 and later.

Also updated F64->F32 and F32->F16 DemoteTo on RVV to use the __riscv_vfncvt_f_f_w_f intrinsics instead of the __riscv_vfncvt_rod_f_f_w_f intrinsics as the __riscv_vfncvt_rod_f_f_w_f intrinsics round using round-to-odd mode instead of round-to-nearest mode.

Also updated F64->F16 DemoteTo on RVV to first convert from F64 to F32 using detail::DemoteToF32WithRoundToOdd (which converts F64 values to F32 values using round-to-odd rounding), and then demote from F32 to F16.

### DIFF
--- a/hwy/ops/rvv-inl.h
+++ b/hwy/ops/rvv-inl.h
@@ -2760,9 +2760,15 @@ HWY_API VFromD<Simd<int16_t, N, kPow2>> DemoteTo(
   }
 
 #if HWY_HAVE_FLOAT16 || HWY_RVV_HAVE_F16C
-HWY_RVV_FOREACH_F32(HWY_RVV_DEMOTE_F, DemoteTo, fncvt_rod_f_f_w_f, _DEMOTE_VIRT)
+HWY_RVV_FOREACH_F32(HWY_RVV_DEMOTE_F, DemoteTo, fncvt_f_f_w_f, _DEMOTE_VIRT)
 #endif
-HWY_RVV_FOREACH_F64(HWY_RVV_DEMOTE_F, DemoteTo, fncvt_rod_f_f_w_f, _DEMOTE_VIRT)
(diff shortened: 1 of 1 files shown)


hwy/ops/rvv-inl.h
```diff
@@ -5342,23 +5360,99 @@ HWY_API V Trunc(const V v) {
 }
 
 // ------------------------------ Ceil
+#if (HWY_COMPILER_GCC_ACTUAL && HWY_COMPILER_GCC_ACTUAL >= 1400) || \
```
jan-wassenberg, line 5363: We could also simplify such expressions to just GCC_ACTUAL >= 1400.

jan-wassenberg (approved):
Oh! Thank you for figuring this out.
I believe we previously had the tests passing on FPGA and thus attributed the recent break to QEMU, but it may also have been a compiler change.
BTW clang requires `-frounding-math`, otherwise it reserves the right to ignore or break user-specified rounding modes. I find this behavior very surprising.

PR #2166, opened 2024-05-14
Updated detail::BitsFromMask on Z14/Z15

johnplatts:
Updated detail::BitsFromMask on Z14/Z15 to use ExtractSignBits as Z14/Z15 has the vec_bperm_u128 intrinsic that is equivalent to PPC8/PPC9/PPC10 vec_vbpermq intrinsic.


jan-wassenberg (approved):
Nice :)

PR #2168, opened 2024-05-14
Fix compilation errors if HWY_DISPATCH_MAP is 1

johnplatts:
Fixed compilation errors if HWY_DISPATCH_MAP is 1 by moving the declaration of hwy::AllExports to before the declaration of hwy::FunctionCache as `hwy::FunctionCache<RetType, Args...>::ChooseAndCall<kHash>(Args...)` requires that hwy::AllExports be declared if HWY_DISPATCH_MAP is 1.


jan-wassenberg (approved):
Thanks, we also just noticed :)

PR #2176, opened 2024-05-15
Made UB fixes in targets_test if HWY_SCALAR is enabled

johnplatts:
Fixes for undefined behavior that can occur in FakeFunctionDispatcher and CallFunctionForTarget in hwy/targets_test.cc if the HWY_SCALAR target is enabled.


jan-wassenberg (approved):
Oh right, these are int64 not uint64. Thanks!

PR #2177, opened 2024-05-15
Made fixes to C++11/C++14 HWY_EXPORT_T

johnplatts:
Reimplemented HWY_EXPORT_T for C++11/C++14 to remove dependency on std::map.

The following lines of code in the updated HWY_EXPORT_T macro for C++11/C++14 declares a const object of an unnamed struct that has a different type from any other object:
```
  static const struct {                                                       \
  } HWY_CONCAT(TABLE_NAME, HighwayDispatchExportsKey) = {};                   \
```

`decltype(HWY_CONCAT(TABLE_NAME, HighwayDispatchExportsKey))` then can then be used as a template argument to FunctionCache::ChooseAndCall as `decltype(HWY_CONCAT(TABLE_NAME, HighwayDispatchExportsKey))` is unique for each table that is declared by `HWY_EXPORT_T`.

The reimplementation of C++11/C++14 HWY_EXPORT_T in this pull request fixes compilation errors that are there with GCC 8 or earlier in C++11 or C++14 mode.

Also updated HWY_EXPORT for C++11/C++14 to use the table-based dispatch approach that was used prior to applying commit c90768a38ff2fd7c79e6d0f92e2bb875aa7615af if compiling with MSVC 19.16 or later, GCC, or Clang.

### DIFF
--- a/hwy/highway.h
+++ b/hwy/highway.h
@@ -29,14 +29,6 @@
 #define HWY_DISPATCH_MAP 0
 #endif
 
-#if !HWY_IDE && HWY_DISPATCH_MAP
-#ifdef HWY_NO_LIBCXX
-#error "HWY_NO_LIBCXX requires -std=c++17 or later"
-#else
-#include <map>
-#endif  // HWY_NO_LIBCXX
-#endif  // !HWY_IDE && HWY_DISPATCH_MAP
-
 // This include guard is checked by foreach_target, so avoid the usual _H_
 // suffix to prevent copybara from renaming it. NOTE: ops/*-inl.h are included
 // after/outside this include guard.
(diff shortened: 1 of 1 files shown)


johnplatts:
I ran some builds of Google Highway that integrated the changes in pull request #2177, pull request #2176, and pull request #2178, and I am able to get Google Highway to build correctly on x86_64 with GCC 5 through 14, Clang 5 through 15, and MSVC 19.29 after applying the changes in pull requests #2176, #2177, and #2178.

Here is a result of running the build/test workflow with pull requests #2176, #2177, and #2178 applied: https://github.com/johnplatts/jep_google_highway/actions/runs/9103630808

jan-wassenberg (approved):
Wow! This is very clever, hats off. A table pointer from a static function template keyed by an anon struct. Have not seen this done before :D
Thank you for running all the tests and sending fixes. This is very helpful, we are now close to being able to release v1.2.

PR #2178, opened 2024-05-15
Fixed compilation errors with interleaved_test.cc on x86

johnplatts:
Fixed compilation errors with interleaved_test.cc on x86 if AVX3 targets are enabled.


jan-wassenberg (approved):
Thank you for fixing :) That's a better way.
BTW I have on my TODO to gate our CLs on certain Github actions, which would have caught this.

PR #2180, opened 2024-05-16
Updated CMakeLists.txt to compile Highway in C++17 mode by default

johnplatts:
Updated CMakeLists.txt to compile Google Highway in C++17 mode by default if CMAKE_CXX_STANDARD is not set and the C++ compiler supports C++17 mode.

With the changes made to CMakeLists.txt in this pull request, Google Highway will still be compiled in C++11 mode if CMAKE_CXX_STANDARD is set to 11 and Google Highway will still be compiled in C++14 mode if CMAKE_CXX_STANDARD is set to 14.


jan-wassenberg (approved):
Nice, thanks! Looks like if CMake is not aware of compiler support, then we do not change the behavior, sounds good.

Issue #2183, opened 2024-05-16
HwyConvertTestGroup/HwyConvertTest.TestAllF2IPromoteTo/SSE2 test failure on upcomit `gcc-15`


trofi:
Recent `gcc-15` `master` branch started failing these tests:

```
The following tests FAILED:
        464 - HwyConvertTestGroup/HwyConvertTest.TestAllF2IPromoteTo/AVX2  # GetParam() = 512 (Subprocess aborted)
        465 - HwyConvertTestGroup/HwyConvertTest.TestAllF2IPromoteTo/SSE4  # GetParam() = 2048 (Subprocess aborted)
        466 - HwyConvertTestGroup/HwyConvertTest.TestAllF2IPromoteTo/SSSE3  # GetParam() = 4096 (Subprocess aborted)
        467 - HwyConvertTestGroup/HwyConvertTest.TestAllF2IPromoteTo/SSE2  # GetParam() = 16384 (Subprocess aborted)
```

At first I reported it as a `gcc` bug at https://gcc.gnu.org/PR115115. But now there is a question what `(float)2^31` conversion to `int` should do:
- Intel and AMD CPUs return `2^31` (which both happen to be a "valid" conversion and an overflow error code).
- `gcc` is able to constant-fold it into saturating `INT_MAX`: `2^31-1`

As a result we get this instability in the compiler (partial snippet from the bug)

```c
__attribute__((noipa))
static void TestAllF2IPromoteTo() {
  const __m128i iv = _mm_set_epi32(0x4f000000, 0x4f000000, 0x4f000000, 0x4f000000);
  const __m128  fv = _mm_castsi128_ps(iv);
  const __m128i riv = _mm_cvttps_epi32(fv);

  dump_m128i("riv", &riv);
  assert_m128i(&riv);
}
```

```
$ gcc-15 bug.c -o a -O0 && ./a
riv: 0x80000000 0x80000000 0x80000000 0x80000000
# ok?

$ gcc-15 bug.c -o a -O1 && ./a
riv: 0x7fffffff 0x7fffffff 0x7fffffff 0x7fffffff
Aborted (core dumped)
# bad?
```

AFAIU `highway` exercises it at https://github.com/google/highway/blob/b7cff55896e36e6080909d35ef6fb3d0986b4cfa/hwy/tests/convert_test.cc#L1202:

```c
    constexpr TF kSmallestOutOfToTRangePosVal =
[...]

PR #2189, opened 2024-05-20
Workaround for UB in x86 ConvertInRangeTo with GCC

johnplatts:
Resolves issue #2183

Updated PromoteInRangeTo/DemoteInRangeTo/ConvertInRangeTo/NearestIntInRange ops that are wrappers around SSE2/AVX2/AVX3 float to integer intrinsics on x86 to use inline assembly when compiling with GCC to work around possible undefined behavior if `v[i]` is outside of the range of `TFromD<D>` (after truncation or rounding to nearest).

Similar inline assembly workarounds were implemented for float to integer conversions on PPC8/PPC9/PPC10/NEON in pull request #1919.

Note that `{%1, %0|%0, %1}` is used instead of `%1, %0` in the inline assembly statements as we want the operands to the SSE2/AVX2/AVX3 floating point to integer instructions in the inline assembly code to appear in the correct order in Compiler Explorer if Intel asm syntax is enabled.

`%v` is also used in the inline assembly statements in x86_128-inl.h that are enabled on SSE2/SSSE3/SSE4 as `%v` will prepend `v` before the instruction mnemonic if compiling for AVX2 targets and prepend nothing if compiling for SSE2/SSSE3/SSE4 targets. In other words, `%vcvttpd2dq` in inline assembly code will emit `cvttpd2dq` if compiling for SSE2/SSSE3/SSE4 targets and `vcvttpd2dq` if compiling for AVX2/AVX3 targets, and so on.

The inline assembly code in this pull request is not enabled when compiling with Clang as Clang always expects x86 inline assembly operands to use `%1, %0` syntax instead of `{%1, %0|%0, %1}` syntax (with x86 inline assembly operands using AT&T assembly syntax) and as Clang does not support the `%v` feature that is supported with GCC and ICC.

### DIFF
--- a/hwy/ops/x86_128-inl.h
+++ b/hwy/ops/x86_128-inl.h
@@ -63,6 +63,13 @@ namespace detail {
 #define HWY_AVX3_HAVE_F32_TO_BF16C 0
 #endif
 
+#undef HWY_X86_GCC_INLINE_ASM_VEC_CONSTRAINT
+#if HWY_TARGET <= HWY_AVX3 && HWY_ARCH_X86_64
+#define HWY_X86_GCC_INLINE_ASM_VEC_CONSTRAINT "v"
+#else
+#define HWY_X86_GCC_INLINE_ASM_VEC_CONSTRAINT "x"
+#endif
+
 template <typename T>
 struct Raw128 {
   using type = __m128i;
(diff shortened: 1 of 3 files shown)


hwy/ops/x86_128-inl.h
```diff
@@ -8769,6 +8883,22 @@ HWY_API V AVX2ShlU16Vec128(V v, V bits) {
   return TruncateTo(d, PromoteTo(du32, v) << PromoteTo(du32, bits));
 }
 #elif HWY_TARGET > HWY_AVX2
+
+template <class D32>
+static HWY_INLINE VFromD<D32> Pow2ConvF32ToI32(
+    D32 d32, VFromD<RebindToFloat<D32>> vf32) {
+  const RebindToSigned<decltype(d32)> di32;
+#if HWY_COMPILER_GCC_ACTUAL
+  // ConvertInRangeTo is safe with GCC 11 or later due the inline assembly
```
jan-wassenberg, line 8892: Is it really only GCC 11? I see we use asm on any GCC, and the const-vector check if GCC >= 7.

jan-wassenberg (dismissed):
An extensive workaround, but will be safer for GCC. Thank you!

hwy/ops/x86_128-inl.h
johnplatts: Already made the change to the comment

PR #2195, opened 2024-05-22
Updated Undefined implementation on NEON/PPC/Z14

johnplatts:
Updated `Undefined(d)` on NEON/PPC/Z14 to use the __builtin_nondeterministic_value intrinsic that was added in Clang 17 if available.

`__builtin_nondeterministic_value(Zero(d).raw)` is equivalent to SSE2 _mm_undefined_si128(), _mm_undefined_ps(), or _mm_undefined_si128().

### DIFF
--- a/hwy/ops/ppc_vsx-inl.h
+++ b/hwy/ops/ppc_vsx-inl.h
@@ -244,6 +244,8 @@ HWY_API VFromD<D> Undefined(D d) {
   // Suppressing maybe-uninitialized both here and at the caller does not work,
   // so initialize.
   return Zero(d);
+#elif HWY_HAS_BUILTIN(__builtin_nondeterministic_value)
+  return VFromD<D>{__builtin_nondeterministic_value(Zero(d).raw)};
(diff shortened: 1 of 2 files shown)


hwy/ops/ppc_vsx-inl.h
```diff
@@ -244,6 +244,8 @@ HWY_API VFromD<D> Undefined(D d) {
   // Suppressing maybe-uninitialized both here and at the caller does not work,
   // so initialize.
   return Zero(d);
+#elif HWY_HAS_BUILTIN(__builtin_nondeterministic_value)
```
jan-wassenberg, line 247: Should we move this condition to the first, just in case later GCC add this builtin?

johnplatts: It is probably okay to leave __builtin_nondeterministic_value where it is as the __builtin_nondeterministic_value intrinsic is not available with GCC 14 or the GCC 15 trunk.

PR #2197, opened 2024-05-23
Fix ICX detection

johnplatts:
Fixed compiler error with ICX as HWY_COMPILER_CLANG and HWY_COMPILER_ICX are both non-zero when compiling with icx since icx is a clang-based compiler.


jan-wassenberg (approved):
Thanks for fixing :)

PR #2209, opened 2024-05-29
Allow compilation with -Werror on GNU/Hurd system

malaterre:
For reference:
* https://buildd.debian.org/status/fetch.php?pkg=highway&arch=hurd-i386&ver=1.1.0-1&stamp=1716577724&raw=0


jan-wassenberg (approved):
Thank you for fixing :)

PR #2211, opened 2024-05-29
Fix compilation of shared libaries

malaterre:


jan-wassenberg (changes requested):
Thanks! I think HWY_DLLEXPORT is also required on those function definitions in abort.cc, would you mind adding them as well?

jan-wassenberg (approved):
Thanks for updating!

PR #2213, opened 2024-05-30
Allow compilation of AVX2 on x86

malaterre:
GCC-13 seems to be ok:

* https://buildd.debian.org/status/fetch.php?pkg=highway&arch=i386&ver=1.1.0-3&stamp=1717006825&raw=0

### DIFF
--- a/hwy/detect_targets.h
+++ b/hwy/detect_targets.h
@@ -166,7 +166,12 @@
 
 // 32-bit may fail to compile AVX2/3.
 #if HWY_ARCH_X86_32
+// GCC-13 is ok with AVX2:
+#if (HWY_COMPILER_GCC_ACTUAL && HWY_COMPILER_GCC_ACTUAL >= 1300)
+#define HWY_BROKEN_32BIT (HWY_AVX3 | (HWY_AVX3 - 1))
+#else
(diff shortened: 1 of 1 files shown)


hwy/detect_targets.h
```diff
@@ -164,11 +164,13 @@
 #define HWY_BROKEN_CLANG6 0
 #endif
 
-// 32-bit may fail to compile AVX2/3.
+// 32-bit may fail to compile AVX2/3. GCC-13 was found ok
 #if HWY_ARCH_X86_32
-#define HWY_BROKEN_32BIT (HWY_AVX2 | (HWY_AVX2 - 1))
-#else
+#if (HWY_COMPILER_GCC_ACTUAL && HWY_COMPILER_GCC_ACTUAL >= 1300)
 #define HWY_BROKEN_32BIT 0
```
jan-wassenberg, line 170: HWY_AVX3 and newer are still going to cause trouble (see #2214), so we should change 0 to `(HWY_AVX3 | (HWY_AVX3 - 1))` for GCC13.

jan-wassenberg (approved):
Apologies for the super-long delay! Didn't see your update. Let's land this :)

PR #2223, opened 2024-05-31
Added DemoteToNearestInt and F16/F64 NearestInt ops

johnplatts:
Added F64 to I32 DemoteToNearestInt op as SSE2/SSSE3/SSE4/AVX2/AVX3/RVV have instructions for round-to-nearest F64 to I32 conversions.

Also added NearestInt op for F64 vectors if HWY_HAVE_FLOAT64 is 1 as AVX3/AArch64 NEON have instructions for round-to-nearest F64 to I64 conversions and as x86_64 can do a scalar round-to-nearest conversion from F64 to I64 using the cvtsd2si instruction.

Also added NearestInt op for F16 vectors if HWY_HAVE_FLOAT16 is 1 as NEON_BF16/AVX3_SPR have instructions for round-to-nearest F16 to I16 conversions.

There are also use cases for F64 to I32 DemoteToNearestInt in the implementation of Exp2 in hwy/contrib/math/math-inl.h.

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -1694,9 +1694,17 @@ All functions except `Stream` are defined in cache_control.h.
     Returns an implementation-defined value if the input exceeds the destination
     range.
 
-*   `V`: `f32`; `Ret`: `i32` \
+*   `V`: `f`; `Ret`: `Vec<RebindToSigned<DFromV<V>>>` \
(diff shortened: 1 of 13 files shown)


jan-wassenberg (approved):
Nice to remove one more type restriction. Thanks for adding!

PR #2232, opened 2024-06-03
Made fixes to generic_ops-inl.h BitShuffle impl on big-endian

johnplatts:
Fixed the default BitShuffle implementation on big-endian targets in generic_ops-inl.h, which fixes bugs with BitShuffle on EMU128 on big-endian.

The behavior of BitShuffle in generic_ops-inl.h now matches the behavior of BitShuffle in ppc_vsx-inl.h on big-endian PPC/Z14 with the fixes in this pull request.

TestAllBitShuffle now passes on big-endian EMU128 with the fixes in this pull request.


jan-wassenberg (approved):
Thank you for finding + fixing this :)

PR #2238, opened 2024-06-10
Made enhancements to Exp2

johnplatts:
Added ToNearestInt32 to ExpImpl<float> and ExpImpl<double>, which is a wrapper around NearestInt or DemoteToNearestInt.

Also updated Exp2 to use ExpImpl::ToNearestInt32 as ExpImpl::ToNearestInt32 is more efficient on most targets, including x86, NEON, and RVV.


jan-wassenberg (approved):
Thanks, nice to use the new op :)

PR #2239, opened 2024-06-10
Fixed warning in TestForeach in transform_test.cc

johnplatts:
Changed `pa[i] = 2 * count` to `pa[i] = ConvertScalarTo<T>(2 * count)` to fix compiler warning in TestForeach in hwy/contrib/algo/transform_test.cc.


jan-wassenberg (approved):
👍

PR #2244, opened 2024-06-11
Added Clang 16/17/18 and GCC 13 to GitHub workflow

johnplatts:
Added Clang 16/17/18 and GCC 13 to GitHub workflow

Disabled AVX3_SPR and AVX3_ZEN4 targets with Clang 16/17/18 in GitHub workflow builds due to crashes that occur when several of the Google Highway tests are compiled on x86_64 with Clang 16 or later with all x86 Highway targets enabled. Need to break up the Google Highway tests that crash when compiled with Clang 16 or later with all x86 Highway targets enabled.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -229,6 +229,89 @@ jobs:
           cmake --build out
           ctest --test-dir out
 
+  cmake_ubuntu_2404:
+    name: Build and test ${{ matrix.name }}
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Thanks for adding. Do you happen to have experience with the large Github runners? Those would be helpful to enable so that tests finish faster. We have internal green light to proceed with that.

PR #2247, opened 2024-06-12
Added platform-specific I32/U32 SumOfMulQuadAccumulate on NEON_BF16

johnplatts:
Reimplemented I32/U32 SumOfMulQuadAccumulate on NEON_BF16 since the vdot[q]_s32 and vdot[q]_u32 intrinsics are guaranteed to be available on the NEON_BF16 target, which requires support for the AArch64 DotProd extension.


jan-wassenberg (approved):
Nice, thanks for adding :)

PR #2249, opened 2024-06-18
Added I8/I16/I32/U32/I64/U64 AverageRound ops

johnplatts:
Added I8/I16/I32/U32/I64/U64 AverageRound ops as RVV/SVE2/Z14 have AverageRound instructions for all integer vector types (including both signed integer vectors and unsigned integer vectors) and as NEON/PPC have AverageRound instructions for I8/I16/I32/U32 vectors.


jan-wassenberg (approved):
Super, thank you for adding :)

PR #2255, opened 2024-06-24
Remove HWY_CONTRIB_DLLEXPORT from definition of function

xiaozhuai:
`HWY_CONTRIB_DLLEXPORT` are already set correctly in the declaration. There is no need to dup it on definition of function. Which cause an error on msvc `error C2491: 'hwy::GetThreadAffinity': definition of dllimport function not allowed`

### DIFF
--- a/hwy/contrib/thread_pool/topology.cc
+++ b/hwy/contrib/thread_pool/topology.cc
@@ -64,7 +64,7 @@
 
 namespace hwy {
 
-HWY_CONTRIB_DLLEXPORT bool HaveThreadingSupport() {
+bool HaveThreadingSupport() {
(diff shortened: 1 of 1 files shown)


jan-wassenberg:
Hi @xiaozhuai , just to confirm: is the error still happening in MSVC?

xiaozhuai:
> Hi @xiaozhuai , just to confirm: is the error still happening in MSVC?

No, it has been fixed in https://github.com/google/highway/pull/2229. But it's not include in latest 1.2.0.

jan-wassenberg:
Got it, thanks. We might do another release in a few weeks.

PR #2265, opened 2024-06-28
Added CeilInt and FloorInt ops

johnplatts:
Added CeilInt and FloorInt ops as AArch64 NEON has instructions for round towards negative infinity (floor) and round towards positive infinity (ceiling) floating point to integer conversions and to provide a more efficient platform-specific implementation of CeilInt/FloorInt on SSE2 and SSSE3.


jan-wassenberg (approved):
Nice, thanks for adding :)

Issue #2266, opened 2024-06-28
Implementing Scalbn/ScaleF ops


johnplatts:
I have been working on implementations of the Scalbn/ScaleF ops.

The `Scalbn(V v, VFromD<RebindToUnsigned<DFromV<V>>> exp)` op is equivalent to `std::scalbn(v[i], exp[i])`.

The `ScaleF(V v, V exp)` op is equivalent to `Scalbn(v, ConvertTo(RebindToUnsigned<DFromV<V>>(), Floor(exp)))` or AVX3 _mm*_scalef_ps/_mm*_scalef_pd.

I have added a FloorInt op in pull request #2265 that allows ScaleF to be more efficiently implemented on SSE2/SSSE3/AArch64 NEON.

Should the Scalbn/ScaleF ops be renamed to something else?

jan-wassenberg:
Nice, it seems useful to be able to generate the scalef instructions.
On the naming: I think scalbn is defined in terms of FLT_RADIX, but we only care about FLT_RADIX=2, right?
If so, this is equivalent to ldexp, right? I don't love either of those names. Seems like MulByPow2 would be more clear?

johnplatts:
> Nice, it seems useful to be able to generate the scalef instructions. On the naming: I think scalbn is defined in terms of FLT_RADIX, but we only care about FLT_RADIX=2, right? If so, this is equivalent to ldexp, right? I don't love either of those names. Seems like MulByPow2 would be more clear?

scalbn is equivalent to ldexp for F16/F32/F64 floating point types.

I agree that `MulByPow2(V v, VI exp)` would be more clear for the scalbn/ldexp wrapper and `MulByFloorPow2(V v, V exp)` would make more sense for the wrapper around AVX3 _mm_scalef_ps/pd/ph.

jan-wassenberg:
Implemented, thanks @johnplatts :)

PR #2273, opened 2024-07-16
Added MulByPow2 and MulByFloorPow2 ops

johnplatts:
Resolves issue #2266.

Should `MulByFloorPow2(a, b)` have implementation-defined behavior in the case where `a[i]` is NaN and `b[i]` is negative infinity as the AVX3 _mm_scalef_ph/ps/pd ops return zero in the case where `a[i]` is NaN and `b[i]` is negative infinity whereas the default `MulByFloorPow2(a, b)` implementation returns NaN in any lanes where `a[i]` is NaN and `b[i]` is negative infinity?

### DIFF
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -970,6 +970,154 @@ HWY_API VFromD<RebindToSigned<DFromV<V>>> FloorInt(V v) {
 
 #endif  // HWY_NATIVE_CEIL_FLOOR_INT
 
+// ------------------------------ MulByPow2/MulByFloorPow2
+
+#if (defined(HWY_NATIVE_MUL_BY_POW2) == defined(HWY_TARGET_TOGGLE))
+#ifdef HWY_NATIVE_MUL_BY_POW2
+#undef HWY_NATIVE_MUL_BY_POW2
+#else
+#define HWY_NATIVE_MUL_BY_POW2
+#endif
+
+template <class V, HWY_IF_FLOAT_V(V)>
+HWY_API V MulByPow2(V v, VFromD<RebindToSigned<DFromV<V>>> exp) {
+  const DFromV<decltype(v)> df;
+  const RebindToUnsigned<decltype(df)> du;
+  const RebindToSigned<decltype(df)> di;
+
+  using TF = TFromD<decltype(df)>;
+  using TI = TFromD<decltype(di)>;
+  using TU = TFromD<decltype(du)>;
+
+  using VF = VFromD<decltype(df)>;
+  using VI = VFromD<decltype(di)>;
+
+  constexpr TI kMaxBiasedExp = MaxExponentField<TF>();
+  static_assert(kMaxBiasedExp > 0, "kMaxBiasedExp > 0 must be true");
+
+  constexpr TI kExpBias = static_cast<TI>(kMaxBiasedExp >> 1);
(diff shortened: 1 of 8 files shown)


hwy/ops/generic_ops-inl.h
```diff
@@ -970,6 +970,154 @@ HWY_API VFromD<RebindToSigned<DFromV<V>>> FloorInt(V v) {
...
+
+  // If kZeroOutSignUsingSatSub is true, then val_for_exp_sub will be bitcasted
+  // to a vector that has a smaller lane size than TU for the SaturatedSub
+  // operation below.
+  constexpr bool kZeroOutSignUsingSatSub =
+      ((sizeof(TExpSatSub) * 8) == static_cast<size_t>(kNumOfExpBits));
+
+  // If kZeroOutSignUsingSatSub is true, then the upper
+  // (sizeof(TU) - sizeof(TExpSatSub)) * 8 bits of kExpDecrBy1Bits will be all
+  // ones and the lower sizeof(TExpSatSub) * 8 bits of kExpDecrBy1Bits will be
```
jan-wassenberg, line 1043: This is a bit confusing, we're saying the lower bits are equal to 0...01. The comment is accurate but possible to misunderstand. Not sure there is a better way to say it, though.

jan-wassenberg (approved):
I agree impl-defined for the NaN * -inf case is reasonable.

Very nice, a combination of wizard-like generic/fallback implementation, plus thorough testing and comment. Thanks for adding!

hwy/ops/generic_ops-inl.h
johnplatts: What I meant to say is that if `ExponentBits<TF>()` is a power of 2 that is greater than or equal to 8 (which is true if `TF` is `float` as `ExponentBits<float>()` is equal to 8), then the upper `sizeof(TF)*8 - ExponentBits<TF>()` bits can be zeroed out using a SaturatedSub op using a bitcasted vector that has a lane type of `UnsignedFromSize<(ExponentBits<TF>() / 8)>` on targets other than HWY_SCALAR.

In other words, the following U8 SaturatedSub op can simultaneously decrement the lower 8 bits of an U32 lane by 1 (with unsigned saturation) and zero out the upper 24 bits of each U32 lane on targets other than HWY_SCALAR:
`BitCast(du32, SaturatedSub(BitCast(du8, v), BitCast(du8, Set(du32, uint32_t{0xFFFFFF01u}))))`

jan-wassenberg: Got it, thanks. SaturatedSub is surprisingly useful :)

johnplatts:
> Internal MSVC builds are failing due to timeout. Can we shorten/simplify the test for MSVC?

I have reimplemented TestMulByPow2 and TestMulByFloorPow2, which significantly speeds up build times with MSVC.

I have also discovered that `MulByFloorPow2(nan, inf)` is implementation-defined as AVX3 `_mm_scalef_*(nan, +inf)` returns positive infinity whereas the default implementation of `MulByFloorPow2(nan, inf)` returns NaN.

I have updated TestMulByFloorPow2 to check that `MulByFloorPow2(nan, -inf)` returns 0 or NaN and that `MulByFloorPow2(nan, +inf)` returns positive infinity or NaN.

jan-wassenberg:
Thanks for refactoring the test. CI passed now :)

Interesting behavior from the AVX3 instructions, oh well. I agree it's fine to have differences for nan and +/- inf.

PR #2274, opened 2024-07-16
Workaround for GCC warning in HeapSort/HeapSelect

johnplatts:
Added workaround for -Waggressive-loop-optimizations that is emitted by GCC by `st.Swap(lanes + 0, lanes + i)` in HeapSort/HeapSelect.


jan-wassenberg (approved):

PR #2292, opened 2024-08-07
Made fixes to hwy::IsInteger

johnplatts:
Fixed hwy::IsInteger implementation to ensure that `hwy::IsInteger<T>()` returns true in the case where `!hwy::IsSame<T, RemoveCvRef<T>>() && hwy::IsInteger<RemoveCvRef<T>>()` is true.


jan-wassenberg (approved):
👍

PR #2298, opened 2024-08-09
Resolve TODO in aligned_allocator

scuzqy:
And updated corresponding test.
AllocateAligned was designed to take POD types only.

I also removed `TestAllocateAlignedObjectWithoutDestructor` because objects with constructor/destructor are obviously illegal input.
btw, restriction here can be one `is_trivial`, perhaps?

### DIFF
--- a/hwy/aligned_allocator.h
+++ b/hwy/aligned_allocator.h
@@ -232,7 +232,6 @@ class AlignedFreer {
 
   template <typename T>
   void operator()(T* aligned_pointer) const {
-    // TODO(deymo): assert that we are using a POD type T.
     FreeAlignedBytes(aligned_pointer, free_, opaque_ptr_);
   }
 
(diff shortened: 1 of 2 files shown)


jan-wassenberg (approved):
Thank you for adding this :)

jan-wassenberg:
Thanks for following up. I see that our internal CI had failed and blocked the merge. Investigating..

jan-wassenberg:
Mind if we change the check to:
```
  static_assert(std::is_trivially_copyable<T>::value,
                "AllocateAligned: requires trivially copyable T");
  static_assert(std::is_trivially_destructible<T>::value,
                "AllocateAligned: requires trivially destructible T");
```
? We have internal users that trip over the standard layout and trivial requirements.

jan-wassenberg:
Merged with the proposed modification. Thanks again!

PR #2299, opened 2024-08-11
Enable AVX3_SPR F16 support with Clang 19.1 or later

johnplatts:
Clang 19.1.0 `(++20240810103829+866686180a31-1~exp1~20240810103957.18)` is able to successfully compile all of the Highway tests with AVX3_SPR F16 support enabled with the changes made to dot_test.cc, skeleton_test.cc, and unroller_test.cc in this pull request.

AVX3_SPR F16 support is not enabled when compiling with Clang 18 or earlier as there is a bug in LLVM 18 that causes Clang 18 to crash if AVX3_SPR F16 support is enabled, and that bug is described at https://github.com/llvm/llvm-project/issues/92035.

### DIFF
--- a/hwy/contrib/dot/dot_test.cc
+++ b/hwy/contrib/dot/dot_test.cc
@@ -100,7 +100,7 @@ class TestDot {
     const double max = static_cast<double>(8 * 8 * num);
     HWY_ASSERT(-max <= actual && actual <= max);
     const double tolerance =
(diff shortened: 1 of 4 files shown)


jan-wassenberg (approved):
This is great :D Thank you for filing the Github issue, and updating the tolerances/cast.

PR #2302, opened 2024-08-12
Enable support for F64 vectors on WASM_EMU256 target

johnplatts:
This pull request enables support for F64 vectors on WASM_EMU256 target.


jan-wassenberg (approved):
Thanks for adding, I forgot to do so when enabling it for WASM.

PR #2304, opened 2024-08-13
Fixes to hwy/base.h and vqsort-inl.h

johnplatts:
Updated hwy/base.h to ensure that `<ostream>` is not included by hwy/base.h if HWY_NO_LIBCXX is defined.

Also updated vqsort-inl.h to fix a compiler warning that occurs in the Order typedef in the the Recurse function in non-debug builds.


jan-wassenberg (approved):
Thanks for fixing :)

PR #2309, opened 2024-08-20
Add RVV linker flag in case of building with RVV

k0tran:
As simple as that. One line change in CMakeLists.txt
Tested on risc-v which not supported vectors but compiled it (failed on tests)


k0tran:
Failed because master is failing now. Will rebase on top of working master later, if needed

jan-wassenberg (approved):
Thanks for adding!
I see the arm7 issue, we'll have a look :)

PR #2323, opened 2024-09-07
Made unit testing enhancements if HWY_TEST_STANDALONE is 1

johnplatts:
Added support for `--gtest_filter=` and `--gtest_list_tests` command line arguments if HWY_TEST_STANDALONE is 1 to allow CMake to discover the Highway unit tests using the gtest_discover_tests action without actually running the unit tests and to allow CTest to execute only specific unit tests.

Also added a HWY_TEST_STANDALONE option to CMakeLists.txt that will prevent CMake from downloading Google Test and that will add the `-DHWY_TEST_STANDALONE=1` option to the C++ compiler flags for the Google Highway unit tests if the HWY_TEST_STANDALONE CMake option is set to ON.

The HWY_TEST_STANDALONE CMake option defaults to ON when compiling with MSVC and the HWY_TEST_STANDALONE option defaults to OFF when compiling with GCC or Clang.


hwy/tests/hwy_gtest.h
```diff
@@ -174,18 +178,366 @@ std::string TestParamTargetNameAndT(
...
+          ptr_to_positive_test_filter_patterns->empty()) {
+        ptr_to_positive_test_filter_patterns->emplace_back("*", 1);
+      }
+      ++gtest_filter_str;
+      continue;
+    }
+
+    size_t filter_pattern_len = 1;
+
+    // Find the next filter pattern delimeter character or null terminator
```
jan-wassenberg, line 390: Minor: delimiter

hwy/tests/hwy_gtest.h
```diff
@@ -174,18 +178,366 @@ std::string TestParamTargetNameAndT(
...
+
+  if (!gtest_filter) {
+    gtest_filter = "*";
+  }
+  for (int i = 1; i < argc; i++) {
+    const char* const curr_arg = argv[i];
+    if (!curr_arg) {
+      break;
+    }
+    if (curr_arg[0] == '-' && curr_arg[1] == '-' && curr_arg[2] == 'g' &&
```
jan-wassenberg, line 468: What do you think about hoisting these comparisons into a function that takes two const char*? (for matching list_test's \0, we'd have to do this outside the new function because the test string always has \0)
I'd think shorter/more typical code is preferable for arg parsing.

jan-wassenberg (dismissed):
Nice idea, thanks for adding :)

jan-wassenberg (approved):

PR #2330, opened 2024-09-17
Add cmake check for deducing 32bit or 64bit RISCV

kraj:
Currently its only compilable for RV64 when RVV is enabled, this will extend it to build for RV32 with RVV as well

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,33 @@ if(CHECK_PIE_SUPPORTED)
   endif()
 endif()
 
+if (CMAKE_CXX_COMPILER_ARCHITECTURE_ID MATCHES "RISCV32|RISCV64|RISCV128" OR CMAKE_SYSTEM_PROCESSOR MATCHES "riscv32|riscv64|riscv128")
+  include(CheckCSourceCompiles)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -59,6 +59,32 @@ if(CHECK_PIE_SUPPORTED)
   endif()
 endif()
 
+include(CheckCSourceCompiles)
+
+check_c_source_compiles("
```
jan-wassenberg, line 64: Thanks, this looks good.
Would you mind making one minor change, wrapping this in something like `if (CMAKE_CXX_COMPILER_ARCHITECTURE_ID MATCHES "RISCV32|RISCV64|RISCV128" OR
       CMAKE_SYSTEM_PROCESSOR MATCHES "riscv32|riscv64|riscv128"))` so we do not get extra output on non-RISCV platforms?

kraj: @jan-wassenberg thanks for feedback, I have added the suggestions in the updated PR

jan-wassenberg: Thanks for updating!

jan-wassenberg (approved):

PR #2333, opened 2024-09-27
Reduce LMUL for Dup128MaskFromMaskBits

lsrcz:
This pull request may improve the performance for `Dup128MaskFromMaskBits` on certain micro-architectures.

In the original implementation, the operator uses `Set` operator to load the bits into a full vector register, then reinterprets the result to a mask.

The patch only loads the bits into a sufficient partial vector, which may be faster on certain micro-architectures like SiFive X280.

### DIFF
--- a/hwy/ops/rvv-inl.h
+++ b/hwy/ops/rvv-inl.h
@@ -5278,6 +5278,12 @@ template <size_t kN, HWY_IF_LANES_GT(kN, 31)>
 constexpr unsigned MaxMaskBits() {
   return ~0u;
 }
+
+template <class D>
+constexpr int SufficientPow2ForMask() {
+  return std::max(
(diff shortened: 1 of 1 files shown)


hwy/ops/rvv-inl.h
```diff
@@ -5278,6 +5278,12 @@ template <size_t kN, HWY_IF_LANES_GT(kN, 31)>
 constexpr unsigned MaxMaskBits() {
   return ~0u;
 }
+
+template <class D>
+constexpr int SufficientPow2ForMask() {
+  return std::max(
```
jan-wassenberg, line 5284: How about HWY_MAX instead? Godbolt RVV sometimes has difficulty with std::.

lsrcz: Thanks for pointing out this! I noticed there were no std:: in the codebase, but I wasn't sure what was being used as a substitute.

jan-wassenberg (approved):

PR #2334, opened 2024-09-27
Improve MulEven performance for RVV

lsrcz:
This pull request improves the performance of `MulEven` by using masked multiplication instead of merging the results after computing the higher and lower parts.

A similar optimization can be applied to the `MulOdd` operator, though this would require a `MaskedMulHighOr` operator, which is currently unavailable.

For 8-bit elements, `MulEven` currently uses the immediate value 0x5555, which requires two instructions to construct on RISC-V. There may be a potential optimization by constructing the mask manually instead of relying on `Dup128MaskFromMaskBits`, allowing the use of the smaller immediate 0x55. However, it's just a scalar instruction, so I am unsure if we want to do so.

### DIFF
--- a/hwy/ops/rvv-inl.h
+++ b/hwy/ops/rvv-inl.h
@@ -5655,9 +5655,13 @@ HWY_API V64 BitShuffle(V64 values, VI idx) {
 template <class V, HWY_IF_T_SIZE_ONE_OF_V(V, (1 << 1) | (1 << 2) | (1 << 4)),
           class D = DFromV<V>, class DW = RepartitionToWide<D>>
 HWY_API VFromD<DW> MulEven(const V a, const V b) {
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Nice, looks like we're fusing OddEven into the computation. Thanks for sending the PR!
Is this also automatically found?

I would expect scalar instructions are fine, the scalar pipes might be running ahead and/or idle.

jan-wassenberg:
Nice. I agree CI failures are unrelated and will fix some of that shortly.

I think MulOdd would be rarely used, so let's leave it as-is for now :)

jan-wassenberg:
Sorry about the lengthy delay. I have been manually checking pull request status and rescuing those that got stuck.

PR #2337, opened 2024-10-01
Fixed compilation error in sort_test

johnplatts:
Fixed compilation error in TestAllSortIota in hwy/contrib/sort/sort_test.cc that is triggered if either compiling in C++11/C++14 mode or if VQSORT_ENABLED is 0.


jan-wassenberg (approved):
Thanks for fixing :)

PR #2338, opened 2024-10-01
Create devcontainer.json

mustafacco7:


mustafacco7:
[Foreign architectures](https://github.com/google/highway/actions/workflows/multiarch.yml)

jan-wassenberg:
Thanks for sending the pull request. Can you help me understand what the devcontainer is for?

The "multiarch" check which is failing, times out because it takes long and the default runners are too slow. We have not yet been able to get larger runners which would solve this. However, the failure of the test is not a blocker, it is currently expected.

PR #2339, opened 2024-10-02
Create devcontainer.json

mustafacco7:


mustafacco7 (commented):
ghcr.io/google/highway/run-on-arch-google-highway-foreign-architectures-armv7-ubuntu-latest:latest

PR #2343, opened 2024-10-07
Improve the Eq128 and Ne128 performance for RVV

lsrcz:
This pull request optimizes the `Eq128` and `Ne128` operators with a similar approach as #2079, where the mask values are treated as partial vectors to avoid the huge overhead introduced by the operations on the large vector group coming from `VecFromMask` when LMUL is large.


jan-wassenberg (approved):
Nice, thank you for sending these :)

paperh8rrr (commented):
hoipxid:Rvv-inl.clfr.PPLTM

Issue #2345, opened 2024-10-07
Adding wrappers for __riscv_vget* and __riscv_vset* for non-tuple types


lsrcz:
RVV provides the `__riscv_vset_v_*_*` and `__riscv_vget_v_*_*` intrinsics for not only tuple types but also for vector groups since v0.11, for example:

```C
vint16m4_t __riscv_vset_v_i16m1_i16m4(vint16m4_t dest, size_t index, vint16m1_t value);
// __riscv_vset_v_i16m1_i16m4(dest, 2, value) copies the register `value` to the third register in the dest vector group
```

They are usually translated to whole register move instructions (e.g., `vmv1r`) and is usually efficient on most microarchitectures, and could potentially be eliminated by the register allocator when the compilers are getting more advanced.

These operations are useful when implementing concat operators like `ConcatUpperLower` when LMUL is not fractional. For example, the current `ConcatUpperLower` is implemented as follows

```C++
template <class D, class V>
HWY_API V ConcatUpperLower(D d, const V hi, const V lo) {
  const size_t half = Lanes(d) / 2;
  const V hi_down = detail::SlideDown(hi, half);
  return detail::SlideUp(lo, hi_down, half);
}
```

For `V=vuint8m2_t`, each of the two slide operations will take 4 cycles on x280. If we implement it with `vget` and `vset`, we can do

```C++
vuint8m2_t ConcatUpperLower(const vuint8m2_t hi, const vuint8m2_t lo) {
  auto v0 = __riscv_vget_v_u8m2_u8m1(lo, 0);
  return __riscv_vset_v_u8m1_u8m2(hi, 0, v0);
}
```

This will be translated to a program that takes 2 cycles by clang (trunk version).

```assembly
ConcatUpperLower:
  vmv1r.v v8, v10
  ret
```

However, I have no idea on how to deal with all the macros to add the operations to highway. Any idea or instructions on this?

lsrcz:
Thanks! Let me have a look at it and see whether I can add the GET_VEC based on TRUNC and EXT.

lsrcz:
Regarding the VIRT, RISC-V only provide vget for non-fractional LMULs, and the best implementation for something similar for fractional LMUL is to use vslidedown to extract the upper half, like what we did in the concat operators.

What I can get is something similar to the following, though I need to change the SlideDown operator to intrinsics. Here the _GET and _GET_VIRT data types are defined for non-fractional and fractional types. 

```C++
[...]

PR #2362, opened 2024-10-23
Add Get/Set for vectors and use them to implement Concat* operators for RVV

lsrcz:
This pull request

1. implements the `Get` and `Set` operators for RVV, which correspond to the `__riscv_vget*` and `__riscv_vset*` intrinsics, respectively.
2. optimizes the `Concat*` operators with `Get` and `Set`.

There are new type categories introduced (`_GET_SET`, `_GET_SET_VIRT`, `_GET_SET_SMALLEST`). The reason why new categories are introduces is:

1. RVV only provide `vget` and `vset` intrinsics for non-fractional vector types as the whole register movement instructions are only available for LMUL >= 1. This corresponds to the `_GET_SET` category. We cannot simply reuse the `_TRUNC` category as it contains fractional vector types.
2. The `_GET_SET_VIRT` category simulates the intrinsics using instructions other than whole register movement. The result is a register with half in LMUL.
3. Similar to `LowerHalf`, it is useful to have `Get` and `Set` also for the smallest LMUL for each SEW. This is the `_GET_SET_SMALLEST` category, and the result would be the same as the original register type for this category.

For the `Get` operator for fractional vector types, we use `Trunc`, which is effectively a no-op, to extract the lower part, and we use `SlideDown` for the upper parts.

For the `Set` operator for fractional vector types, we use `vmv` with the tail undisturbed configuration to set the lower part. We use `SlideUp` for the upper parts.

The `Concat` operator can then be optimized with `Get` and `Set`. For operators other than `ConcatLowerUpper`, we need a `Get` and a `Set`, and the program should be optimal for all SEW/LMUL combinations. For the `ConcatLowerUpper` operator, we will need two `Get`s and two `Set`s. This would be optimal only when the `Get` operation is always a no-op, meaning that the LMUL for the extracted part must be greater or equal to zero.

### DIFF
--- a/hwy/ops/rvv-inl.h
+++ b/hwy/ops/rvv-inl.h
@@ -127,6 +127,26 @@ namespace detail {  // for code folding
   X_MACRO(BASE, CHAR, 64, __, 32, m4, m8, m2, 2, /*MLEN=*/16, NAME, OP) \
   X_MACRO(BASE, CHAR, 64, __, 32, m8, __, m4, 3, /*MLEN=*/8, NAME, OP)
 
+#define HWY_RVV_FOREACH_08_GET_SET(X_MACRO, BASE, CHAR, NAME, OP)     \
+  X_MACRO(BASE, CHAR, 8, 16, __, m2, m4, m1, 1, /*MLEN=*/4, NAME, OP) \
+  X_MACRO(BASE, CHAR, 8, 16, __, m4, m8, m2, 2, /*MLEN=*/2, NAME, OP) \
+  X_MACRO(BASE, CHAR, 8, 16, __, m8, __, m4, 3, /*MLEN=*/1, NAME, OP)
+
+#define HWY_RVV_FOREACH_16_GET_SET(X_MACRO, BASE, CHAR, NAME, OP)     \
+  X_MACRO(BASE, CHAR, 16, 32, 8, m2, m4, m1, 1, /*MLEN=*/8, NAME, OP) \
+  X_MACRO(BASE, CHAR, 16, 32, 8, m4, m8, m2, 2, /*MLEN=*/4, NAME, OP) \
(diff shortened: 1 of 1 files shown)


johnplatts:
@lsrcz Could you please go ahead and merge the changes that I made in the lsrcz_hwy_concat_fixes_120224 branch of the https://github.com/johnplatts/jep_google_highway repository (which can be found at https://github.com/johnplatts/jep_google_highway/commit/914cb69c58c814e70d757ddc4d158e7c27ee43e8) into this pull request?

Here are the git commands for merging the changes in commit https://github.com/johnplatts/jep_google_highway/commit/914cb69c58c814e70d757ddc4d158e7c27ee43e8:
```
git remote add jep_hwy https://github.com/johnplatts/jep_google_highway.git
git fetch jep_hwy lsrcz_hwy_concat_fixes_120224
git merge jep_hwy/lsrcz_hwy_concat_fixes_120224
```

lsrcz:
Hi @johnplatts, thanks for the changes! I am super busy at this time, but I will try to have a look at it as soon as possible.

lsrcz:
Hi @johnplatts,

I finally got time to look at it -- thank you for your excellent work!

I think your solution addresses the issue in the original code where partial vectors weren't handled correctly (related to #2345), which I hadn't had time to resolve. It's an elegant approach, and I'll submit the code for review.

jan-wassenberg (approved):
Very nice, thank you @johnplatts and @lsrcz :)

johnplatts:
@jan-wassenberg What was causing the feedback/copybara check from failing? Are there any additional changes that still need to be made to hwy/ops/rvv-inl.h, or does CI need to be re-run?

jan-wassenberg:
We had an unrelated internal failure, sorry about that. Re-running now.

PR #2365, opened 2024-10-30
Fixed compile error in concat_test.cc

johnplatts:
Updated TestConcatOddEven in concat_test.cc to fix compiler warning with hwy::Unpredictable1 (which returns a signed integer).


jan-wassenberg (commented):
Thanks for fixing! Iota takes care of casting its second argument. Would you like to remove the ConvertScalarTo to shorten the code a bit?

jan-wassenberg (approved):
Thanks for updating 👍

PR #2366, opened 2024-10-31
Made enhancements to SSE2/SSSE3 U16 Min/Max

johnplatts:
Updated implementation of U16 Min on SSE2/SSSE3 to use `_mm_sub_epi16(a.raw, _mm_subs_epu16(a.raw, b.raw))` as `_mm_sub_epi16(a.raw, _mm_subs_epu16(a.raw, b.raw))` is more efficient than `detail::MinU(a, b)` for U16 vectors on SSE2/SSSE3.

Updated implementation of U16 Max on SSE2/SSSE3 to use `_mm_add_epi16(a.raw, _mm_subs_epu16(b.raw, a.raw))` as `_mm_add_epi16(a.raw, _mm_subs_epu16(b.raw, a.raw))` is more efficient than `detail::MaxU(a, b)` for U16 vectors on SSE2/SSSE3.

Clang 14 and later will optimize U16 `detail::MinU(a, b)` down to movdqa+psubusw+psubw on SSE2/SSSE3 when optimizations are enabled, but Clang 13 and earlier, GCC 14 and earlier, and MSVC will fail to optimize U16 `detail::MinU(a, b)` down to movdqa+psubusw+psubw when optimizations are enabled.

Likewise, Clang 14 and later will optimize U16 `detail::MaxU(a, b)` down to psubusw+paddw on SSE2/SSSE3 when optimizations are enabled, but Clang 13 and earlier, GCC 14 and earlier, and MSVC will fail to optimize U16 `detail::MaxU(a, b)` down to psubusw+paddw when optimizations are enabled.

The updated implementation of U16 Min/Max on SSE2/SSSE3 in this pull request improves codegen with GCC, MSVC, and Clang 13 and earlier.

### DIFF
--- a/hwy/ops/x86_128-inl.h
+++ b/hwy/ops/x86_128-inl.h
@@ -5753,7 +5753,8 @@ HWY_API Vec128<uint8_t, N> Min(Vec128<uint8_t, N> a, Vec128<uint8_t, N> b) {
 template <size_t N>
 HWY_API Vec128<uint16_t, N> Min(Vec128<uint16_t, N> a, Vec128<uint16_t, N> b) {
 #if HWY_TARGET >= HWY_SSSE3
-  return detail::MinU(a, b);
+  return Vec128<uint16_t, N>{
+      _mm_sub_epi16(a.raw, _mm_subs_epu16(a.raw, b.raw))};
 #else
   return Vec128<uint16_t, N>{_mm_min_epu16(a.raw, b.raw)};
 #endif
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Nice :) You've often been able to get speedups using saturating arithmetic 👍 
I'm curious how you noticed the codegen here was suboptimal?

johnplatts:
> Nice :) You've often been able to get speedups using saturating arithmetic 👍 I'm curious how you noticed the codegen here was suboptimal?

Here is a snippet on Compiler Explorer that shows the code that is generated for two different implementations of the SSE2 U16 vector Min/Max op with GCC, MSVC, and Clang: https://godbolt.org/z/oPnvrsWfc

jan-wassenberg:
Wow, indeed a big difference 👍

Issue #2356, opened 2024-10-16
Fail to compile vqsort with clang 16 on Darwin M1


1yefuwang1:
Hi maintainers, 
Thank you for making this great library. When [compiling this project on MacOS M1 with clang 16](https://github.com/1yefuwang1/vectorlite/issues/32), I got this error "Calling convention SVE_VectorCall is unsupported on Darwin"
The full error message is:
```
**Version of C++:**
/Library/Developer/CommandLineTools/usr/bin/c++ -v
Apple clang version 16.0.0 (clang-1600.0.26.3)
Target: arm64-apple-darwin24.0.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

**Error detail:**
[...]

PR #2367, opened 2024-10-31
Mark SVE targets as broken on macOS on A64

johnplatts:
Fixes issue #2356.

This pull request markets the SVE targets as broken on macOS/iOS/iPadOS on AArch64 as none of the current Apple CPU's (at least up to and including the latest M4 and A18) currently support SVE or SVE2.


jan-wassenberg (approved):
Thanks for taking care of this :)
FYI I haven't yet heard back from Arm but this seems like a sensible way forward for now.

PR #2377, opened 2024-11-16
Fixed linker error in bench_sort.cc if HWY_TEST_STANDALONE is 1

johnplatts:
Added HWY_TEST_MAIN() to hwy/contrib/sort/bench_sort.cc to fix linker error if HWY_TEST_STANDALONE is 1.


jan-wassenberg (approved):
Thanks for fixing this :)

PR #2396, opened 2024-12-02
Fix for RISC-V CMake detection if cross-compiling for RISC-V target with Clang

johnplatts:
Made a fix to CMakeLists.txt to correctly detect RISCV_XLEN if cross-compiling for RISC-V target with Clang.


jan-wassenberg (approved):
👍

PR #2397, opened 2024-12-02
Enable tuples on RVV with Clang 17 or later

johnplatts:
Updated hwy/ops/set_macros-inl.h to enable tuples for the HWY_RVV target with Clang 17 or later.


jan-wassenberg (approved):
Glad this is now working :)

PR #2411, opened 2024-12-13
Fixes for ZSeries with GCC 9 or earlier or Clang 18 or earlier

johnplatts:
Resolves issue #2409.

Made fixes to F32->F64 PromoteTo and F64->F32 DemoteTo to work around compiler errors on Z14/Z15 with GCC 8 or GCC 9.

Also made fixes to Reverse on Z14 with GCC 8 to work around the missing vec_reve intrinsic on Z14 with GCC 8 or earlier.

Also marked Z15 target as broken with GCC 8 or earlier as the Z15 target requires GCC 9 or later.

Also marked Z14 and Z15 targets as broken with Clang 18 or earlier due to LLVM bugs in LLVM 18 or earlier on Z14/Z15 that were fixed in LLVM 19, including a bug in LLVM 18 and earlier that was fixed in commit https://github.com/llvm/llvm-project/commit/7e4c6e98fa05f5c3bf14f96365ae74a8d12c6257 (which made its way into LLVM 19 and later).

### DIFF
--- a/hwy/detect_targets.h
+++ b/hwy/detect_targets.h
@@ -294,15 +294,30 @@
 #define HWY_BROKEN_LOONGARCH 0
 #endif
 
+#if HWY_ARCH_S390X
+#if HWY_COMPILER_CLANG && HWY_COMPILER_CLANG < 1900
+// Clang 18 and earlier have bugs with some ZVector intrinsics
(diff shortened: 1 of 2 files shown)


jan-wassenberg (approved):
Great, thanks for working around the compiler issues :)

PR #2416, opened 2024-12-19
Unroller: allow const input

eugeneo:
Currently unroller does not allow const inputs, which seems like a bug to me. The options are either `const_cast` or copying to writeable memory...

The proposed change is technically a breaking one, but only for clients who override unroller `*Load*` methods. Fix is just adding `const`...

Ping me internally if this breaks anything at Google, I'll help.

### DIFF
--- a/hwy/contrib/unroller/unroller_test.cc
+++ b/hwy/contrib/unroller/unroller_test.cc
@@ -452,8 +452,9 @@ struct TestFind {
 
       FindUnit<T> cvtfn(ConvertScalarTo<T>(num - 1));
       MakeSigned<T> idx = 0;
-      Unroller(cvtfn, const_cast<const T*>(a), &idx,
-               static_cast<ptrdiff_t>(num));
+      // Explicitly test input can be const
+      const T* const_a = a;
+      Unroller(cvtfn, const_a, &idx, static_cast<ptrdiff_t>(num));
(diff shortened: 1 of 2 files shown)


eugeneo:
Bazel build fix is in PR#2415

hwy/contrib/unroller/unroller_test.cc
```diff
@@ -452,7 +452,8 @@ struct TestFind {
 
       FindUnit<T> cvtfn(ConvertScalarTo<T>(num - 1));
       MakeSigned<T> idx = 0;
-      Unroller(cvtfn, a, &idx, static_cast<ptrdiff_t>(num));
+      Unroller(cvtfn, const_cast<const T*>(a), &idx,
```
jan-wassenberg, line 455: No need for const_cast here, I think. It's always OK to take a const view of a non-const thing.

jan-wassenberg (changes requested):
This makes sense, thanks for updating. No worries about the incompatible change, unroller's readme says everything in it is subject to change.

hwy/contrib/unroller/unroller_test.cc
eugeneo: Updated the test case to make it more clear.

I added that intentionally so test checks if const inputs are allowed.

jan-wassenberg (approved):
I see, thanks :) Fair enough, makes sense to explicitly show the type.

PR #2420, opened 2024-12-31
Fixes to RVV Concat/Combine ops

johnplatts:
Made fixes to `detail::Set<kIndex>(d, dest, v)` in hwy/ops/rvv-inl.h to ensure that any lanes past the first `Lanes(d) / 2` lanes of `dest` are not overwritten if `!detail::IsFull(d)` is true.

The fixes to `detail::Set<kIndex>(d, dest, v)` in this pull request fix bugs in the RVV Combine, ConcatLowerLower, ConcatLowerUpper, ConcatUpperLower, and ConcatUpperUpper ops if `!detail::IsFull(d)` is true.

### DIFF
--- a/hwy/ops/rvv-inl.h
+++ b/hwy/ops/rvv-inl.h
@@ -3228,67 +3228,101 @@ Get(D d, VFromD<D> v) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Golly, RVV's lack of guarantee of at least 128 bits, plus the combination of LMUL and avl is leading to some complicated handling :/ Thanks for fixing.

PR #2429, opened 2025-01-06
MulRound, MulLower and MulAddLower ops

mazimkhan:
Adding special multiply operations for `arm_sve-inl.h` and `generic_ops-inl.h`: 
- MulRound multiplies two vectors of floats and returns the answer rounded to the nearest integer
- MulLower multiplies the first lane of both input vectors and passes the lanes of vector a for all other lanes.
- MulAddLower performs a multiply-accumulate operation on the first lane of the input vectors and passes the lanes of vector a for all other lanes.

Tests have been added for the operations.

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -726,6 +726,10 @@ All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
     truncating it to the lower half for integer inputs. Currently unavailable on
     SVE/RVV; use the equivalent `Mul` instead.
 
+*   `V`: `f`
+    <code>V **MulRound**(V a, V b)</code>: Multiplies `a[i]` by `b[i]` and rounds
+    the result to the nearest int with ties going to even.
+
 *   `V`: `f`, `VI`: `Vec<RebindToSigned<DFromV<V>>>` \
     <code>V **MulByPow2**(V a, VI b)</code>: Multiplies `a[i]` by `2^b[i]`.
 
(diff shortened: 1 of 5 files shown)


g3doc/quick_reference.md
```diff
@@ -734,6 +738,9 @@ All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
     <code>V **MulHigh**(V a, V b)</code>: returns the upper half of `a[i] *
     b[i]` in each lane.
 
+*   <code>V **MulLower**(V a, V b)</code>: returns `a[0] * b[0]` in the
```
jan-wassenberg, line 741: Ah, this op is also mentioned in another PR. There I commented: this is implemented using the existing MaskedMulOr, and a svptrue. We could expose the latter as a new First1 op, or add a `__builtin_constant_p` check to FirstN's N argument and if that is 1, then we use SV_VL1. This seems like it would be much more widely useful, e.g. for an AddLower etc, and then we would not require two new MulAddLower/MulLower ops. What do you think?

hwy/ops/arm_sve-inl.h
```diff
@@ -1783,6 +1840,18 @@ HWY_API svbool_t IsFinite(const V v) {
   return RebindMask(d, detail::LtN(exp, hwy::MaxExponentField<T>()));
 }
 
+// ------------------------------ MulByPow2/MulByFloorPow2
```
jan-wassenberg, line 1843: This one doesn't have a test yet?

jan-wassenberg (changes requested):

g3doc/quick_reference.md
wbb-ccl: That makes sense. I've removed this op and I'll add the First1 op in a future PR.

hwy/ops/arm_sve-inl.h
wbb-ccl: This is adding an SVE implementation for an existing op. It's covered by the TestMulByPow2 test in mul_by_pow2_test.cc

jan-wassenberg (approved):
👍

PR #2431, opened 2025-01-06
MultiShift, Masked shift and masked interleave

mazimkhan:
Introduces the following operations for `generic_ops-inl.h` and `arm_sve-inl.h` where the use of sve intrinsics will provide a performance gain:
- MultiShift - An arbitrary set of shifts performed on each byte in each 64-bit wide lane of a vector. A fuller description can be found in `g3doc/quick_reference.md`.
- Masked shifts - shifting operations where a mask is applied for alternative behaviour where the mask is false. 
- Masked interleave operations which interleave either the even or odd elements of the two input vectors, unless the mask is false for a given lane. In this case, the lane is set to zero.

Tests have been introduced for all the listed operations.


johnplatts:
AVX3_DL has _mm_multishift_epi64_epi8, _mm256_multishift_epi64_epi8, and _mm512_multishift_epi64_epi8 intrinsics that can be used to implement MultiShift on AVX3_DL.

g3doc/quick_reference.md
```diff
@@ -966,6 +966,49 @@ Per-lane variable shifts (slow if SSSE3/SSE4, or 16-bit, or Shr i64 on AVX2):
     (a[i] << ((sizeof(T)*8 - b[i]) & shift_amt_mask))`, where `shift_amt_mask` is
     equal to `sizeof(T)*8 - 1`.
 
+A compound shift on 64-bit values:
+
+*   `V`: `{u,i}64`, `VI`: `{u,i}8` \
+    <code>V **MultiShift**(V vals, VI indices)</code>: returns a
+    vector with `(vals[i] >> indices[i*8+j]) & 0xff` in byte `j` of `r[i]` for each
```
jan-wassenberg, line 973: Let's define `r`, for example "vector `r`".

g3doc/quick_reference.md
```diff
@@ -966,6 +966,49 @@ Per-lane variable shifts (slow if SSSE3/SSE4, or 16-bit, or Shr i64 on AVX2):
...
+    `VI` must be either `Vec<Repartition<int8_t, DFromV<V>>>` or
+    `Vec<Repartition<uint8_t, DFromV<V>>>`.
+
+    `MultiShift(V vals, VI indices)` is equivalent to the following loop (where `N` is
+    equal to `Lanes(DFromV<V>())`):
+    ```
+    for(size_t i = 0; i < N; i++) {
+      uint64_t shift_result = 0;
+      for(int j = 0; j < 8; j++) {
+        uint64_t rot_result = (v[i] >> indices[i*8+j]) | (v[i] << (64 - indices[i*8+j]));
```
jan-wassenberg, line 988: Because we're only using the lower 8 bits, we can just talk about the right-shift and not a rotate, right?

g3doc/quick_reference.md
```diff
@@ -966,6 +966,49 @@ Per-lane variable shifts (slow if SSSE3/SSE4, or 16-bit, or Shr i64 on AVX2):
...
+        shift_result |= (rot_result & 0xff) << (j * 8);
+      }
+      r[i] = shift_result;
+    }
+    ```
+
+#### Masked Shifts
+*   `V`: `{u,i}` \
+    <code>V **MaskedShiftLeftOrZero**&lt;int&gt;(M mask, V a)</code> returns `a[i] << int` or `0` if
```
jan-wassenberg, line 997: As before, let's drop the OrZero suffix?

g3doc/quick_reference.md
```diff
@@ -2081,6 +2124,12 @@ Ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
     `InterleaveOdd(d, a, b)` is usually more efficient than `OddEven(b,
     DupOdd(a))`.
 
+*   <code>V **InterleaveEvenOrZero**(M m, V a, V b)</code>: Performs the same
```
jan-wassenberg, line 2127: Add Masked prefix and remove OrZero suffix? I notice this doesn't have an SVE implementation yet?

hwy/ops/generic_ops-inl.h
```diff
@@ -574,6 +592,27 @@ HWY_API V MaskedSatSubOr(V no, M m, V a, V b) {
 }
 #endif  // HWY_NATIVE_MASKED_ARITH
 
+// ------------------------------ MaskedShift
+template <int kshift, class V, class M>
```
jan-wassenberg, line 596: Nit: please rename kshift -> kShift.

hwy/ops/generic_ops-inl.h
```diff
@@ -7299,6 +7338,62 @@ HWY_API V BitShuffle(V v, VI idx) {
 
 #endif  // HWY_NATIVE_BITSHUFFLE
 
+// ------------------------------ MultiShift (Rol)
```
jan-wassenberg, line 7341: The ops in parentheses are ops that the implementation of the one in this section uses, for purposes of sorting the ops in the source file. I think this is a copy-paste remnant? We can remove it because this implementation does not seem to use any ops defined in generic_ops-inl.h.

jan-wassenberg (changes requested):
Nice implementation of MultiShift :) As John mentions, please add a TODO for the corresponding x86 _mm512_multishift so we can track that.

hwy/ops/generic_ops-inl.h
```diff
@@ -7299,6 +7338,62 @@ HWY_API V BitShuffle(V v, VI idx) {
...
+  // Extract unaligned bytes from 16 bit segments
+  const auto odd_idx_shift = ZipUpper(du16, idx_shift, Zero(du8));
+  const auto extracted_odd_bytes =
+      Shr(BitCast(du16, odd_segments), odd_idx_shift);
+
+  // Extract the even bytes of each 128 bit block and pack into lower 64 bits
+  const auto extract_mask = Dup128VecFromValues(du8, 0, 2, 4, 6, 8, 10, 12, 14,
+                                                0, 0, 0, 0, 0, 0, 0, 0);
+  const auto even_lanes =
+      BitCast(d64, TableLookupBytes(extracted_even_bytes, extract_mask));
```
jan-wassenberg, line 7389: I think we could use ConcatEven here?

g3doc/quick_reference.md
wbb-ccl: Done

g3doc/quick_reference.md
wbb-ccl: Sorted

g3doc/quick_reference.md
wbb-ccl: I've removed this, we'll add it to a future PR once we have an x86 implementation.

hwy/ops/generic_ops-inl.h
wbb-ccl: Sorted

hwy/ops/generic_ops-inl.h
wbb-ccl: Ah, that makes sense. I've corrected this.

hwy/ops/generic_ops-inl.h
wbb-ccl: That's a better way to do it. I've made that change.

g3doc/quick_reference.md
wbb-ccl: Unfortunately not, in the cases where `indices[x]` are greater than 55 you start to get the lower bits wrapping round.

i.e. For  `indices[x] = 59` the output bits are: 2, 1, 0, 63, 62, 61, 60, 59

jan-wassenberg: Oh, I see - the x86 definition has & 63. Do we want to call it MultiRotateRight then?

g3doc/quick_reference.md
```diff
@@ -1020,6 +1020,49 @@ Per-lane variable shifts (slow if SSSE3/SSE4, or 16-bit, or Shr i64 on AVX2):
...
+*   `V`: `{u,i}` \
+    <code>V **MaskedShiftLeft**&lt;int&gt;(M mask, V a)</code> returns `a[i] << int` or `0` if
+    `mask[i]` is false.
+
+*   `V`: `{u,i}` \
+    <code>V **MaskedShiftRight**&lt;int&gt;(M mask, V a)</code> returns `a[i] >> int` or `0` if
+    `mask[i]` is false.
+
+*   `V`: `{u,i}` \
+    <code>V **MaskedShiftRightOr**&lt;int&gt;(V no, M mask, V a)</code> returns `a[i] >> int` or `no` if
```
jan-wassenberg, line 1059: Minor clarification: no[i] here and below?

hwy/ops/arm_sve-inl.h
```diff
@@ -1103,6 +1103,35 @@ HWY_SVE_FOREACH_I(HWY_SVE_SHIFT_N, ShiftRight, asr_n)
...
+    auto shifts = static_cast<HWY_SVE_T(uint, BITS)>(kBits);                \
+    return sv##OP##_##CHAR##BITS##_z(m, v, shifts);                         \
+  }
+HWY_SVE_FOREACH_UI(HWY_SVE_SHIFT_Z, MaskedShiftLeft, lsl_n)
+HWY_SVE_FOREACH_I(HWY_SVE_SHIFT_Z, MaskedShiftRight, asr_n)
+HWY_SVE_FOREACH_U(HWY_SVE_SHIFT_Z, MaskedShiftRight, lsr_n)
+
+#undef HWY_SVE_SHIFT_Z
+
+// ------------------------------ MaskedShiftRightSameOr
```
jan-wassenberg, line 1120: Stale comment, it's now called MaskedShiftRightOr?

jan-wassenberg (changes requested):

g3doc/quick_reference.md
wbb-ccl: Fixed

g3doc/quick_reference.md
wbb-ccl: Yes, that's a better name

hwy/ops/generic_ops-inl.h
wbb-ccl: It looks like this is not consistently returning the correct answer.

jan-wassenberg: The new implementation looks good to me. What's the issue you see with clang?

jan-wassenberg: hm. We do see compiler bugs several times a year. On the plus side, [LLVM issues](https://github.com/llvm/llvm-project/issues) are pretty lightweight. They will ask for a minimal example, preferably in Godbolt (can paste the code into [this one](https://gcc.godbolt.org/z/rGnjMevKG) as a starting point).

If you can produce such an example with minimal code, I'm happy to have a look to see if we can find an alternative implementation to work around this.

johnplatts: @jan-wassenberg, @wbb-ccl, and @mazimkhan - I made some fixes to MultiRotateRight and TestMultiRotateRight over in the https://github.com/johnplatts/jep_google_highway/commits/cc_shift_fixes_020525 branch

Here are the changes that I made in commit https://github.com/johnplatts/jep_google_highway/commit/56606cc5e267ef0ceba7a71fec794d78561c343c over in the [cc_shift_fixes_020525](https://github.com/johnplatts/jep_google_highway/commits/cc_shift_fixes_020525) branch:
- Updated documentation of MultiRotateRight to reflect differences between little-endian and big-endian targets
- Fixed bugs in MultiRotateRight on big-endian targets
- Added additional overload of MultiRotateRight on RVV to cover the LMUL=1/2 case (where `DFromV<decltype(v)>().Pow2() == 0` (virtual LMUL) and `DFromV<decltype(idx)>().Pow2() == -1` on RVV)
- Added AVX3_DL-specific implementations of MultiRotateRight op
- Fixed bugs in TestMultiRotateRight on big-endian targets
- Fixed undefined behavior in TestMultiRotateRight by replacing `(v_i >> idx) | (v_i << (64 - idx))` (which has undefined behavior if `idx == 0`) with `(static_cast<uint64_t>(v_i) >> idx) | (static_cast<uint64_t>(v_i) << ((-idx) & 63))` (which does not have undefined behavior if `idx == 0` and which also ensures that `v_i` is right-shifted using an unsigned right shift)
- Updated TestMultiRotateRight to test the MultiRotateRight op in the case where `Lanes(DFromV<decltype(v)>()) == 1`

Here are the git commands for merging the changes in commit https://github.com/johnplatts/jep_google_highway/commit/56606cc5e267ef0ceba7a71fec794d78561c343c:
```
git remote add jep_hwy https://github.com/johnplatts/jep_google_highway.git
git fetch jep_hwy cc_shift_fixes_020525
git merge jep_hwy/cc_shift_fixes_020525
```

I tested the changes that I made in commit https://github.com/johnplatts/jep_google_highway/commit/56606cc5e267ef0ceba7a71fec794d78561c343c with different versions of GCC and Clang on both little-endian targets and big-endian targets, and I am able to get TestAllMultiRotateRight to pass on both little-endian targets and big-endian targets with the changes I made in commit https://github.com/johnplatts/jep_google_highway/commit/56606cc5e267ef0ceba7a71fec794d78561c343c.

johnplatts: Here are the links to the workflow runs of building commit https://github.com/johnplatts/jep_google_highway/commit/56606cc5e267ef0ceba7a71fec794d78561c343c:
- https://github.com/johnplatts/jep_google_highway/actions/runs/13170257358
- https://github.com/johnplatts/jep_google_highway/actions/runs/13170257356 (Foreign architectures)

jan-wassenberg (dismissed):

jan-wassenberg (dismissed):

PR #2440, opened 2025-01-14
Defer the call to get timer resolution until needed

kccqzy:
We have a use case where the highway library is linked into a shared object. Currently whenever it is dlopen'ed, the static initializer runs including running the `platform::TimerResolution`. This has been observed in profiling and raised some questions, namely it shouldn't be necessary to initialize the benchmarking code if we are using the highway library but not to benchmark anything.

Furthermore, doing so is against many companies' C++ guidelines, making this a blocker for inclusion into companies' code bases.

This commit changes this variable from a namespace-scope static variable into a static local variable.

### DIFF
--- a/hwy/nanobenchmark.cc
+++ b/hwy/nanobenchmark.cc
@@ -30,7 +30,10 @@
 
 namespace hwy {
 namespace {
-static const timer::Ticks timer_resolution = platform::TimerResolution();
+const timer::Ticks& GetTimerResolution() {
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Thanks for the PR! I agree this is an improvement.
FWIW our Bazel build has nanobenchmark as a separate optional build target/dependency, but in CMake it's unfortunately bundled together with the hwy files.

I had a quick look for other non-local static objects and did not see any.

PR #2446, opened 2025-01-17
Added PositiveInfOrHighestValue and NegativeInfOrLowestValue

johnplatts:
Added `hwy::PositiveInfOrHighestValue<T>()`, which returns positive infinity if `T` is a floating point type or `hwy::HighestValue<T>()` if `T` is an integer type.

Also added `hwy::NegativeInfOrLowestValue<T>()` which returns negative infinity if `T` is a floating point type or `hwy::HighestValue<T>()` if `T` is an integer type.

Also updated EMU128 ReduceMin/ReduceMax to use `hwy::PositiveInfOrHighestValue<T>()` and `hwy::NegativeInfOrLowestValue<T>()`.

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -2414,6 +2414,45 @@ constexpr MakeSigned<T> MaxExponentField() {
   return (MakeSigned<T>{1} << ExponentBits<T>()) - 1;
 }
 
+namespace detail {
+
+template <typename T>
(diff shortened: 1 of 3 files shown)


jan-wassenberg (dismissed):
Nice improvement, thanks :)

PR #2447, opened 2025-01-17
Fix for compiler error with GCC 9 or earlier

johnplatts:
Replaced `__has_builtin(__builtin_unreachable)` with `HWY_HAS_BUILTIN(__builtin_unreachable)` in hwy/base.h to fix a compiler error with GCC 9 or earlier as `__has_builtin` is not available with GCC 9 or earlier.


jan-wassenberg (approved):
Oops, sorry about that. Thanks for fixing!

PR #2453, opened 2025-01-27
Optimise LoadMaskBits for uint8_t on SVE

wbb-ccl:
SVE can load a uint8_t pointer directly into a mask through casting.

This technique is used by Arm and is known to work for GCC and Clang.
See https://gitlab.arm.com/networking/ral/-/blob/7ea3a5f7a03e7a3dd1fecd898db9802d4cd36e89/src/intrinsics.h#L422

### DIFF
--- a/hwy/ops/arm_sve-inl.h
+++ b/hwy/ops/arm_sve-inl.h
@@ -4872,6 +4872,9 @@ HWY_API V AverageRound(const V a, const V b) {
 // `p` points to at least 8 readable bytes, not all of which need be valid.
 template <class D, HWY_IF_T_SIZE_D(D, 1)>
 HWY_INLINE svbool_t LoadMaskBits(D d, const uint8_t* HWY_RESTRICT bits) {
+#if HWY_COMPILER_CLANG || HWY_COMPILER_GCC
+  return *(const svbool_t *)bits;
+#endif
   // TODO(janwas): with SVE2.1, load to vector, then PMOV
   const RebindToUnsigned<D> du;
   const svuint8_t iota = Iota(du, 0);
--- a/hwy/ops/arm_sve-inl.h
+++ b/hwy/ops/arm_sve-inl.h
@@ -4872,8 +4872,7 @@ HWY_API V AverageRound(const V a, const V b) {
 // `p` points to at least 8 readable bytes, not all of which need be valid.
 template <class D, HWY_IF_T_SIZE_D(D, 1)>
 HWY_INLINE svbool_t LoadMaskBits(D d, const uint8_t* HWY_RESTRICT bits) {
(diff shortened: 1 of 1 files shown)


wbb-ccl:
We're currently looking into the company's CLA.

I'm not convinced this method is strictly documented but has been used in Arm's public code and does appear to work. It looks to compile into an ldr instruction that directly loads into the predicate register.

I believe the PMOV instruction coming in SVE2.1 will formalise this into a proper intrinsic.

Whilst the above bug report lists this being fixed in GCC 12.0, godbolt appears to be happy to compile down to 10.2. I've stuck to GCC 12.0 for now.

jan-wassenberg:
Sounds good. Thanks for looking into the CLA and checking versions.
If you want, we can simplify the check to `HWY_COMPILER_CLANG >= 1901 || HWY_COMPILER_GCC_ACTUAL >= 1200`, but I'm OK to leave as-is.

cc-wvda:
> Thanks for the pull request! Are you able to sign a CLA?
> 
@jan-wassenberg I have been informed that our organisation has indeed signed the CLA, but we are still failing the CLA check on the PRs. We have attempted to query this using the email address for CLA submissions, however we are receiving no response. Would it be possible to flag this internally?

jan-wassenberg:
Thanks for following up! I just checked the email address and it's "wil******er​@cambridgeconsultants.com>". Oh, we had a similar issue with Azim's series of recent patches.

What does https://cla.developers.google.com/clas say for you?

I manually confirmed that your corporate CLA is indeed signed. This can take a couple of days to propagate to the automated check. In the meantime, I can approve this PR manually.

jan-wassenberg (approved):

hwy/ops/arm_sve-inl.h
```diff
@@ -4872,6 +4872,9 @@ HWY_API V AverageRound(const V a, const V b) {
 // `p` points to at least 8 readable bytes, not all of which need be valid.
 template <class D, HWY_IF_T_SIZE_D(D, 1)>
 HWY_INLINE svbool_t LoadMaskBits(D d, const uint8_t* HWY_RESTRICT bits) {
+#if HWY_COMPILER_CLANG >= 1901 || HWY_COMPILER_GCC_ACTUAL >= 1200
+  return *(const svbool_t*)bits;
+#endif
```
eustas, line 4877: Could this be changed to `#if/#else/#endif` to avoid unreachable statements? Thanks

wbb-ccl: Certainly

jan-wassenberg (dismissed):

jan-wassenberg (approved):

PR #2456, opened 2025-01-29
Made changes to SVE_256 and SVE2_128 target detection

johnplatts:
This pull request moves detection of support for the SVE_256 and SVE2_128 targets into DetectTargets().

Moving detection of support for SVE_256 and SVE2_128 into DetectTargets() on AArch64 avoids the need for a VectorBytes() call in SupportedTargets() on AArch64.

Checking for SVE_256 and SVE2_128 support in DetectTargets() on AArch64 also allows the SVE_256 and SVE2_128 targets to be correctly detected in the case where the SVE_256 or SVE2_128 targets are enabled but both the HWY_SVE and HWY_SVE2 targets are disabled.

This pull request also ensures that SVE_256 and SVE2_128 support is detected by using the result of the SVE cntb instruction (even in the case where the `__ARM_FEATURE_SVE_BITS` macro is defined) as GCC or Clang might possibly optimize `svcntb_pat(SV_ALL)` to a constant value if Google Highway is compiled with the `-msve-vector-bits` option.

### DIFF
--- a/hwy/targets.cc
+++ b/hwy/targets.cc
@@ -489,6 +489,28 @@ int64_t DetectTargets() {
 }  // namespace x86
 #elif HWY_ARCH_ARM && HWY_HAVE_RUNTIME_DISPATCH
 namespace arm {
+
+#if HWY_ARCH_ARM_A64 && !HWY_OS_APPLE &&        \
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Nice improvements, makes sense, thank you :)

PR #2458, opened 2025-01-31
Added GetBiasedExponent op and AVX3 implementation of GetExponent

johnplatts:
Added the GetBiasedExponent op as PPC9/PPC10 have the vec_extract_exp op that returns the biased exponent of `v[i]`.

The addition of the GetBiasedExponent op also allows the default implementation of GetExponent in generic_ops-inl.h to take advantage of the vec_extract_exp op on PPC9/PPC10.

There are also use cases for the GetBiasedExponent op in hwy/contrib/math/math-inl.h.

Also added platform-specific implementation of GetExponent op on AVX3 targets.


jan-wassenberg (approved):
Nice, thanks for adding the x86/POWER versions :)

PR #2474, opened 2025-02-19
Fix compilation on ppc64le and s390x with gcc-12

miladfarca:
Currently getting the following errors on a native machine with gcc-12 after [this PR](https://github.com/google/highway/pull/2471/files): 
```
spin.h:162:1: error: control reaches end of non-void function [-Werror=return-type]
  162 | }

error: ‘my_reps’ may be used uninitialized [-Werror=maybe-uninitialized]
spin_test.cc:76:14: note: ‘my_reps’ was declared here
spin_test.cc:76:14: note: ‘my_reps’ was declared here
```


jan-wassenberg (approved):
Thanks for fixing :)

PR #2493, opened 2025-02-25
Made enhancements to AverageRound

johnplatts:
Updated AverageRound op to use `(a[i] | b[i]) - ((a[i] ^ b[i]) >> 1)` as `(a[i] >> 1) + (b[i] >> 1) + ((a[i] | b[i]) & 1)`  is equal to `(a[i] | b[i]) - ((a[i] ^ b[i]) >> 1)` for all values of `a[i]` and `b[i]` and as  `(a[i] | b[i]) - ((a[i] ^ b[i]) >> 1)` is usually more efficient than `(a[i] >> 1) + (b[i] >> 1) + ((a[i] | b[i]) & 1)`.

Clang 19 also optimizes `(a[i] | b[i]) - ((a[i] ^ b[i]) >> 1)` for I8/U8/I16/U16/I32/U32 vectors down to a SRHADD/URHADD instruction (which is equivalent to AverageRound) on AArch64 when optimizations are enabled according to a Compiler Explorer snippet over at https://godbolt.org/z/eo9T6TaYd.

### DIFF
--- a/hwy/ops/arm_sve-inl.h
+++ b/hwy/ops/arm_sve-inl.h
@@ -5245,8 +5245,7 @@ HWY_SVE_FOREACH_UI(HWY_SVE_RETV_ARGPVV, AverageRound, rhadd)
 #else
 template <class V, HWY_IF_NOT_FLOAT_NOR_SPECIAL_V(V)>
 HWY_API V AverageRound(const V a, const V b) {
-  return Add(Add(ShiftRight<1>(a), ShiftRight<1>(b)),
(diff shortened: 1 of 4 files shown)


jan-wassenberg (approved):
Impressive. I'm curious how you found/devised this alternate implementation?

aqrit:
formula also appears in the sample chapter of _Hacker's Delight 2nd Edition_

jan-wassenberg:
Thanks for sharing the derivation, I am able to follow, and impressed :)

PR #2495, opened 2025-02-25
Made fixes to Win32 build with GCC/Clang

johnplatts:
The changes to CMakeLists.txt in this pull request fixes build errors with GCC/Clang on Win32 if `HWY_DISABLE_FUTEX` is not defined.

Checks to detect the presence of libsynchronization.a or synchronization.lib, WaitOnAddress, WakeByAddressAll, and WakeByAddressSingle on Win32 are also added to CMakeLists.txt in this pull request.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,6 +136,38 @@ check_cxx_source_compiles(
   HWY_RISCV
 )
 
+if (WIN32)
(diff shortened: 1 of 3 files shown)


jan-wassenberg (approved):
Thanks for fixing!

PR #2498, opened 2025-02-26
Fix build with gcc

miladfarca:
Currently failing on gcc-12 with:
```
error: control reaches end of non-void function [-Werror=return-type]
```


jan-wassenberg (approved):
Thanks for fixing.

miladfarca:
Not a  problem.

PR #2514, opened 2025-03-04
Fix C++23-only attribute warning for HWY_ASSUME

emilmelnikov:
clang with `-Wpedantic` issues a `-Wc++23-attribute-extensions` warning in the strict C++ <=20 mode.


jan-wassenberg (approved):

PR #2517, opened 2025-03-05
Fix build with gcc

miladfarca:
Fixing the following errors with gcc-12:
```
thread_pool.h:77:29: error: conversion from 'uint64_t' {aka 'long unsigned int'} to 'uint32_t' {aka 'unsigned int'} may change value

thread_pool.h:396:63: error: unused parameter 'seq' [-Werror=unused-parameter]

thread_pool.h:390:28: error: unused parameter 'thread' [-Werror=unused-parameter]
```


jan-wassenberg:
Thank you :) I have another large update of ThreadPool underway. Is it OK with you if we wait for that to go in?

PR #2526, opened 2025-03-10
Fix build with gcc

miladfarca:
Currently getting this error with gcc-12:
```
thread_pool.h:1043:48: error: unused parameter 'task' [-Werror=unused-parameter]
```


miladfarca:
Hi @jan-wassenberg , just curious, do you build with gcc or only clang?

jan-wassenberg:
Yes, thanks, please do :) I hope the change goes in today, latest tomorrow.

PR #2528, opened 2025-03-11
Fixes for compiler warnings in thread_pool.h

johnplatts:
Fixes compiler warning with Clang in ThreadPool::SendConfig as the task parameter in the closure lambda is unused in debug builds.

### DIFF
--- a/hwy/contrib/thread_pool/thread_pool.h
+++ b/hwy/contrib/thread_pool/thread_pool.h
@@ -1040,7 +1040,7 @@ class alignas(HWY_ALIGNMENT) ThreadPool {
 
     SetBusy();
 
-    const auto closure = [this, copy](uint64_t task, size_t worker) {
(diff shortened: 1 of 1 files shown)


hwy/contrib/thread_pool/thread_pool.h
```diff
@@ -307,7 +307,14 @@ class alignas(HWY_ALIGNMENT) Worker {  // HWY_ALIGNMENT bytes
   const size_t num_threads_;
   Worker* const workers_;
 
-  HWY_MAYBE_UNUSED uint8_t padding_[HWY_ALIGNMENT - 64 - sizeof(victims_)];
+#if HWY_COMPILER_CLANG
```
jan-wassenberg, line 310: hm, this seems rather clunky. Would (void)padding_ in the ctor work?

jan-wassenberg (commented):
Thanks for fixing! One suggestion/comment:

PR #2529, opened 2025-03-11
Updated AArch64 workflow to use GitHub ARM64 runner

johnplatts:
Updated AArch64 build in .github/workflows/multiarch.yml to use a GitHub ARM64 runner (which runs on an actual AArch64 CPU) instead of using the uraimo/run-on-arch-action (which runs the AArch64 build in a Docker container on an x86_64 CPU where AArch64 executables are run using the QEMU emulator) as using an GitHub ARM64 runner is much faster than running the AArch64 build in a Docker container on an x86_64 host.

### DIFF
--- a/.github/workflows/multiarch.yml
+++ b/.github/workflows/multiarch.yml
@@ -16,9 +16,6 @@ jobs:
       fail-fast: true
       matrix:
         include:
-          - arch: aarch64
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Nice, thanks for adding!

PR #2530, opened 2025-03-11
Added workaround for compiler warning in SVE I8/U8 LoadMaskBits

johnplatts:
Updated SVE I8/U8 LoadMaskBits to cast `bits` to `UnalignedSveMaskT` (which is a typedef of `svbool_t` with the `__aligned__(1)` and `__may_alias__` attributes) using `reinterpret_cast` to work around GCC/Clang compiler warnings.

`*reinterpret_cast<const UnalignedSveMaskT*>(bits)` also assumes that either `Lanes(ScalableTag<uint8_t>()) <= 64` or `Lanes(d) == Lanes(ScalableTag<uint8_t>())` is true, and it is possible for `Lanes(ScalableTag<uint8_t>()) > 64` to be true on the HWY_SVE/HWY_SVE2 targets.

Should an opt-out or opt-in be added for the optimized `*reinterpret_cast<const UnalignedSveMaskT*>(bits)` code path to avoid a crash or undefined behavior in the case where `Lanes(ScalableTag<uint8_t>()) > 64 && Lanes(ScalableTag<uint8_t>()) < Lanes(d)` as `LoadMaskBits(d)` does not expect that any bytes past `bits + 8` be dereferenceable if `Lanes(d) <= 64`?


jan-wassenberg (approved):

PR #2532, opened 2025-03-11
Fixes to CompressBlendedStore on AVX3_ZEN4/AVX3_SPR with Clang

johnplatts:
Updated AVX3 implementation of CompressBlendedStore to fix test failures in TestCompress on AVX3_ZEN4/AVX3_SPR with Clang.

### DIFF
--- a/hwy/ops/x86_avx3-inl.h
+++ b/hwy/ops/x86_avx3-inl.h
@@ -107,6 +107,8 @@ HWY_API V RotateRight(V v) {
 
 // ------------------------------ Compress
 
+#pragma push_macro("HWY_X86_SLOW_COMPRESS_STORE")
+
 #ifndef HWY_X86_SLOW_COMPRESS_STORE  // allow override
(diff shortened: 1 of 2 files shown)


hwy/ops/x86_avx3-inl.h
```diff
@@ -487,6 +495,8 @@ HWY_API size_t CompressBitsStore(VFromD<D> v, const uint8_t* HWY_RESTRICT bits,
   return CompressStore(v, LoadMaskBits(d, bits), d, unaligned);
 }
 
+#pragma pop_macro("HWY_X86_SLOW_COMPRESS_STORE")
```
jan-wassenberg, line 498: Can you help me understand why we pop_macro? It seems `HWY_X86_SLOW_COMPRESS_STORE` is only defined in this file.

jan-wassenberg: Oh, I see, it depends on the target and we want to reset it for each target.

jan-wassenberg (approved):

PR #2545, opened 2025-03-26
Fix build with gcc

miladfarca:
Fixing a number of gcc compilation errors:
```
error: conversion from ‘int64_t’ {aka ‘long int’} to ‘double’ may change value
error: conversion from ‘long int’ to ‘double’ may change value
```


jan-wassenberg (approved):
Thank you for fixing the warnings :)

miladfarca:
> Thank you for fixing the warnings :)

Not a problem.

PR #2548, opened 2025-03-31
Fix the index of the cluster in LP on macOS

ufownl:
This was discussed in [google/gemma.cpp#508](https://github.com/google/gemma.cpp/issues/508).


jan-wassenberg (approved):
Thank you for finding + fixing!

Issue #2568, opened 2025-05-07
Build issues using Visual Studio (latest)


anauta00:
Hi,

Since this revision:
---
Revision: c02c0e318b22b1e456601ec392d80e679ef6bc1e
Author: Highway <no-reply@google.com>
Date: 28/04/2025 09:37:13
Message:
Work around an LLVM inliner crash in HWY_DYNAMIC_DISPATCH(...)

LLVM may attempt to inline incompatible intrinsics (and crash) when this macro
is used. As a workaround, we add a zero-instruction inline asm blob.

PiperOrigin-RevId: 752193594
----
Modified: hwy/highway.h
---
hwy can no longer be build using Visual Studio.

The problem lies with the adjusted code in highway.h:
`// We call hwy::PreventElision(...) to work around a compiler crash where the
// LLVM inliner crashes due to inlining incompatible intrinsics.
#define HWY_DYNAMIC_DISPATCH(FUNC_NAME) ({ \
  auto p = *(HWY_DYNAMIC_POINTER(FUNC_NAME)); \
  hwy::PreventElision(p); \
  p; \
})`

This is not understood by VS.
Gemini suggests to change it to:
`namespace hwy_internal {
template <typename FuncPtrType>
inline FuncPtrType DynamicDispatchWrapper(FuncPtrType ptr) {
  PreventElision(ptr);
  return ptr;
}
} // namespace hwy_internal

#define HWY_DYNAMIC_DISPATCH(FUNC_NAME) \
    hwy_internal::DynamicDispatchWrapper(*(HWY_DYNAMIC_POINTER(FUNC_NAME)))`

For my own use, I changed the code back to the previous line:
`#define HWY_DYNAMIC_DISPATCH(FUNC_NAME) (*(HWY_DYNAMIC_POINTER(FUNC_NAME)))`
but it would be nice if this incompatibility could be fixed.

Thanks and greetings,
Auke Nauta

jan-wassenberg:
Hi, thanks for pointing this out. Out of curiosity, what's the build error, and on which VS?

FYI we will likely roll this back soon.

anauta00:
[...]

PR #2563, opened 2025-04-30
Renamed AVX10_2_512 to AVX10_2 target and other AVX10 fixes

johnplatts:
The AVX10.2 whitepaper has been changed to reflect that all implementations of AVX10.2 and later will support 512-bit vectors, according to a update to the AVX10.2 whitepaper that can be found at https://cdrdv2.intel.com/v1/dl/getContent/784343.

With the change recently made to the AVX10.2 specification, AVX10.2 support implies support for 512-bit vectors and support for the AVX3, AVX3_DL, and AVX3_SPR targets.

Also updated HWY_BASELINE_AVX3, HWY_BASELINE_AVX3_DL, and HWY_BASELINE_AVX3_SPR detection as GCC 15 can fail to define the `__AVX512F__`, `__AVX512BW__`, `__AVX512DQ__`, `__AVX512VL__`, `__AVX512VNNI__`, `__AVX512VBMI__`, `__AVX512VBMI2__`, `__AVX512VPOPCNTDQ__`, `__AVX512BITALG__`, `__AVX512BF16__`, and `__AVX512FP16__` macros if targeting a baseline target of AVX10.2 using the `-march=diamondrapids` or `-mavx10.2` GCC compiler options, even if the presence of AVX10.2 now implies support for all these AVX512 extensions (including support for 512-bit vectors).

Also need to compile Google Highway with the `-DHWY_ENABLE_UMONITOR=0 -DHWY_ENABLE_MONITORX=0` options included in CMAKE_CXX_FLAGS if testing the AVX10_2 target with Intel SDE as Intel SDE cannot emulate the UMONITOR instruction.

Also need to pass the `-DCMAKE_CROSSCOMPILING_EMULATOR='<path-to-sde64>;-dmr;--'` option to cmake (with `<path-to-sde64>` replaced with the path to the `sde64` executable) to run Google Highway unit tests in Intel SDE. Need to download the latest version of Intel SDE, which is version 9.53, to run the Google Highway unit tests (which can be downloaded from https://www.intel.com/content/www/us/en/developer/articles/tool/software-development-emulator.html).

[...]

PR #2571, opened 2025-05-09
Fix compilation error with HWY_DYNAMIC_DISPATCH with MSVC

johnplatts:
Fixes issue #2568.

HWY_DYNAMIC_DISPATCH is updated to use hwy::PreventElision if compiling with GCC or Clang (including clang-cl), but not if compiling with MSVC.


jan-wassenberg (approved):
Thanks :) We are still considering whether to roll back our change, but your solution is definitely an improvement.

johnplatts:
@jan-wassenberg Are the feedback/copybara build errors in this pull request and pull requests #2562 and #2563 caused by the MSVC 2019 compiler error that is fixed in pull request #2572, or are there build errors elsewhere causing the feedback/copybara errors?

jan-wassenberg:
Thanks for following up. It seems to be a regression in QEMU 9.2 affecting Armv7 Abs(). I've reported this internally. We will temporarily disable Armv7 CI.

PR #2572, opened 2025-05-09
Workaround for WorkerAdapter compiler error with MSVC 2019

johnplatts:
Workaround for template deduction compiler error in `WorkerAdapter::operator()(const Spin&, const Wait&)` or `WorkerAdapter::operator()(const Spin&, const Barrier&)` with MSVC 2019.

`this->CallImpl` is used in the updated implementation of `WorkerAdapter::operator()` in this pull request to avoid unwanted ADL (argument-dependent lookup) in `WorkerAdapter::operator()` and to ensure that `WorkerAdapter::CallImpl` is selected.

`WorkerAdapter::operator()` is also declared with `hwy::EnableIf` in the return type instead of a template parameter to work around template deduction compiler errors with MSVC 2019. Note that the return type of `WorkerAdapter::operator()` is still `void` with the changes in this pull request as `hwy::EnableIf<Cond>` is a typedef for `void` if `Cond` is true.

The changes in this pull request fixes compiler errors with matvec_test.cc, thread_pool_test.cc, and sort_test.cc with MSVC 2019.

### DIFF
--- a/hwy/contrib/thread_pool/thread_pool.h
+++ b/hwy/contrib/thread_pool/thread_pool.h
@@ -918,18 +918,36 @@ class WorkerAdapter {
 
   void SetEpoch(uint32_t epoch) { epoch_ = epoch; }
 
-  // Split into separate wait/barrier functions because `ThreadFunc` latches
-  // the config in between them.
-  template <class Spin, class Wait,
-            HWY_IF_SAME(decltype(Wait().Type()), WaitType)>
-  void operator()(const Spin& spin, const Wait& wait) const {
+ private:
+  template <class Spin, class Wait>
+  HWY_INLINE void CallImpl(hwy::SizeTag<1> /* second_param_type_tag */,
+                           const Spin& spin, const Wait& wait) const {
     wait.UntilWoken(worker_, spin, epoch_);
   }
-
-  template <class Spin, class Barrier,
-            HWY_IF_SAME(decltype(Barrier().Type()), BarrierType)>
-  void operator()(const Spin& spin, const Barrier& barrier) const {
+  template <class Spin, class Barrier>
+  HWY_INLINE void CallImpl(hwy::SizeTag<2> /* second_param_type_tag */,
+                           const Spin& spin, const Barrier& barrier) const {
     barrier.WorkerReached(worker_, spin, epoch_);
+  }
+
+ public:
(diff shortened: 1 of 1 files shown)


hwy/contrib/thread_pool/thread_pool.h
```diff
@@ -918,20 +918,38 @@ class WorkerAdapter {
...
+                           const Spin& spin, const Barrier& barrier) const {
     barrier.WorkerReached(worker_, spin, epoch_);
   }
 
+ public:
+  // Split into separate wait/barrier functions because `ThreadFunc` latches
+  // the config in between them.
+  template <class Spin, class Param2>
+  hwy::EnableIf<
+      hwy::IsSame<hwy::RemoveCvRef<decltype(hwy::RemoveCvRef<Param2>().Type())>,
```
jan-wassenberg, line 938: One idea to simplify: how about we add after `, class Param2` a `typename Type = hwy::RemoveCvRef<decltype(hwy::RemoveCvRef<Param2>().Type())`, then we don't have to repeat this below?
Another option: we could just use hwy::IsSameEither.

hwy/contrib/thread_pool/thread_pool.h
```diff
@@ -918,20 +918,38 @@ class WorkerAdapter {
...
+                  WaitType>() ||
+      hwy::IsSame<hwy::RemoveCvRef<decltype(hwy::RemoveCvRef<Param2>().Type())>,
+                  BarrierType>()>
+  operator()(const Spin& spin, const Param2& wait_or_barrier) const {
+    this->CallImpl(
+        hwy::SizeTag<
+            hwy::IsSame<
```
jan-wassenberg, line 945: To maybe reduce the number of lines, we could have a `constexpr int kType = hwy::IsSame<
                hwy::RemoveCvRef<decltype(hwy::RemoveCvRef<Param2>().Type())>,
                WaitType>() ? 1 : 2;` then pass `SizeTag<kType>`?

jan-wassenberg (changes requested):
Thanks for fixing! Two suggestions for simplifying the code:

hwy/contrib/thread_pool/thread_pool.h
johnplatts: Made the requested changes

PR #2573, opened 2025-05-12
Fixed CMake build if HWY_ENABLE_CONTRIB is OFF

johnplatts:
Updated CMakeLists.txt to only build hwy/auto_tune_test.cc if HWY_ENABLE_CONTRIB is ON as hwy/auto_tune.h and hwy/auto_tune_test.cc have dependencies on VQSort.


jan-wassenberg (approved):
Thanks! FYI we have another fix upcoming for profiler_example (adding hwy_contrib there).

Issue #2570, opened 2025-05-08
Question about the Profiler


FabianSchuetze:
Thanks for the wonderful library. 

I am a bit surprised about the profiler results. Do costs for zones include costs for nested zones? 

For example, the result of the profiler example is:
```
Level1                                  :          1 x               0 =  0.000000
Level2                                  :          1 x               0 =  0.000000
ProfilerExample                         :          1 x             135 =  0.000000
Level3                                  :          1 x           37413 =  0.000015
Spin20                                  :          1 x           48426 =  0.000020
spin30                                  :          1 x           72654 =  0.000030
spin60                                  :          1 x          145200 =  0.000060
Spin10                                  :          1 x        72643447 =  0.030028
```
I expected, e.g., ProfilerExample to have the highest costs. Furthermore, costs for Spin10 seem high, especially compared to Spin20 (I expected Spin10 < Spin20). Maybe the compiler is reordering something? Will the memory fences in the ctor and dtor prevent such behaviour in multi-threaded code? 

Also, I wondered if VQSort could be replaced with std::sort in `profiler.h`? VQSort is phenomenal, but the profiler example is then dependent on contrib. I could only build the example by linking against contrib, as follows:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index e1c6bb83..c97cb8c5 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -719,7 +719,7 @@ target_sources(hwy_profiler_example PRIVATE
     hwy/profiler.h)
 target_compile_options(hwy_profiler_example PRIVATE ${HWY_FLAGS})
 target_compile_features(hwy_profiler_example PRIVATE ${HWY_CXX_STD_TGT_COMPILE_FEATURE})
-target_link_libraries(hwy_profiler_example PRIVATE hwy)
+target_link_libraries(hwy_profiler_example PRIVATE hwy hwy_contrib)
 target_link_libraries(hwy_profiler_example PRIVATE ${ATOMICS_LIBRARIES})
 set_target_properties(hwy_profiler_example
     PROPERTIES RUNTIME_OUTPUT_DIRECTORY "examples/")
```
Maybe that can be done in another way, though?

jan-wassenberg:
Thanks for pointing out the link issue, we can quickly fix that.

std::sort could be used. I suppose Print is not always time-critical, but we've also had issues with the C++ library for example on RVV, so I'd prefer to keep VQSort, but feel free to change it locally.

[...]

PR #2575, opened 2025-05-15
Build profiler_example only if HWY_ENABLE_CONTRIB is ON

po3ts:
As the profiler example depends on contrib (see issue #2570, commit db1facc), it should only be built if the contrib target is enabled (i.e. if `HWY_ENABLE_CONTRIB` is set to `ON`). Otherwise, Make will fail with a linker error when `HWY_ENABLE_EXAMPLES=ON` and `HWY_ENABLE_CONTRIB=OFF`.


jan-wassenberg (approved):
Makes sense, thanks for fixing that case.

PR #2603, opened 2025-06-24
Fix errors when building with CMake 4.0

johnplatts:
Changed cmake_minimum_required line to `cmake_minimum_required(VERSION 3.10)` in CMakeLists.txt.in to fix errors when configuring Google Highway build with CMake 4.


jan-wassenberg:
It seems this change has been merged into another and there is no diff remaining.

johnplatts:
> It seems this change has been merged into another and there is no diff remaining.

Thanks for pointing this out. I will close this pull request.

PR #2605, opened 2025-06-26
Fix to I8/I16/I32/I64 Add/Sub/Mul on NEON if compiling with -ftrapv

johnplatts:
Updated NEON I8/I16/I32/I64 `operator+` and `operator-` ops to use unsigned addition or subtraction (a similar workaround was implemented for PPC targets in pull request #1733).

Also updated NEON I8/I16/I32 `operator*` to use unsigned multiplication.

The changes in this pull request avoid test failures if compiling with `-ftrapv` to match the behavior of x86 `_mm_add_epi*`, `_mm_sub_epi*`, and `_mm_mul_epi*` intrinsics.


jan-wassenberg (approved):
👍

PR #2606, opened 2025-06-26
Fix for InvariantTicksPerSecond on PPC if running on QEMU

johnplatts:
It is possible for `__ppc_get_timebase_freq()` to return 0 if running in qemu-ppc64 or qemu-ppc64le, causing `InvariantTicksPerSecond()` to incorrectly return 0.

`InvariantTicksPerSecond()` falls back to `MeasureNominalClockRate()` on PPC64 if `__ppc_get_timebase_freq()` returns a result that is less than or equal to 0, which fixes the issue with `InvariantTicksPerSecond()` when running in qemu-ppc64 or qemu-ppc64le.

SpinTest.TestPingPong now passes when running in qemu-ppc64 or qemu-ppc64le with the changes in this pull request.

### DIFF
--- a/hwy/timer.cc
+++ b/hwy/timer.cc
@@ -76,6 +76,23 @@ static HWY_MAYBE_UNUSED double MeasureNominalClockRate() {
   return max_ticks_per_sec;
 }
 
+#if HWY_ARCH_PPC && defined(__GLIBC__) && defined(__powerpc64__)
+namespace ppc {
+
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Thanks @johnplatts for finding and fixing this :)

jan-wassenberg:
I see that the two pending PRs have failed CI due to a compiler update on MSAN. This is now fixed and we are re-starting CI.

Issue #2294, opened 2024-08-08
Meson WrapDB Entry


WillAyd:
Would it be possible to add Highway to the [Meson WrapDB](https://mesonbuild.com/Wrapdb-projects.html)? This would make it easy for meson users to use this as a subproject

jan-wassenberg:
Sounds like a good idea :) Would you like to send them a pull request to do that?

WillAyd:
Sure!

jan-wassenberg:
(Looks like this is stuck due to mesonbuild's CI.)

WillAyd:
Thanks for the reminder on this. I'll take another look this week - generally Meson doesn't require every platform to be green in CI

WillAyd:
@jan-wassenberg would you also consider housing the meson.build configuration files in this repository? After having done a few more projects, it is a bit easier to manage when the files are in the host repository rather than applied as patches downstream in the Meson wrapdb.

jan-wassenberg:
We can do that, making an effort to add new source files there as well, though without promising it will continue to work.

PR #2645, opened 2025-08-01
Add support for Meson build system

WillAyd:
Closes https://github.com/google/highway/issues/2294


WillAyd:
Oh cool I just noticed @jcapriot did a similar change in https://github.com/google/highway/pull/2641

Let's stick with that - will add any comments from this to the other PR

Issue #2679, opened 2025-08-16
1.3.0 fails on powerpc: `thread_pool.h: error: static assertion failed`


barracuda156:
I think this is a new breakage:
```
[ 97%] Building CXX object CMakeFiles/hwy_profiler_example.dir/hwy/examples/profiler_example.cc.o
/opt/local/bin/g++-mp-14 -DHWY_SHARED_DEFINE -DTOOLCHAIN_MISS_ASM_HWCAP_H -DTOOLCHAIN_MISS_SYS_AUXV_H -I/opt/local/var/macports/build/highway-ccda4a2a/work/highway-1.3.0 -pipe -Os -DNDEBUG -I/opt/local/include -D_GLIBCXX_USE_CXX11_ABI=0 -std=c++17 -arch ppc -mmacosx-version-min=10.6 -fPIE -fvisibility=hidden -fvisibility-inlines-hidden -Wno-builtin-macro-redefined -D__DATE__=\"redacted\" -D__TIMESTAMP__=\"redacted\" -D__TIME__=\"redacted\" -fmerge-all-constants -Wall -Wextra -Wconversion -Wsign-conversion -Wvla -Wnon-virtual-dtor -Wcast-align -fmath-errno -fno-exceptions -Wno-psabi -MD -MT CMakeFiles/hwy_profiler_example.dir/hwy/examples/profiler_example.cc.o -MF CMakeFiles/hwy_profiler_example.dir/hwy/examples/profiler_example.cc.o.d -o CMakeFiles/hwy_profiler_example.dir/hwy/examples/profiler_example.cc.o -c /opt/local/var/macports/build/highway-ccda4a2a/work/highway-1.3.0/hwy/examples/profiler_example.cc
In file included from /opt/local/var/macports/build/highway-ccda4a2a/work/highway-1.3.0/hwy/examples/profiler_example.cc:19:
/opt/local/var/macports/build/highway-ccda4a2a/work/highway-1.3.0/hwy/contrib/thread_pool/thread_pool.h:283:30: error: static assertion failed
  283 | static_assert(sizeof(Config) == 8, "");
      |               ~~~~~~~~~~~~~~~^~~~
/opt/local/var/macports/build/highway-ccda4a2a/work/highway-1.3.0/hwy/contrib/thread_pool/thread_pool.h:283:30: note: the comparison reduces to '(11 == 8)'
[...]

PR #2680, opened 2025-08-16
thread_pool.h: avoid ambiguous size types in Config struct

barracuda156:
Config must have exact size, but bool is not guaranteed to be 1 byte. Use uint8_t instead, which is always 1 byte. Fixes the build on powerpc-darwin.

Fixes: https://github.com/google/highway/issues/2679


barracuda156:
Bazel CI failure is unrelated:
```
2025/08/16 09:29:59 could not download Bazel: failed to download bazel: failed to download bazel: HTTP GET 
```

jan-wassenberg (approved):
Makes sense, thanks for fixing!

PR #2686, opened 2025-08-22
Use std::sort in auto_tune.h

### LINKED ISSUE #2685: Make `auto_tune` useable without `contrib`
Currently attempting to use `contrib/thread_pool` on it's own (without compiling `contrib/vsort`). The only hangup is here:

https://github.com/google/highway/blob/a6e149cc4e5f239b54e82bb1088c38527e8ab1d0/hwy/auto_tune.h#L28

Does this absolutely need to use `VQSort` or would using `std::sort` be sufficient at the benefit of making the `auto_tune.h` independent of `contrib/vqsort`? I would suspect that the availability of this header within the `hwy` directory would suggest it shouldn't use anything in `contrib`.

Note:
*  It seems that `VQSort` was also added to `profiler.h` in the same PR (#2533) that added it to `auto_tune.h`, but it has since been removed from `profiler.h` (#2657) which now uses `std::sort`.

jcapriot:
Considering `auto_tune.h` is in the `hwy` directory, I've removed the dependence on `VQSort`, allowing `auto_tune.h` to be used without the `contrib` library.

The vector it was sorting was at most 14 elements, so I would guess using `VQSort` is a bit overkill anyways.

Closes #2685

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -847,6 +847,7 @@ endif()  # HWY_TEST_STANDALONE
 set(HWY_TEST_FILES
   hwy/abort_test.cc
   hwy/aligned_allocator_test.cc
+  hwy/auto_tune_test.cc
   hwy/base_test.cc
   hwy/bit_set_test.cc
   hwy/highway_test.cc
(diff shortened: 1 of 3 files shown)


jcapriot:
I've added a configuration option to allow using `auto_tune.h` without the `VQSort` option.

It currently is defaulted to use std:sort only in header only mode, (as using VQSort does require a compiled VQSort), otherwise it will continue to use `VQSort` as before.

Local testing of vectors of 2 and 4 element doubles shows ~<50 ns difference in runtime, so turning it off shouldn't have too much an impact on runtime if so chosen.

jan-wassenberg (approved):
Nice, LGTM :)

Issue #2691, opened 2025-08-28
Preferable fallback when pthread_setname_np is unavailable?


barracuda156:
Turns out that there is a new build error on macOS 10.5.8 with `highway` 1.3.0 due to usage of `pthread_setname_np` here:
https://github.com/google/highway/blob/b01b199e247dff4396a31ba5a25f41d4aa098274/hwy/contrib/thread_pool/thread_pool.h#L62-L65

1.2.0 had the following: https://github.com/google/highway/blob/457c891775a7397bdb0376bb1031e6e027af1c48/hwy/contrib/thread_pool/thread_pool.h#L502-L505

Since that is just an assert, I can simply add a macro to make it `#elif HWY_OS_APPLE && (MAC_OS_X_VERSION_MIN_REQUIRED >= 1060)` (and include the corresponding header), and this fixes the error.
I am not sure, however, whether it is better to add a fallback from 1.2.0.

What would you recommend?

P. S. While it will be nice to have a fix in upstream here, this looks like a third-party library, so perhaps should rather be fixed in its upstream. So at the moment I am perfectly fine with a local fix, I just look for advice on this.

jan-wassenberg:
Thanks for raising.
Please do send a pull request changing that to `#elif HWY_OS_APPLE && (MAC_OS_X_VERSION_MIN_REQUIRED >= 1060)` as you suggest.
It is not just an assert: this is actually setting the thread name for display in debuggers. 1.2 did not support this at all for OSX, so it's fine to skip it if unavailable.

Also, restoring the `#else ` branch sounds like a good idea, this will avoid warnings.

PR #2692, opened 2025-08-28
thread_pool.h: allow build when pthread_setname_np is missing

barracuda156:
Fixes: https://github.com/google/highway/issues/2691


jan-wassenberg (approved):
Thanks!

Issue #2694, opened 2025-08-30
Build error on android NDK r29


bluebutcold:
Building with Android NDK r29-beta3 for x86_64 fails with | 

```[ 56%] Building CXX object CMakeFiles/hwy.dir/hwy/timer.cc.o
In file included from /home/b/build/x/highway/hwy/per_target.cc:27:
In file included from /home/b/build/x/highway/hwy/foreach_target.h:149:
In file included from /home/b/build/x/highway/hwy/per_target.cc:28:
In file included from /home/b/build/x/highway/hwy/highway.h:688:
In file included from /home/b/build/x/highway/hwy/ops/x86_avx3-inl.h:20:
In file included from /home/b/build/x/highway/hwy/ops/x86_512-inl.h:84:
/home/b/build/x/highway/hwy/ops/x86_256-inl.h:6922:20: error: use of undeclared identifier '_mm256_cvtts_pd_epi32'; did you mean '_mm256_cvttspd_epi32'?
 6922 |   return VFromD<D>{_mm256_cvtts_pd_epi32(v.raw)};
      |                    ^~~~~~~~~~~~~~~~~~~~~
      |                    _mm256_cvttspd_epi32
/opt/android-ndk-r29-beta3/toolchains/llvm/prebuilt/linux-x86_64/lib/clang/21/include/avx10_2satcvtdsintrin.h:93:1: note: '_mm256_cvttspd_epi32' declared here
   93 | _mm256_cvttspd_epi32(__m256d __A) {
      | ^
In file included from /home/b/build/x/highway/hwy/per_target.cc:27:
In file included from /home/b/build/x/highway/hwy/foreach_target.h:149:
In file included from /home/b/build/x/highway/hwy/per_target.cc:28:
In file included from /home/b/build/x/highway/hwy/highway.h:688:
In file included from /home/b/build/x/highway/hwy/ops/x86_avx3-inl.h:20:
In file included from /home/b/build/x/highway/hwy/ops/x86_512-inl.h:84:
/home/b/build/x/highway/hwy/ops/x86_256-inl.h:6954:20: error: use of undeclared identifier '_mm256_cvtts_pd_epu32'; did you mean '_mm256_cvttspd_epu32'?
 6954 |   return VFromD<D>{_mm256_cvtts_pd_epu32(v.raw)};
      |                    ^~~~~~~~~~~~~~~~~~~~~
      |                    _mm256_cvttspd_epu32
[...]

PR #2695, opened 2025-09-01
Fix for AVX10_2 target with Android NDK r29-beta3

johnplatts:
Resolves issue #2694.

Android NDK r29-beta3 uses a pre-release version of Clang 21 that does not include the updates made to avx10_2satcvtdsintrin.h and avx10_2_512convertintrin.h that was made in pull request https://github.com/llvm/llvm-project/pull/131592 for Clang 21 and later (and backported to Clang 20.1.x in pull request https://github.com/llvm/llvm-project/pull/135549 that made its way into Clang 20.1.3).


BillyONeal:
Instead of disabling that shouldn't the typo just be fixed? The comment after it says `_mm256_cvttpd_epu32`...

BillyONeal:
Oh I see because it wants saturation...

PR #2701, opened 2025-09-10
Dispatch LSX/LASX at runtime

xry111:
Currently the CMake and meson build systems of highway silently appends -mlasx to compiler flags.  But the Software Development and Build Convention for LoongArch Architectures manual only states the LoongArch desktop and server processors should support LSX, not LASX.  And there is now at least one LoongArch desktop processor model which lacks LASX, namely Loongson 3B6000M.

Silently appending -mlasx to flags is also very unfriendly to the downstream builders like distro maintainers: if they'd like to make a distro specifically optimized with LASX (and give up the compatibility with less-capable processors) they'd make -mlasx the default of their toolchain or packaging system anyway so we don't need to appending it; but if they want to be compatible with less-capable processors like 3B6000M this behavior will just surprise them by triggering some SIGILL at runtime and requiring special build system option to be turned off.

To take the advantage of LASX, it's better to make it dispatched at runtime.  To achieve that we need to utilize a pragma provided by the compiler, similar to how we implement it for x86 and aarch64. Unfortunately, the target pragma support of LoongArch GCC is only recently introduced in GCC 15, and it's still not enough to support highway due to https://gcc.gnu.org/PR121875.  So the runtime dispatch is only implemented for Clang >= 18 (the minimum version with LSX and LASX support) as at now.  The GCC support can be added once the GCC issue is fixed.


hwy/ops/loongarch_lasx-inl.h
```diff
@@ -15,11 +15,24 @@
...
-#include <lasxintrin.h>
-
 #include "hwy/ops/loongarch_lsx-inl.h"
 #include "hwy/ops/shared-inl.h"
 
+#ifndef __loongarch_asx
+// If LASX is to be runtime dispatched (instead of in baseline), we need
+// to enable it *and* define __loongarch_asx or the intrinsic header will
+// fail to compile.
+HWY_PUSH_ATTRIBUTES("lsx,lasx")
```
jan-wassenberg, line 25: Given that you are also setting HWY_TARGET_STR, it seems that we could remove the attribute part here.

Looks like you want it to apply to the lasx header. This can also be done by simply moving the `#include` after HWY_BEFORE_NAMESPACE, and before namespace hwy, right?
(Same comment for LSX below)

hwy/ops/loongarch_lsx-inl.h
```diff
@@ -5928,6 +5942,10 @@ HWY_API V TrailingZeroCount(V v) {
 }  // namespace HWY_NAMESPACE
 }  // namespace hwy
 
+#ifdef HWY_LSX_FEATURE_MACRO_FORCED
```
xry111, line 5945: Oops, this is just a leftover from early development.  I'll remove it.

hwy/ops/loongarch_lasx-inl.h
jan-wassenberg: I see, that makes sense, thanks for explaining. Would you like to add a similar comment explaining this?

xry111:
Any pointer to debug and fix the internal CI failure?

jan-wassenberg:
Thanks for following up. It is an unrelated failure and we are resuming the CI.

Issue #2699, opened 2025-09-08
Scalar BFloat16 to float conversion has very suboptimal codegen on GCC


vekterli:
Hi, and first of all thanks for a great library!

When doing benchmarking of the dot product kernels included in Highway (`contrib/dot/dot-inl.h`) when compiled under GCC 14.2 for x64 targets, I noticed some rather interesting performance characteristics of the BFloat16 overload.

Here's a graph showing input length (in bf16 elements) vs latency in nanoseconds:

<img width="986" height="746" alt="Image" src="https://github.com/user-attachments/assets/df2faccb-6f9a-4aac-807f-3766d9f9c85c" />

The latency increases linearly (and sharply) until the length is at least that of a single AVX2/3 vector register.

Short inputs with fewer elements than that of a full vector register's lane count are deferred to a scalar fallback loop:
```c++
    // Won't be able to do a full vector load without padding => scalar loop.
    if (!kIsAtLeastOneVector && !kIsMultipleOfVector && !kIsPaddedToVector &&
        HWY_UNLIKELY(num_elements < N)) {
      float sum0 = 0.0f;  // Only 2x unroll to avoid excessive code size for..
      float sum1 = 0.0f;  // this unlikely(?) case.
      for (; i + 2 <= num_elements; i += 2) {
        sum0 += F32FromBF16(pa[i + 0]) * F32FromBF16(pb[i + 0]);
        sum1 += F32FromBF16(pa[i + 1]) * F32FromBF16(pb[i + 1]);
      }
      if (i < num_elements) {
        sum1 += F32FromBF16(pa[i]) * F32FromBF16(pb[i]);
      }
      return sum0 + sum1;
    }
```
There is nothing wrong or seemingly suboptimal with this code at all. But on GCC the actual codegen of `F32FromBF16` is not what one would expect (well, certainly not what _I_ would expect).

Example:
```c++
float hwy_bf16_to_float(hwy::bfloat16_t x) noexcept {
    return hwy::F32FromBF16(x);
}
```

Code generated by GCC 14.2 (`-std=c++23 -O3 -march=icelake-server`):
```asm
hwy_bf16_to_float(hwy::bfloat16_t):
        sub     rsp, 8
        vmovd   xmm0, edi
        call    __extendbfsf2
        add     rsp, 8
        ret
```
Note the `call __extendbfsf2` instruction. This is seemingly done to handle signalling NaNs at conversion time.

[...]

PR #2703, opened 2025-09-10
Use `LoadN` for short BF16 dot product inputs instead of scalar conversion

vekterli:
Resolves issue #2699.

This changes the fallback logic for inputs with fewer elements than that of a full vector's lane count to use partial vector loads with implicit zeroing (via `LoadN`) instead of scalar loads and conversions.

I think these should be fairly conservative changes, as the main vector loops remain untouched. An alternative approach would be to generalize the final boundary handling code for < `N` remaining elements to use `LoadN` instead of `FirstN`+"adjusted" `LoadU`, which would remove the need for such an initial special case in the first place. Since I don't have access to enough exotic testing hardware to know how/if this would affect performance on non-x64/arm64 systems, I skipped on doing that. 🙂

It's possible that the use of `ReorderWidenMulAccumulate` should be replaced with explicit `Promote(Upper|Lower)To` and multiplication since the input sums will always be zero, but I am not sure how much it matters in practice. This just mirrors what's in the main loops.

Rationale/background:

Falling back to scalar conversion of `__bf16` to `float` elements will on GCC 13+ end up silently generating a call to the `__extendbfsf2` library function per conversion, rather than just zero-extending to 32 bits and shifting left by 16. This both bloats the generated code and has a substantial runtime cost; on x64 it takes take more time to process 8 BF16 elements with the scalar fallback path than it takes to process 1024 BF16 elements in the vector path..!

Partial vector loads are introduced for the following dot product function overloads:

 * `bfloat16_t` vs. `bfloat16_t` (used `F32FromBF16`)
 * `float` vs. `bfloat16_t` (used `ConvertScalarTo<float>` which transitively calls `F32FromBF16`)

Some before/after graphs from a Sapphire Rapids system:

`bfloat16_t` vs. `bfloat16_t`:

Before:
dot_product_bfloat16_vs_bfloat16

After:
dot_product_bfloat16_vs_bfloat16

`float` vs. `bfloat16_t`:

Before:
dot_product_float_vs_bfloat16

After:
dot_product_float_vs_bfloat16

### DIFF
--- a/hwy/contrib/dot/dot-inl.h
+++ b/hwy/contrib/dot/dot-inl.h
@@ -174,12 +174,30 @@ struct Dot {
         (kAssumptions & kMultipleOfVector) != 0;
     constexpr bool kIsPaddedToVector = (kAssumptions & kPaddedToVector) != 0;
 
-    // Won't be able to do a full vector load without padding => partial load.
+    // Won't be able to do a full vector load without padding. Use a scalar
+    // loop under Clang. GCC has very suboptimal codegen for scalar BF16->float
+    // conversions, so use vector ops with LoadN instead.
+    // TODO: https://github.com/google/highway/pull/2703
     if (!kIsAtLeastOneVector && !kIsMultipleOfVector && !kIsPaddedToVector &&
         HWY_UNLIKELY(num_elements < NF)) {
+#if HWY_COMPILER_GCC_ACTUAL
       const VF a = LoadN(df, pa, num_elements);
       const VF b = PromoteTo(df, LoadN(dbfh, pb, num_elements));
       return ReduceSum(df, Mul(a, b));
+#else
+      // Only 2x unroll to avoid excessive code size.
+      float sum0 = 0.0f;
+      float sum1 = 0.0f;
+      size_t i = 0;
+      for (; i + 2 <= num_elements; i += 2) {
+        sum0 += pa[i + 0] * ConvertScalarTo<float>(pb[i + 0]);
+        sum1 += pa[i + 1] * ConvertScalarTo<float>(pb[i + 1]);
+      }
+      for (; i < num_elements; ++i) {
+        sum1 += pa[i] * ConvertScalarTo<float>(pb[i]);
+      }
+      return sum0 + sum1;
+#endif
     }
 
     // Compiler doesn't make independent sum* accumulators, so unroll manually.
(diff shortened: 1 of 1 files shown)


jan-wassenberg:
Re-importing so we can see what went wrong with internal CI..

jan-wassenberg:
Looks like a compiler crash, sigh. I raised an LLVM issue, linked above.

vekterli:
Changing
```c++
V sum1 = Zero(df32);
V sum0 = ReorderWidenMulAccumulate(df32, a, b, Zero(df32), sum1);
return ReduceSum(df32, Add(sum0, sum1));
```
to
```c++
const auto ab_lo = Mul(PromoteLowerTo(df32, a), PromoteLowerTo(df32, b));
const auto ab_hi = Mul(PromoteUpperTo(df32, a), PromoteUpperTo(df32, b));
return ReduceSum(df32, Add(ab_lo, ab_hi));
```
seems to avoid triggering the Clang crash ([Godbolt](https://godbolt.org/z/oEE6KK4hY)), and should be semantically identical. Performance-wise it also looks very similar to me (after re-running the short input  BF16xBF16 benchmarks from the original comment).

If this sounds good (please let me know if there are preferred alternate approaches), I can update this branch with these changes.

jan-wassenberg:
I agree that's semantically equivalent, and could be a viable workaround, but this should really be a two-instruction sequence : BFDOT+ADDV. Perhaps we can put your workaround inside an `#if HWY_TARGET_IS_SVE && (!HWY_COMPILER_CLANG || HWY_COMPILER_CLANG >= 2300)`?

vekterli:
Hmm it looks like this crashes under `NEON_BF16` as well (https://godbolt.org/z/YxsnhP49q - `armv8.2-a+aes+fp16+bf16+dotprod`). 

The slight irony of the situation is that this is all due to GCC BF16 scalar codegen, so the scalar fallback code under Clang is probably perfectly fine as it is... 🙂

jan-wassenberg:
;) This is BTW the *third* compiler issue we are seeing within a week. It seems LLVM's testing coverage is not yet there.

I'd be fine to land your current change (with ReorderWidenMulAccumulate) `#if HWY_COMPILER_GCC_ACTUAL`, and otherwise the prior scalar fallback, with a TODO comment mentioning this PR.

vekterli:
> I'd be fine to land your current change (with ReorderWidenMulAccumulate) #if HWY_COMPILER_GCC_ACTUAL, and otherwise the prior scalar fallback, with a TODO comment mentioning this PR.

Changes have been made.

Looks like some build failures are happening in the realm of `hwy/contrib/thread_pool`, I presume these are unrelated.

FWIW we created an upstream bug report to GCC a few weeks ago about this issue (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=121853).

If this behavior remains unchanged, scalar conversion will continue to be a bit of a performance mine field. A sledge hammer option (that I'm also pretty sure would remove the need for this particular PR entirely) is to change `F32FromBF16` to pretend `__bf16` is still effectively just a typedef of `short` on GCC :eyes:

```diff
 HWY_API HWY_BF16_CONSTEXPR float F32FromBF16(bfloat16_t bf) {
-#if HWY_HAVE_SCALAR_BF16_OPERATORS
+#if HWY_HAVE_SCALAR_BF16_OPERATORS && !HWY_COMPILER_GCC_ACTUAL
   return static_cast<float>(bf);
 #else
   return BitCastScalar<float>(static_cast<uint32_t>(
```

jan-wassenberg:
Thanks for making the change! Re-imported for internal review, hopefully landing soon.

> Looks like some build failures are happening in the realm of hwy/contrib/thread_pool, I presume these are unrelated.

ACK, triggered by clang warning fixes, thanks for the heads up.

> If this behavior remains unchanged, scalar conversion will continue to be a bit of a performance mine field.

Even if this is fixed tomorrow, we'd still have to live with current compilers for several years. I am open to this suggestion. Want to send a separate PR to change the default value of `HWY_HAVE_SCALAR_BF16_OPERATORS` to `HWY_HAVE_SCALAR_BF16_TYPE && (HWY_COMPILER_GCC_ACTUAL >= 1700)` with a comment mentioning the issue?

vekterli:
Since #2733 is now in, I don't think this PR serves much purpose. Should we just close it alongside #2699?

jan-wassenberg:
I have mixed feelings. If it weren't for the clang bug, it would still be better to have vector code for the remainders. As is, it's indeed regrettable to have it enabled only for GCC. Maybe we can keep this open and un-merged until that's fixed, WDYT?

vekterli:
> Maybe we can keep this open and un-merged until that's fixed, WDYT?

Sounds good to me :+1:

PR #2704, opened 2025-09-12
Enable VQSORT for RISC-V

ixgbe:
Compile environment: Banana Pi BPI-F3 (SpacemiT K1 8-core RISC-V chip)         gcc/g++ version 14.2.0
alias gcc='gcc -march=rv64gcv'
alias g++='g++ -march=rv64gcv'

Configure build with RVV
meson setup builddir \
  -Dcpp_std=c++17 \
  -Dwarning_level=2 \
  -Dbuildtype=release \
  -Dgtest:cpp_eh=default \
  -Drvv=true

Compile
  meson compile -C builddir -j$(nproc) -v

I found that the project compiles successfully on RISC-V if I remove the HWY_ARCH_RISCV check from the preprocessor conditional.


jan-wassenberg:
Thanks for re-checking! Let's be a bit more conservative and only enable it for the config you have tested: HWY_COMPILER_GCC_ACTUAL >= 1400.

PR #2712, opened 2025-09-17
Fix build on powerpc

miladfarca:
Getting the following error after https://github.com/google/highway/pull/2708
```
unused parameter 'num_threads'.
```


miladfarca:
Not a problem :) would you mind landing it as well as I don't have permission to do so.

jan-wassenberg:
:) Landing happens automatically after internal review, which is triggered by adding the label "ready to pull".

miladfarca:
Ah makes sense, thank you.

PR #2719, opened 2025-09-19
meson: remove unused LSX/LASX build options

kleisauke:
These became unused after 10d7ab41fda33b14d9f25cc059e9623c864292a7.


jan-wassenberg (approved):
👍

PR #2733, opened 2025-09-26
Disable scalar BF16 operators on current GCC versions

vekterli:
GCC >= 13 will insert a function call to the `__extendbfsf2` helper function for scalar conversions from `__bf16` to `float`. This is prohibitively expensive, so refrain from using scalar BF16 operators on these compiler versions.

See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=121853

Ref: https://github.com/google/highway/pull/2703 and https://github.com/google/highway/issues/2699


jan-wassenberg (approved):
Nice comment, thanks for adding.

PR #2749, opened 2025-10-05
Fix for Timer::start() result truncation in thread_pool.h

johnplatts:
Updated `ThreadPool::ThreadFunc::operator()()` in thread_pool.h to change the type of `before_run` to `uint64_t` to fix a bug on 32-bit platforms (which triggers a GCC/Clang compiler warning on 32-bit platforms).


jan-wassenberg (approved):
Oops, thanks for fixing this.

Issue #2756, opened 2025-10-08
Compilation failure on OSX Apple Silicon M1


smcallis:
I get an error compiling on my M1 macbook air:

```
../../subprojects/highway/hwy/contrib/sort/sorting_networks-inl.h:28:1: error: unused attribute 'target' in '#pragma clang attribute push' region [-Werror,-Wpragma-clang-attribute]
   28 | HWY_BEFORE_NAMESPACE();
      | ^
../../subprojects/highway/hwy/ops/set_macros-inl.h:827:3: note: expanded from macro 'HWY_BEFORE_NAMESPACE'
  827 |   HWY_PUSH_ATTRIBUTES(HWY_TARGET_STR) \
      |   ^
../../subprojects/highway/hwy/base.h:205:3: note: expanded from macro 'HWY_PUSH_ATTRIBUTES'
  205 |   HWY_PRAGMA(clang attribute push(__attribute__((target(targets_str))), \
      |   ^
../../subprojects/highway/hwy/base.h:141:28: note: expanded from macro 'HWY_PRAGMA'
  141 | #define HWY_PRAGMA(tokens) _Pragma(#tokens)
      |                            ^
<scratch space>:159:38: note: expanded from here
  159 |  clang attribute push(__attribute__((target("+aes"))), apply_to = function)
      |                                      ^
../../subprojects/highway/hwy/contrib/sort/sorting_networks-inl.h:900:1: note: '#pragma clang attribute push' regions ends here
  900 | HWY_AFTER_NAMESPACE();
      | ^
../../subprojects/highway/hwy/ops/set_macros-inl.h:839:3: note: expanded from macro 'HWY_AFTER_NAMESPACE'
  839 |   HWY_POP_ATTRIBUTES          \
      |   ^
../../subprojects/highway/hwy/base.h:207:28: note: expanded from macro 'HWY_POP_ATTRIBUTES'
  207 | #define HWY_POP_ATTRIBUTES HWY_PRAGMA(clang attribute pop)
      |                            ^
../../subprojects/highway/hwy/base.h:141:28: note: expanded from macro 'HWY_PRAGMA'
  141 | #define HWY_PRAGMA(tokens) _Pragma(#tokens)
      |                            ^
<scratch space>:162:8: note: expanded from here
  162 |  clang attribute pop
```

[...]

PR #2758, opened 2025-10-13
Fix for Clang warning if there are no functions defined between HWY_BEFORE_NAMESPACE and HWY_AFTER_NAMESPACE

johnplatts:
Resolves issue #2756.

The following snippet can possibly trigger a `-Wpragma-clang-attribute` warning with Clang:
```
HWY_BEFORE_NAMESPACE();
namespace test {
namespace HWY_NAMESPACE {

}  // namespace HWY_NAMESPACE
}  // namespace test
HWY_AFTER_NAMESPACE();
```

This pull request fixes this issue by ignoring any `-Wpragma-clang-attribute` warnings caused by the following pragma:
```
  HWY_PRAGMA(clang attribute push(__attribute__((target(targets_str))),   \
                                  apply_to = function))
```

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -201,9 +201,15 @@ namespace hwy {
 #define HWY_PUSH_ATTRIBUTES(targets_str)
 #define HWY_POP_ATTRIBUTES
 #elif HWY_COMPILER_CLANG
-#define HWY_PUSH_ATTRIBUTES(targets_str)                                \
-  HWY_PRAGMA(clang attribute push(__attribute__((target(targets_str))), \
-                                  apply_to = function))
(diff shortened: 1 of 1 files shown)


johnplatts:
> Thanks @johnplatts ! I am a bit hesitant to suppress the warning entirely, that could hide some actual problems. The root cause is likely `!VQSORT_ENABLED` which might for example happen in aarch64 asan builds, or when building HWY_SCALAR. That would explain why we are seeing this (only) in sorting_networks-inl.h. As an alternative, how about we, in any `-inl.h` containing `#if VQSORT_ENABLED`, add an `#else` branch that defines an empty function?

Already made the changes in pull request #2759, which has already been merged.

PR #2772, opened 2025-10-21
Fix SVE2_128 MakeMask to avoid duplicate return statement

johnplatts:
Updated MakeMask to avoid duplicate return statement on SVE2_128.

Also refactored the `IsFull(d)` case to use `HWY_IF_CONSTEXPR(IsFull(d)) { return PTrue(d); }` for SVE targets other than SVE2_128.


jan-wassenberg (approved):
LGTM. @MDevereau, please let us know if you have any concerns with this implementation.

PR #2777, opened 2025-10-23
Fix compilation with Clang 19 on s390x

miladfarca:
Currently getting:
```
error: unused parameter 'caller'
```


jan-wassenberg (approved):
Thank you for fixing :)

johnplatts:
@jan-wassenberg What is causing the feedback/copybara error? Does `(void)caller` need to be added inside of the body of the RunWithoutAutotune function and `HWY_MAYBE_UNUSED` removed from the caller parameter to the RunWithoutAutotune function?

jan-wassenberg:
Thanks for following up. We had an unrelated error, I am restarting the CI.

PR #2781, opened 2025-10-24
Re-enable BF16 on x86 in debug builds with Clang 22 or later

johnplatts:
There is a recent fix in LLVM 22 that fixes the bug with `_BFloat16` in debug builds on x86, and this bug was fixed in LLVM pull request https://github.com/llvm/llvm-project/pull/164565 (which has been recently merged into the LLVM trunk).

It is now safe to define `HWY_SSE2_HAVE_SCALAR_BF16_TYPE` to 1 in debug builds with Clang 22 and later on x86 with the fix in https://github.com/llvm/llvm-project/pull/164565.

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -1744,9 +1744,9 @@ HWY_F16_CONSTEXPR inline std::partial_ordering operator<=>(
 // bf16 <-> f32 in convert_test results in 0x2525 for 1.0 instead of 0x3f80.
 // Reported at https://github.com/llvm/llvm-project/issues/151692.
 #ifndef HWY_SSE2_HAVE_SCALAR_BF16_TYPE
-#if HWY_ARCH_X86 && defined(__SSE2__) &&                      \
-    ((HWY_COMPILER_CLANG >= 1700 && !HWY_COMPILER_CLANGCL &&  \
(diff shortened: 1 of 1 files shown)


jan-wassenberg (changes requested):
How about we make the check >= 2300? The issue is that all the various pre-release LLVM versions already call themselves "22", but they might be before or after the fix. Or can we see more with HWY_COMPILER_CLANG3?

stefson:
@johnplatts do you think you can ask upstream llvm for a backport to llvm-21 branch? I could do that as well, but you might be better at explaining them why this would be a good candidate (:

johnplatts:
> How about we make the check >= 2300? The issue is that all the various pre-release LLVM versions already call themselves "22", but they might be before or after the fix. Or can we see more with HWY_COMPILER_CLANG3?

A better alternative is `HWY_COMPILER3_CLANG >= 220101` as Clang 22.1.1 is guaranteed to have the required fixes. I have updated this pull request to check for `HWY_COMPILER3_CLANG >= 220101`.

jan-wassenberg (approved):
Sounds good, thanks for updating.

PR #2782, opened 2025-10-25
Avoid setting or pushing targets attributes when `HWY_COMPILE_ONLY_STATIC` is defined

seiko2plus:
This change is needed to fix GCC ambiguous errors on PPC64 when the `-mcpu=power10` flag is set; This issue appears to be resolved in GCC 13 and later.

### DIFF
--- a/hwy/ops/set_macros-inl.h
+++ b/hwy/ops/set_macros-inl.h
@@ -827,7 +827,7 @@
 
 // Clang <9 requires this be invoked at file scope, before any namespace.
 #undef HWY_BEFORE_NAMESPACE
-#if defined(HWY_TARGET_STR)
+#if defined(HWY_TARGET_STR) && !defined(HWY_COMPILE_ONLY_STATIC)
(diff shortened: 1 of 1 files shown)


seiko2plus:
Loong error seems not related to this change.

seiko2plus:
@jan-wassenberg, This PR is blocking other work. We need a way, in general, to disable defined #pragma with CPU targets to bypass the build errors.

jan-wassenberg:
Thanks for following up. Sometimes import/merge fails. The logs have already been garbage-collected; retrying now.

jan-wassenberg:
Now we have the logs again. The issue is that we have a build failure in an internal project which sets `HWY_COMPILE_ONLY_STATIC` and wants to use avx2, but does not set -mavx2 and instead relies on `HWY_ATTR`.

This seems like legit usage, and incompatible with this change. Is there some other way we can fix the PPC issue? For example disabling PPC prior to the fixed GCC 13?

PR #2800, opened 2025-11-19
Made enhancements to PopulationCount

johnplatts:
Updated U8 PopulationCount on RVV to use table lookup using TableLookupLanes for all LMUL values as `v` will be resized to a U8 vector with a LMUL of 1 if the LMUL of `v` is less than 1 to allow the entire lookup table to be used.

Also removed the U8 PopulationCount implementation in rvv-inl.h for vectors that have a LMUL <= 1 as the updated implementation of PopulationCount in this pull request can deal with LMUL <= 1 U8 vectors.

Also updated PPC/Z14/Z15 PopulationCount to use the vec_popcnt intrinsic as PPC8/Z14 have vector population count instructions.

Also updated PopulationCount on SSE2 to avoid table lookup as SSE2 lacks table lookup instructions.

### DIFF
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -4541,43 +4541,46 @@ HWY_API V CLMulUpper(V a, V b) {
 #define HWY_NATIVE_POPCNT
 #endif
 
-// This overload requires vectors to be at least 16 bytes, which is the case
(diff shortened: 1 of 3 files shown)


jan-wassenberg (approved):
Nice simplification 👍

PR #2808, opened 2025-12-04
Updates to GitHub build/test workflows

johnplatts:
Updated all build/test workflows that use CMake or Meson to use C++17 mode or newer.

Also added GCC 14 and Clang 19 to GitHub build/test workflow.

Also added `sudo apt-get update` to GitHub build/test workflows that install dependencies using `sudo apt-get install` to fix errors that sometimes occur with installing dependencies (which is described over at https://github.com/actions/runner/issues/3720).


jan-wassenberg (approved):
Thanks for updating. I see that the build/test and Foreign Arch actions are about as fast, hence OK to add new configs.

jan-wassenberg:
Thanks for updating. Internal CI failed due to an unrelated error. Retrying..

PR #2821, opened 2025-12-11
Use svbfdot for ReorderWidenMulAccumulate

MDevereau:
A comment claims there is a difference between SVE and NEON's rounding behaviour for BFDOT. According to the arm developer documentation, both instruction sets have the same rounding to odd behaviour by default when FEAT_EBF16 is not implemented or FPCR.EBF is 0:

https://developer.arm.com/documentation/ddi0602/2025-09/SIMD-FP-Instructions/BFDOT--vector---BFloat16-dot-product-to-single-precision--vector-- https://developer.arm.com/documentation/ddi0602/2025-09/SVE-Instructions/BFDOT--vectors---BFloat16-dot-product-to-single-precision-

### DIFF
--- a/hwy/ops/arm_neon-inl.h
+++ b/hwy/ops/arm_neon-inl.h
@@ -7352,6 +7352,9 @@ static HWY_INLINE bfloat16x8_t BitCastToRawNeonBF16(bfloat16x8_t raw) {
 // The uint16x4_t or uint16x8_t vector neets to be bitcasted to a bfloat16x4_t
 // or a bfloat16x8_t vector for the vbfdot_f32 and vbfdotq_f32 intrinsics if
 // HWY_NEON_HAVE_F32_TO_BF16C && !HWY_NEON_HAVE_BFLOAT16 is true
+
+// NOTE: vbfdot uses round to odd unless the additional FEAT_EBF16 feature is
+// available and enabled.
 static HWY_INLINE bfloat16x4_t BitCastToRawNeonBF16(uint16x4_t raw) {
   return vreinterpret_bf16_u16(raw);
 }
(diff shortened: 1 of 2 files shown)


jan-wassenberg:
Thanks for pointing this out. I agree the comment is incorrect and that NEON matches SVE. It still seems useful to document that both of them behave differently than perhaps expected, unless FEAT_EBF16 is supported and on. Can we add a quick comment to quick_reference saying that this might round to odd on Arm? (BTW: I don't have a good picture of how widespread EBF16 is expected to be?)

johnplatts:
@jan-wassenberg and @MDevereau - None of the hardware AArch64 CPU's listed at https://gpages.juszkiewicz.com.pl/arm-socs-table/arm-socs.html support the AArch64 FEAT_EBF16 feature (and this table only lists QEMU as supporting the FEAT_EBF16 feature, but QEMU emulates an AArch64 CPU)

Most of the Google Highway ops that can generate inexact floating point results (including ReorderWidenMulAccumulate) also assume that the floating point mode of the CPU is in the default round-to-nearest mode.

`ConvertScalarTo<float>(a) * ConvertScalarTo<float>(b)` is guaranteed to produce exact results if `a` and `b` are both BF16 values unless `0 < |a * b| < 2^-134` or `|a * b| >= 2^128`.

The incorrect rounding in svbfdot_f32 usually only occurs if an addition is inexact (which is possible in the case where `a[2*i] * b[2*i]`, `a[2*i+1] * b[2*i+1]`, and `sum0[i]` are all normal F32 values but the sum is inexact), `0 < |a[2*i] * b[2*i]| < 2^-134`, or `0 < |a[2*i+1] * b[2*i+1]| < 2^-134` (the latter two cases only occurring if the magnitude of the exact product is nonzero and less than 2^-134 (the smallest positive denormal F32 value that has 16 bits of precision)).

PR #2828, opened 2025-12-18
Move i8mm AArch64 intrinsics behind compiler options

MDevereau:
+bf16 and +i8mm have no relationship, but hardware that does not have access to +i8mm will be barred from using the target HWY_NEON_BF16.

This patch moves vusdotq and vusdot to be gated behind compile time options, similar to how svbfmlalb and svbfmlalt are gated behind +bf16 for the target HWY_SVE2 with HWY_SVE_HAVE_BF16_FEATURE.

### DIFF
--- a/hwy/detect_targets.h
+++ b/hwy/detect_targets.h
@@ -1,4 +1,5 @@
 // Copyright 2021 Google LLC
+// Copyright 2025 Arm Limited and/or its affiliates <open-source-office@arm.com>
 // SPDX-License-Identifier: Apache-2.0
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


MDevereau:
> Makes sense, i8mm is less widely supported. Is there any place we could document that some benefit can be had from i8mm by building with the required flag?

I was just about to push a commit with:
`// Note: svusdot requires compilation with the flag "+i8mm"`
and
`// Note: vusdot and vusdotq require compilation with the flag "+i8mm"`
Where the intrinsics are, but it got merged. Do you think these are important enough to warrant another PR?

jan-wassenberg:
Thanks, no worries, I can add such a comment :)

PR #2837, opened 2026-01-03
Dont include CTest if HWY_ENABLE_TESTS is off.

### LINKED ISSUE #2833: CTest included when tests are disabled
Hey,

When HWY_ENABLE_TESTS is OFF: [CTest is still included](https://github.com/google/highway/blob/master/CMakeLists.txt#L759).

This isn't a big problem but it does put a bunch of projects into my solution that I don't need.

One hacky workaround is to set `CTEST_TARGETS_ADDED 1` before making highway available.

xoorath:
Introduces [the change proposed](https://github.com/google/highway/issues/2833#issuecomment-3705075551) by jan-wassenberg

This change avoids including CTest or creating unused testing targets when HWY_ENABLE_TESTS is off.

Fixes #2833

## Testing done:

1. Ran `cmake -B my_build -S . -G "Visual Studio 17 2022"`
2. Opened the visual studio solution
3. ✅ All test projects were in the solution (including the CTest generated Continuous, Experimental, etc.)
4. Ran `cmake -B my_build -S . -DHWY_ENABLE_TESTS=OFF  -G "Visual Studio 17 2022"`
5. Opened the visual studio solution
6. ✅ No test projects were in the solution (including the CTest generated Continuous, Experimental, etc.)

Screenshot 2026-01-03 142846
Screenshot 2026-01-03 142925

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -756,9 +756,10 @@ endif()  # HWY_ENABLE_CONTRIB
 endif()  # HWY_ENABLE_EXAMPLES
 # -------------------------------------------------------- Tests
 
+if(HWY_ENABLE_TESTS)
 include(CTest)
 
-if(BUILD_TESTING AND HWY_ENABLE_TESTS)
+if(BUILD_TESTING)
 enable_testing()
 include(GoogleTest)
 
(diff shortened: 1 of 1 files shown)


xoorath:
I've now signed the CLA for my account. That check can be re-run.

jan-wassenberg:
Thanks for making the change! I think the CLA check is still failing because the email address is x*​@noreply.git.xo.codes, but the check requires that this match the CLA email. Can we change that?

xoorath:
Alright, the CLA should be sorted now. Had to re-author the commit with a real email unfortunately, but c'est la vie.

jan-wassenberg (approved):
Thank you for re-doing the commit :)

PR #2840, opened 2026-01-07
Made enhancements to default MulByPow2 implementation in generic_ops-inl.h

johnplatts:
Updated the default MulByPow2 implementation in generic_ops-inl.h to compute `factor1[i]`, `factor2[i]`, and `factor3[i]` in a more efficient manner.

`v[i] * factor1[i] * factor2[i] * factor3[i]` turns out to be correctly rounded if `(exp1[i] >= 0 && exp2[i] >= 0) || (exp3[i] == 1 - kExpBias && kExpBias >= kNumOfMantBits + 3)` is true. `kExpBias >= kNumOfMantBits + 3` is known to be true for the F16, F32, and F64 floating point formats.


jan-wassenberg (approved):
Impressive. Looks like three ops less, nice!

Issue #2831, opened 2025-12-18
`hwy::DispatchedTarget` always returns HWY_EMU128 on WASM/Emscripten


LanderlYoung:
The reason is the highway library isn't compiled with `-msimd128`.

My project is compiled with Emscripten with `-msimd128` flag.
I'm using CMake and `FetchContent` to download source of highway, and `FetchContent` used `add_subdirectory` to integrate highway in to my build.
And the `HWY_DYNAMIC_DISPATCH` macro is doing great by selecting `HWY_WASM` target.

Current viable solution is using `include` instead of `add_subdirectory` in cmake, and add `-msimd128` flag to highway manually.

Or, I'd like to discuss:
1. Is it possible for WASM to auto detect and enable WASM SIMD by pragma like avx targets? (without manually adding compiler flags).
2. Maybe landing https://github.com/google/highway/issues/2114 is also a great idea. (I'd like to try that, but it may produce a big change list)

jan-wassenberg:
Ah, makes sense. From a quick test on Godbolt Clang webassembly, adding `__attribute__((target("simd128")))` seems to be accepted, whereas other strings do raise a warning.
You could try, in detect_targets.h, adding something like HWY_HAVE_RUNTIME_DISPATCH_APPLE for WASM, checking HWY_ARCH_WASM and HWY_COMPILER_CLANG (perhaps a recent version), then add that as another precondition for HWY_HAVE_RUNTIME_DISPATCH. This might then work out of the box. If it does, we'd welcome a PR (to the dev branch) :)

LanderlYoung:
Thanks for the reply, I'll give it a try and maybe create a PR latter😜

LanderlYoung:
Turns out highway has already defined the compiler attribute in `set_macros-inl.h`
```cpp
#define HWY_NAMESPACE N_WASM

#define HWY_TARGET_STR "simd128"
```

## The Reason
The reason is  `detect_targets.h` also checks for `__wasm_simd128__` (which relays on `-msimd128` compiler flag)

```cpp
#if HWY_ARCH_WASM && defined(__wasm_simd128__)
```

After remove the check, everthing works fine. And my project don't need to pass `-msimd128` explicitly.

## The Catch
For users who disable target by `HWY_DISABLED_TARGETS`, the problem still exist.

> Some use may want to disable simd feature for some reason (like browser compatility), but the `per_target.cc` will still has a N_WASM implementation generated, luckly the generated code doesn't contain any simd instruction.

[...]

PR #2851, opened 2026-01-22
Optimize auto-detection of `HWY_ARCH_WASM` target

LanderlYoung:
1. removed check for `__wasm_simd128__` macro
2. add documentation in `quick_reference.md` at `Targets` and `Detecting supported targets` section.

As discussed in #2831


jan-wassenberg (approved):
Nice, thanks for making the change and adding the thorough comments.

jan-wassenberg:
Unfortunately, this breaks internal users and has been rolled back. We're coordinating internally whether we can add -msimd128 to the affected builds.

jan-wassenberg:
(Quick update, some progress but more projects affected than initially believed. More work required..)

PR #2864, opened 2026-02-19
Fix null placement-new UB and exception leak in MakeUniqueAligned 

rhythmcache:
fix null deref and exception leak in MakeUniqueAligned functions if AllocateAlignedBytes returns null we were doing placement new on nullptr which is UB. also if T's constructor throws after allocation succeeds the memory was leaking since unique_ptr hadn't taken ownership yet. added null checks and try/catch with proper cleanup in both the alloc and no-alloc variants.

### DIFF
--- a/hwy/aligned_allocator.h
+++ b/hwy/aligned_allocator.h
@@ -122,7 +122,7 @@ AlignedUniquePtr<T> MakeUniqueAlignedWithAlloc(AllocPtr alloc, FreePtr free,
   if (HWY_UNLIKELY(ptr == nullptr)) {
     return AlignedUniquePtr<T>(nullptr, AlignedDeleter(free, opaque));
   }
-#if HWY_EXCEPTIONS_ENABLED
+#ifdef HWY_EXCEPTIONS_ENABLED
   try {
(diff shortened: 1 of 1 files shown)


hwy/aligned_allocator.h
```diff
@@ -119,17 +119,43 @@ template <typename T, typename... Args>
 AlignedUniquePtr<T> MakeUniqueAlignedWithAlloc(AllocPtr alloc, FreePtr free,
                                                void* opaque, Args&&... args) {
   T* ptr = static_cast<T*>(AllocateAlignedBytes(sizeof(T), alloc, opaque));
+  if (HWY_UNLIKELY(ptr == nullptr)) {
+    return AlignedUniquePtr<T>(nullptr, AlignedDeleter(free, opaque));
+  }
+#if HWY_EXCEPTIONS_ENABLED
```
jan-wassenberg: This makes sense, thanks for adding it. I like the idea of an opt-in macro, let's just check it via `#ifdef`.

rhythmcache: changed to #ifdef in both functions.

PR #2872, opened 2026-02-22
Fix build on ppc64 and s390x

miladfarca:
Currently getting the following compilation errors in a few spots using Clang 19 after this commit: https://github.com/google/highway/commit/41d3a6b
```
error: implicit conversion loses floating-point precision: 'double' to 'const T'
```


Nikhil0250:
Hi @miladfarca 
Thanks for catching this. Can you update your pr to do the explicit type casting (static_cast<T>) for all the places it is required in fast_math-inl.h? 
I see you have missed some places (eg. all the coefficient arrays need to be type casted explicitly )

miladfarca:
Hi @Nikhil0250 
I have updated the remaining instances, please let me know if anything is missing, thanks.

jan-wassenberg (approved):
Thanks Milad, LGTM :)

PR #2878, opened 2026-02-24
Fix build on ppc64/s390x

miladfarca:
Currently failing with Clang 19:
```
Implicit conversion loses floating-point precision: 'double' to 'float'.
```


Nikhil0250:
Thanks @miladfarca for catching this again

miladfarca:
@Nikhil0250 no issues at all!

PR #2966, opened 2026-04-04
Add Count/CountIf algos

RaviTriv:
Implementing Count/CountIf from [wishlist](https://github.com/google/highway/blob/master/g3doc/op_wishlist.md?plain=1#L48)

- Thoughts on the mask accumulation technique described in the [SIMD Masking Counting Values section](https://en.algorithmica.org/hpc/simd/masking/#counting-values).
  - Mainly wondering about the magic required to ensure this works with all types.
    - safely handling `uint8_t` ? Limit to `x` iterations, accumulate, continue?
    - Dealing with float and double.
    - Apply this optimization only for specific types ?
  - Think if would look something like this ?
  ```C++
  const RebindToSigned<D> di;
  const size_t N = Lanes(d);
  const Vec<D> broadcasted = Set(d, value);

  size_t total = 0;
  size_t i = 0;

  if (count >= 4 * N) {
    Vec<decltype(di)> acc0 = Zero(di);
    Vec<decltype(di)> acc1 = Zero(di);
    Vec<decltype(di)> acc2 = Zero(di);
    Vec<decltype(di)> acc3 = Zero(di);

    for (; i <= count - 4 * N; i += 4 * N) {
      acc0 = Add(acc0, BitCast(di, VecFromMask(d, Eq(broadcasted, LoadU(d, in + i)))));
      acc1 = Add(acc1, BitCast(di, VecFromMask(d, Eq(broadcasted, LoadU(d, in + i + N)))));
      acc2 = Add(acc2, BitCast(di, VecFromMask(d, Eq(broadcasted, LoadU(d, in + i + 2 * N)))));
      acc3 = Add(acc3, BitCast(di, VecFromMask(d, Eq(broadcasted, LoadU(d, in + i + 3 * N)))));
    }

    acc0 = Add(Add(acc0, acc1), Add(acc2, acc3));
    total += static_cast<size_t>(-1 * ReduceSum(di, acc0));
    // Handle floats, double and small types edge cases!!
  }

  // handle remaining
  return total;
  ```

### DIFF
--- a/hwy/contrib/algo/count-inl.h
+++ b/hwy/contrib/algo/count-inl.h
@@ -48,7 +48,7 @@ size_t Count(D d, T value, const T* HWY_RESTRICT in, size_t count) {
     const Repartition<int32_t, D> di32;
     auto wide_sum = Zero(di32);
 
-    if (count >= 4 * N) {
+    if (count >= 4 * N && N >= 4) {
       while (i <= count - 4 * N) {
         VI acc0 = Zero(di);
         VI acc1 = Zero(di);
@@ -106,7 +106,7 @@ size_t Count(D d, T value, const T* HWY_RESTRICT in, size_t count) {
     const Repartition<int32_t, D> di32;
     auto wide_sum = Zero(di32);
 
-    if (count >= 4 * N) {
+    if (count >= 4 * N && N >= 2) {
       while (i <= count - 4 * N) {
         VI acc0 = Zero(di);
         VI acc1 = Zero(di);
@@ -230,7 +230,7 @@ size_t CountIf(D d, const T* HWY_RESTRICT in, size_t count, const Func& func) {
     const Repartition<int32_t, D> di32;
     auto wide_sum = Zero(di32);
 
-    if (count >= 4 * N) {
+    if (count >= 4 * N && N >= 4) {
       while (i <= count - 4 * N) {
         VI acc0 = Zero(di);
         VI acc1 = Zero(di);
@@ -282,7 +282,7 @@ size_t CountIf(D d, const T* HWY_RESTRICT in, size_t count, const Func& func) {
     const Repartition<int32_t, D> di32;
     auto wide_sum = Zero(di32);
 
-    if (count >= 4 * N) {
+    if (count >= 4 * N && N >= 2) {
       while (i <= count - 4 * N) {
         VI acc0 = Zero(di);
         VI acc1 = Zero(di);


hwy.gni
```diff
@@ -45,6 +45,7 @@ hwy_sources = [
 
 hwy_contrib_public = [
   "$_hwy/contrib/algo/copy-inl.h",
+  "$_hwy/contrib/algo/count-inl.h",
```
kleisauke, line 48: This header should also be added to the CMake and Meson build files, see PR #2963 for reference.

hwy/contrib/algo/count-inl.h
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  if (count >= N) {
+    for (; i <= count - N; i += N) {
+      total += CountTrue(d, Eq(broadcasted, LoadU(d, in + i)));
+    }
+  }
+
+  if (i != count) {
+#if HWY_MEM_OPS_MIGHT_FAULT
```
jan-wassenberg: This approach is obsolete, we now have a better alternative: LoadN is always safe, and as efficient as MaskedLoad when masking is reliable.

hwy.gni
RaviTriv: Oops, sorry i missed this. Thanks for letting me know. Made this change locally, will push the change and update in this thread when I make some other changes. Thanks for your patience!

RaviTriv:
> SatWidenMulPairwiseAdd

Thanks for taking time to give feedback!

The reference and your thoughts are incredibly useful. 

This is roughly what I have in mind, will probably take me a few tries to get this right 😅

```
if(8 bit){
    while (less than count){
        while(less than 128){
            if(HWY_NATIVE_MASK){
                if(register >= 32){
                    // 4 accumulators using MaskedAddOr
                }else{
                    // 2 accumulators using MaskedAddOr
                }
            } else {
                if(register >= 32){
                    // 4 accumulators using VecFromMask
                }else{
                    // 2 accumulators using VecFromMask
                }
            }
        }
        // SatWidenMulPairwiseAdd
    }
    // ReduceSum
}else if(16 bit){
    while(less than count){
        while(less than 32k){
            // similar to above
        }
        // SatWidenMulPairwiseAdd
    }
    // ReduceSum
}else{
    // similar to above no cap
}
```

Will keep you posted when i push code changes.

Appreciate your patience :)

jan-wassenberg:
Glad to hear it's useful.

Note that because we are in hwy/contrib/, it's fine to use if constexpr, which is cleaner/maybe faster to compile than the SFINAE approaches we previously used.

I think your register usage is low enough that we can always use 4 accumulators, even SSE4's 8 regs should be enough for that.
Note that the 8-bit path will want to do the widening dot product twice as mentioned.

hwy.gni
RaviTriv: Pushed change for this.

RaviTriv:
> Glad to hear it's useful.
> 
> Note that because we are in hwy/contrib/, it's fine to use if constexpr, which is cleaner/maybe faster to compile than the SFINAE approaches we previously used.
> 
> I think your register usage is low enough that we can always use 4 accumulators, even SSE4's 8 regs should be enough for that. Note that the 8-bit path will want to do the widening dot product twice as mentioned.

Thanks!
Attempted to refactor `Count`, will update `CountIf` once I get `Count` right :)

**Edit:** Investigating build/test fails

hwy/contrib/algo/count-inl.h
```diff
@@ -0,0 +1,267 @@
...
+      if (count >= 4 * N) {
+        while (i <= count - 4 * N) {
+          VI acc0 = Zero(di);
+          VI acc1 = Zero(di);
+          VI acc2 = Zero(di);
+          VI acc3 = Zero(di);
+          const size_t cap = HWY_MIN(i + 128 * 4 * N, count);
+
+#ifdef HWY_NATIVE_MASK
+          const auto one_i = Set(di, TFromD<decltype(di)>(1));
```
jan-wassenberg: Naming is a bit unusual. Suggest k1, or just one? I know we have some other typed 1 constants below, but we could also BitCast them from this.

hwy/contrib/algo/count-inl.h
```diff
@@ -0,0 +1,267 @@
...
+
+      if (count >= 4 * N) {
+        while (i <= count - 4 * N) {
+          VI acc0 = Zero(di);
+          VI acc1 = Zero(di);
+          VI acc2 = Zero(di);
+          VI acc3 = Zero(di);
+          const size_t cap = HWY_MIN(i + 128 * 4 * N, count);
+
+#ifdef HWY_NATIVE_MASK
```
jan-wassenberg: Can also be if constexpr, this is nicer for the IDE (allows code folding).

hwy/contrib/algo/count-inl.h
```diff
@@ -0,0 +1,267 @@
...
+            acc3 = MaskedAddOr(
+                acc3, RebindMask(di, Eq(broadcasted, LoadU(d, in + i + 3 * N))),
+                acc3, one_i);
+          }
+          acc0 = Add(Add(acc0, acc1), Add(acc2, acc3));
+#else
+          for (; i <= cap - 4 * N; i += 4 * N) {
+            acc0 = Add(
+                acc0,
+                BitCast(di, VecFromMask(d, Eq(broadcasted, LoadU(d, in + i)))));
```
jan-wassenberg: Formatting is a bit unfortunate here. Can consider using M = Mask<D>; Then m0 = Eq(broadcasted, ...); or maybe also pull VecFromMask into that line, then the Add+BitCast would definitely fit into one line.

hwy/contrib/algo/count-inl.h
```diff
@@ -0,0 +1,267 @@
...
+                acc1, RebindMask(di, Eq(broadcasted, LoadU(d, in + i + N))),
+                acc1, one_i);
+            acc2 = MaskedAddOr(
+                acc2, RebindMask(di, Eq(broadcasted, LoadU(d, in + i + 2 * N))),
+                acc2, one_i);
+            acc3 = MaskedAddOr(
+                acc3, RebindMask(di, Eq(broadcasted, LoadU(d, in + i + 3 * N))),
+                acc3, one_i);
+          }
+          acc0 = Add(Add(acc0, acc1), Add(acc2, acc3));
```
jan-wassenberg: I think this can already overflow. We can either reduce the number of iterations 4x, which is a bit of a waste, or first do the widening add before the accumulator tree here?

hwy/contrib/algo/count-inl.h
```diff
@@ -0,0 +1,267 @@
...
+                                d, Eq(broadcasted, LoadU(d, in + i + 2 * N)))));
+            acc3 = Add(
+                acc3,
+                BitCast(di, VecFromMask(
+                                d, Eq(broadcasted, LoadU(d, in + i + 3 * N)))));
+          }
+          acc0 = Neg(Add(Add(acc0, acc1), Add(acc2, acc3)));
+#endif
+          const auto acc_u8 = BitCast(RebindToUnsigned<decltype(di)>(), acc0);
+          const auto one_i8 = Set(di, TFromD<decltype(di)>(1));
```
jan-wassenberg: We can eliminate the Neg by using -1 as the multiplier here :D

hwy/contrib/algo/count-inl.h
```diff
@@ -0,0 +1,267 @@
...
+  const size_t N = Lanes(d);
+  const Vec<D> broadcasted = Set(d, value);
+  const RebindToSigned<D> di;
+  using VI = Vec<decltype(di)>;
+
+  size_t total = 0;
+  size_t i = 0;
+
+  if constexpr (sizeof(T) == 1) {
+    if constexpr (HWY_MAX_LANES_D(D) >= 4) {
```
jan-wassenberg: Quite deep if-nesting here. Not sure we want to handle this one. We could simply DASSERT that d.MaxBytes() >= 4?

hwy/contrib/algo/count-inl.h
```diff
@@ -0,0 +1,267 @@
...
+                BitCast(di, VecFromMask(
+                                d, Eq(broadcasted, LoadU(d, in + i + 2 * N)))));
+            acc3 = Add(
+                acc3,
+                BitCast(di, VecFromMask(
+                                d, Eq(broadcasted, LoadU(d, in + i + 3 * N)))));
+          }
+          acc0 = Neg(Add(Add(acc0, acc1), Add(acc2, acc3)));
+#endif
+          const auto one_i16 = Set(di, TFromD<decltype(di)>(1));
```
jan-wassenberg: Same here, -1 multiplier avoids Neg.

hwy/contrib/algo/count-inl.h
```diff
@@ -0,0 +1,267 @@
...
+          wide_sum =
+              SatWidenMulPairwiseAccumulate(di32, acc0, one_i16, wide_sum);
+        }
+      }
+      total += static_cast<size_t>(ReduceSum(di32, wide_sum));
+    }
+  } else {
```
jan-wassenberg, line 155: Comment: lane type is wide enough that no widening is necessary

hwy/contrib/algo/count-inl.h
RaviTriv: hmm, compiler was not happy without, `error: no matching function for call to 'SatWidenMulPairwiseAdd'`

jan-wassenberg: Ah, I see. Might help to add a comment. I think SatWidenMulPairwiseAdd does require at least two bytes of inputs to make sense, otherwise there is no pairwise sum.

hwy/contrib/algo/count-inl.h
```diff
@@ -0,0 +1,256 @@
...
+                VecFromMask(d, Eq(broadcasted, LoadU(d, in + i + 2 * N)));
+            const auto v3 =
+                VecFromMask(d, Eq(broadcasted, LoadU(d, in + i + 3 * N)));
+            acc0 = Add(acc0, BitCast(di, v0));
+            acc1 = Add(acc1, BitCast(di, v1));
+            acc2 = Add(acc2, BitCast(di, v2));
+            acc3 = Add(acc3, BitCast(di, v3));
+          }
+        }
+        const auto mul = HWY_NATIVE_MASK ? k1 : Neg(k1);
```
jan-wassenberg: Compilers might see through and elide the Neg, but to be sure, we can replace Neg(k1) with Set(di, 1). If the concern is casting the 1 to the right TI, we can have a using TI = TFromD<decltype(di)> and then write Set(di, TI{1}) here and in the definition of k1.

hwy/contrib/algo/count-inl.h
jan-wassenberg: Oh, we're doing two pairwise sums, hence the original 2*2 was correct.

hwy/contrib/algo/count-inl.h
```diff
@@ -229,20 +367,10 @@ size_t CountIf(D d, const T* HWY_RESTRICT in, size_t count, const Func& func) {
   }
 
   if (i != count) {
```
jan-wassenberg, line 369: I think here also we require a whole-vector loop, in addition to the remainder handling below, because we may have up to 3 remaining vectors.
Otherwise LGTM :)

RaviTriv: cool!

to clarify, i should look to update from

```c++
  if (count >= N) {
    for (; i <= count - N; i += N) {
      total += CountTrue(d, func(d, LoadU(d, in + i)));
    }
  }
  
  if(i != count)
```

to something like 

```c++
if(count >= N){
  // masked accumulation technique
} 

if(i != count) 
 ``` 

for `Count` and `CountIf`?

jan-wassenberg: Oh, the commit I was looking at did not seem to have the above loop - perhaps it was already present in the base version. I think it's already good as-is :)

jan-wassenberg:
I see we have GCC build failures due to unused k1 variables at lines 157, 327. Please remove or fold into the blocks where they are used.

jan-wassenberg:
Nice, builds are green :) Just one remaining bug/failure in GCC with SVE, where HWY_NATIVE_MASK=1. Are you able to test on an Arm machine to see what's failing?

RaviTriv:
> Nice, builds are green :) Just one remaining bug/failure in GCC with SVE, where HWY_NATIVE_MASK=1. Are you able to test on an Arm machine to see what's failing?

haha working on trying get access to one right now :) 

gonna try to debug test this further, will keep you posted when I make progress / run into issues. thanks!!

RaviTriv:
> > Nice, builds are green :) Just one remaining bug/failure in GCC with SVE, where HWY_NATIVE_MASK=1. Are you able to test on an Arm machine to see what's failing?
> 
> haha working on trying get access to one right now :)
> 
> gonna try to debug test this further, will keep you posted when I make progress / run into issues. thanks!!

image

**Error:** `i8x2 count 27 misalign 0: Count(-4.000000) expected 3 got 1`

It seems like were trying to do widening when we didn't have enough lanes on SVE due to variable size vector length. Added a run time guard to help protect against this.

hopefully this will do the trick

**Edit:** Think a rebase once PR 2977 lands should do it.

jan-wassenberg:
Oh nice, that makes sense. Our tests are quite strict in the sense that they include even tiny d. This edge case wouldn't often be seen, but still good to handle it.
LGTM!

PR #2981, opened 2026-04-14
Add information about NumKong SIMD accelerated math library

bkmgit:


jan-wassenberg (approved):
Golly, that's a lot of code/per-CPU variants they have there :)

PR #2987, opened 2026-04-15
add loongarch64 support for BUILD

zhangwenlong8911:


jan-wassenberg:
Thanks for adding. Happy to land this but we are currently failing builds.

Debugging: we already have bazel_dep(name = "platforms", version = "1.0.0") in MODULE.bazel which should be new enough, and the bazel version is 9.0.2. USE_BAZEL_VERSION is not set, and there is no `.bazelversion`.

It's probably the WORKSPACE. In that file, please replace the old platforms with this:

```
http_archive(
    name = "platforms",
    sha256 = "3384eb1c30762704fbe38e440204e114154086c8fc8a8c2e3e28441028c019a8",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/platforms/releases/download/1.0.0/platforms-1.0.0.tar.gz",
        "https://github.com/bazelbuild/platforms/releases/download/1.0.0/platforms-1.0.0.tar.gz",
    ],
)
```

jan-wassenberg:
Thanks for updating. Still seeing a build error. The cause might be that maybe() skips the archive if the module is already present, which it is: Bazel already has an older platforms built in.

Please try changing the WORKSPACE block to remove the `maybe` like so:
```
http_archive(
    name = "platforms",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/platforms/releases/download/1.0.0/platforms-1.0.0.tar.gz",
        "https://github.com/bazelbuild/platforms/releases/download/1.0.0/platforms-1.0.0.tar.gz",
    ],
    sha256 = "3384eb1c30762704fbe38e440204e114154086c8fc8a8c2e3e28441028c019a8",
)
```

zhangwenlong8911:
> Thanks for adding. Happy to land this but we are currently failing builds.
> 
> Debugging: we already have bazel_dep(name = "platforms", version = "1.0.0") in MODULE.bazel which should be new enough, and the bazel version is 9.0.2. USE_BAZEL_VERSION is not set, and there is no `.bazelversion`.
> 
> It's probably the WORKSPACE. In that file, please replace the old platforms with this:
> 
> ```
> http_archive(
>     name = "platforms",
>     sha256 = "3384eb1c30762704fbe38e440204e114154086c8fc8a8c2e3e28441028c019a8",
>     urls = [
>         "https://mirror.bazel.build/github.com/bazelbuild/platforms/releases/download/1.0.0/platforms-1.0.0.tar.gz",
>         "https://github.com/bazelbuild/platforms/releases/download/1.0.0/platforms-1.0.0.tar.gz",
>     ],
> )
> ```

ok

jan-wassenberg:
Thanks for also trying that. I've misunderstood the problem. It is not that we are using a too old platforms module - loongarch is not yet 'supported' (added to Bazel) even in v1.0 of its platforms.
I've pinged https://github.com/bazelbuild/platforms/issues/107 to support adding it.

(Out of curiosity, what prompted you to add @platforms//cpu:loongarch64 - are you using a locally patched version of Bazel?)

To move forward, I believe we can add the following to BUILD:

```
# loongarch64 is not yet in @platforms//cpu, define it locally.
constraint_value(
    name = "cpu_loongarch64",
    constraint_setting = "@platforms//cpu:cpu",
)

config_setting(
    name = "loongarch64",
    constraint_values = [":cpu_loongarch64"],
)
```
Then use :loongarch64 in the select() calls instead of @platforms//cpu:loongarch64.

zhangwenlong8911:
> (Out of curiosity, what prompted you to add @platforms//cpu:loongarch64 - are you using a locally patched version of Bazel?)

yes, I am using a locally patched version of Bazel from https://github.com/Loongson-Cloud-Community/bazel/tree/7.5.0

jan-wassenberg:
I see, thanks. We do not have that on the Github actions CI, hence we could only merge after it's added to bazel platforms (sounds like there is some movement on that), or we use the local config_setting as proposed above.

zhangwenlong8911:
> I see, thanks. We do not have that on the Github actions CI, hence we could only merge after it's added to bazel platforms (sounds like there is some movement on that), or we use the local config_setting as proposed above.

Thank you for your review. We are trying to promote the merging of bazel and hope more people will contribute to it

zhangwenlong8911:
> Thanks for also trying that. I've misunderstood the problem. It is not that we are using a too old platforms module - loongarch is not yet 'supported' (added to Bazel) even in v1.0 of its platforms. I've pinged [bazelbuild/platforms#107](https://github.com/bazelbuild/platforms/issues/107) to support adding it.
> 
> (Out of curiosity, what prompted you to add @platforms//cpu:loongarch64 - are you using a locally patched version of Bazel?)
> 
> To move forward, I believe we can add the following to BUILD:
> 
> ```
> # loongarch64 is not yet in @platforms//cpu, define it locally.
> constraint_value(
>     name = "cpu_loongarch64",
>     constraint_setting = "@platforms//cpu:cpu",
> )
> 
> config_setting(
>     name = "loongarch64",
>     constraint_values = [":cpu_loongarch64"],
> )
> ```
> 
> Then use :loongarch64 in the select() calls instead of @platforms//cpu:loongarch64.

Thank you for your review. 

[platforms ](https://github.com/bazelbuild/platforms/issues/107) already supports loongarch64,


https://github.com/bazelbuild/platforms/blob/main/cpu/BUILD

jan-wassenberg:
Congrats, glad to see that land :)
I just [re-ran](https://github.com/google/highway/actions/runs/24644207459/job/76189936888). I think we can land this change in Highway once there is a Bazel release including the new loongarch64, and we change the Bazel version in the MODULE.bazel here.

PR #2989, opened 2026-04-15
Add cbrt

RaviTriv:
Implementing `cbrt` from [wishlist](https://github.com/google/highway/blob/master/g3doc/op_wishlist.md?plain=1#L41).

hehe this one is a bit tricky, as always I really appreciate your time, effort and thoughts!

(Leaving in draft for a bit to verify passing in CI)

### DIFF
--- a/hwy/contrib/math/math-inl.h
+++ b/hwy/contrib/math/math-inl.h
@@ -1472,9 +1472,9 @@ HWY_INLINE V Atanh(const D d, V x) {
              Xor(kHalf, sign));
 }
 
-// Based on musl libc cbrt (MIT-licensed).
-// Copyright © 2005-2020 Rich Felker, et al.
-// See https://git.musl-libc.org/cgit/musl/tree/COPYRIGHT
+// Modified from BSD-licensed code
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+// See https://github.com/libjxl/libjxl/blob/main/LICENSE.
 template <bool kHandleSubnormals, class D, class V>
 HWY_INLINE V Cbrt(const D d, V x) {
   using T = TFromD<D>;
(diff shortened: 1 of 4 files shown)


hwy/contrib/math/math-inl.h
```diff
@@ -1457,6 +1472,71 @@ HWY_INLINE V Atanh(const D d, V x) {
...
+  }
+
+  if constexpr (kHandleSubnormals) {
+    // 1 / cbrt(kScale), 1 / 2^8 or 1 / 2^18
+    const auto kUnscale = Set(d, kIsF32 ? static_cast<T>(1.0 / 256.0)
+                                        : static_cast<T>(1.0 / 262144.0));
+    y = MaskedMulOr(y, is_denormal, y, kUnscale);
+  }
+
+  y = IfThenElse(Or(Eq(abs_x, Zero(d)), Or(IsInf(abs_x), IsNaN(abs_x))), abs_x,
```
jan-wassenberg: Use IsFinite instead of IsInf/IsNaN?

hwy/contrib/math/math-inl.h
```diff
@@ -141,6 +141,21 @@ HWY_NOINLINE V CallAtan2(const D d, VecArg<V> y, VecArg<V> x) {
   return Atan2(d, y, x);
 }
 
+/**
+ * Highway SIMD version of std::cbrt(x).
+ *
+ * Valid Lane Types: float32, float64
+ *        Max Error: ULP = 2
```
jan-wassenberg: Do you have particular accuracy goals?
Consider also basing on https://github.com/libjxl/libjxl/blob/main/lib/jxl/base/fast_math-inl.h, which is 6 ULP. This file also has Erf.

hwy/contrib/math/math-inl.h
```diff
@@ -1457,6 +1472,71 @@ HWY_INLINE V Atanh(const D d, V x) {
              Xor(kHalf, sign));
 }
 
+template <bool kHandleSubnormals, class D, class V>
+HWY_INLINE V Cbrt(const D d, V x) {
+  using T = TFromD<D>;
+
+  const V sign = And(SignBit(d), x);
+  const V abs_x = Abs(x);
```
jan-wassenberg: Consider sign = Xor(abs_x, x).

hwy/contrib/math/math-inl.h
```diff
@@ -1457,6 +1472,71 @@ HWY_INLINE V Atanh(const D d, V x) {
...
+  x = abs_x;
+
+  constexpr bool kIsF32 = (sizeof(T) == 4);
+
+  MFromD<D> is_denormal;
+  if constexpr (kHandleSubnormals) {
+    const V kMinNormal =
+        Set(d, kIsF32 ? static_cast<T>(1.175494351e-38f)
```
jan-wassenberg: Consider hoisting this into base.h, like Epsilon()?

hwy/contrib/math/math-inl.h
```diff
@@ -1457,6 +1472,71 @@ HWY_INLINE V Atanh(const D d, V x) {
...
+  }
+
+  const RebindToUnsigned<D> du;
+  using TU = TFromD<decltype(du)>;
+  // Reciprocal of 3 for Barrett reduction, ceil(2^33 / 3) or ceil(2^65 / 3)
+  const auto kMagicRecip =
+      Set(du, kIsF32 ? static_cast<TU>(0xAAAAAAABu)
+                     : static_cast<TU>(0xAAAAAAAAAAAAAAABULL));
+  // Exponent bias correction, (bias - bias / 3 - 0.03306235651) * 2^mantissa
+  // See https://git.musl-libc.org/cgit/musl/tree/src/math/cbrtf.c#n24
```
jan-wassenberg, line 1507: If this code is based on Musl, please add a comment on its license status (MIT), plus at least a reference to the MIT license header.

hwy/contrib/math/math-inl.h
```diff
@@ -1457,6 +1472,71 @@ HWY_INLINE V Atanh(const D d, V x) {
...
+  x_bits = Add(ShiftRight<1>(MulHigh(x_bits, kMagicRecip)), kMagic);
+  auto y = BitCast(d, x_bits);
+
+  // Newton method, aproximately double accuracy each iteration
+  const auto kOneThird = Set(d, static_cast<T>(1.0 / 3.0));
+  const auto kTwo = Set(d, static_cast<T>(2.0));
+  y = Mul(kOneThird, MulAdd(kTwo, y, Div(x, Mul(y, y))));
+  y = Mul(kOneThird, MulAdd(kTwo, y, Div(x, Mul(y, y))));
+  y = Mul(kOneThird, MulAdd(kTwo, y, Div(x, Mul(y, y))));
+  if (!kIsF32) {
```
jan-wassenberg: if constexpr

jan-wassenberg (commented):
Nice :) Some initial suggestions.

hwy/contrib/math/math-inl.h
RaviTriv: nope, nothing particular in mind. Thanks for sharing, this is cool!
- reading up on this some more and messing around with this approach, will keep you posted :)

RaviTriv: Took a closer look here and while researching this further came across this useful resource on [computing real cube root](https://csclub.uwaterloo.ca/~pbarfuss/qbrt.pdf).

So looks like we have two options:
- `cbrt(x)` directly, the musl approach
  - full range but has float div in newton
- `1 / cbrt(x)`, jpegxl and [vectorclass](https://github.com/libjxl/libjxl/blob/main/lib/jxl/base/fast_math-inl.h#L177).
  - no div but looks like there is some overflow / underflow risk
    - in newton we raise r^4
    - comment in vectormath_exp.h, line 12221 warning that there may be overflow
 
 thoughts / suggestions ?

jan-wassenberg: I'm thinking the JXL approach is still good. A division in Newton/Halley can be quite costly, especially when the next iteration depends on its result.

One thing that can be improved is the initial estimate based on exp/3. First, we can use exp%3 to derive a correction factor and look that up in a table (Lookup8 was not yet available when that code was written).
Second, we might be able to approximate that exp/3 using a TableLookupBytes of exp>>4. I figure that could save a decent number of ops.

jan-wassenberg:
Some more ideas:
For exp/3, (E * 85) >> 8 is correct for -126..127. We can probably also use 16-bit MulHigh to get a >> 16 for free.
Given exp remainder (0,1,2) and mantissa MSB, we can precompute a table of 6 initial guesses, and use Lookup8.
Then MulByFloorPow2 to include the exp/3 part. I think one NR iteration plus the cbrt finalizer step are enough.

RaviTriv:
> Some more ideas: For exp/3, (E * 85) >> 8 is correct for -126..127. We can probably also use 16-bit MulHigh to get a >> 16 for free. Given exp remainder (0,1,2) and mantissa MSB, we can precompute a table of 6 initial guesses, and use Lookup8. Then MulByFloorPow2 to include the exp/3 part. I think one NR iteration plus the cbrt finalizer step are enough.

thanks :) 

gonna continue experimenting and get back to you soon

RaviTriv:
> Some more ideas: For exp/3, (E * 85) >> 8 is correct for -126..127. We can probably also use 16-bit MulHigh to get a >> 16 for free. Given exp remainder (0,1,2) and mantissa MSB, we can precompute a table of 6 initial guesses, and use Lookup8. Then MulByFloorPow2 to include the exp/3 part. I think one NR iteration plus the cbrt finalizer step are enough.

Thanks for your patience, quick check in on what I roughly have in mind based on feedback.

```C++
// Extract sign
// Handle subnormals
constexpr bool kIsF32 = (sizeof(T) == 4);
if constexpr (kIsF32){
    // Get val as bits
    const RebindToSigned<D> di;
    using TI = TFromD<decltype(di)>;
    auto x_int = BitCast(di, x);

    // get biased exp
    auto biased_exp = ShiftRight<23>(x_int);
    auto exp = Sub(biased_exp, Set(di, static_cast<TI>(127)));
    
    // Barrett reduction
    // shift 3 x 128 to avoid off by one issues with negatives 
    auto exp_shifted = Add(exp, Set(di, static_cast<TI>(384)));
    auto exp_shifted_div_3 =
      ShiftRight<16>(Mul(exp_shifted, Set(di, static_cast<TI>(0x5556))));
    auto exp_div_3 = Sub(exp_shifted_div_3, Set(di, static_cast<TI>(128)));
    auto exp_mod_3 =
      Sub(exp_shifted, Mul(exp_shifted_div_3, Set(di, static_cast<TI>(3))));

    // get mantissa msb
    auto mantissa_msb = And(ShiftRight<22>(x_int), Set(di, static_cast<TI>(1)));
    auto idx = Add(ShiftLeft<1>(exp_mod_3), mantissa_msb);

    HWY_ALIGN static constexpr float arr[8] = {
      0.937f, // (1/cbrt(1) + 1/cbrt(1.5)) / 2,
      //.....
      0.525f, // (1/(cbrt(6) + 1/cbrt(8))) / 2 
      };
    auto r = Lookup8(d, arr, idx);

    r = MulByPow2(r, Neg(exp_div_3));

    // Newton
    const auto kThird = Set(d, static_cast<T>(1.0f / 3.0f));
    const auto kFourThirds = Set(d, static_cast<T>(4.0f / 3.0f));
    const auto x_3 = Mul(kThird, x);
    const auto r2 = Mul(r, r);
    r = NegMulAdd(x_3, Mul(r2, r2), Mul(kFourThirds, r));
    
    // Final with better precision:
    r2 = Mul(r, r);
    r = MulAdd(kThird, NegMulAdd(x, Mul(r2, r2), r), r);
    
    // get cbrt from 1 / cbrt(x)
    y = Mul(Mul(r, r), x);
}else{
  // Hanlde f64
}

// Handle Subnormals
// Handle Edge Cases
// Reapply Sign

return y;
```
- When i experimented with a lookup table for exp / 3 may it seemed like it may not work because the initial estimate was too far off.
- needa optimize the mantissa_mtb lookup table
- probably need apply scaling factor for values close to over/underflow threshold? 
- still needa sort out f64 😅

thoughts?

jan-wassenberg:
Looks pretty good already :) Some further improvements:
It's important to minimize the relatively slow 32-bit Mul. Was thinking to BitCast to u16, then MulHigh; only the even lanes would be valid. We'd require a 32-bit shift to shift even u16 into the correct bits in u32 lanes. I suppose we could also do a 16-bit Mul instead.
Remainder calculation should also use u16 Mul.
Instead of Neg(exp_div_3), we could just compute 128-exp_shifted_div_3?
I think the rcp_cbrt -> cbrt step can be fused with the final Newton step - we compute the tentative result `y0 = r2*x`, then feed that into the a Newton step for cbrt: `y0*(5/3 - 2*3*h)`, where `h=y0*r`.

RaviTriv:
thanks! Pushed up some changes for the refactored approach and optimizations, have a few CI fails that i needa investigate.

hwy/contrib/math/math-inl.h
```diff
@@ -1457,6 +1472,122 @@ HWY_INLINE V Atanh(const D d, V x) {
...
+  const VI x_int = BitCast(di, x);
+
+  // Extract exponent and shift (3*128 or 3*512) to keep non-negative for
+  // Barrett reduction
+  const VI exp_shifted =
+      Add(ShiftRight < kIsF32 ? 23 : 52 > (x_int),
+          Set(di, kIsF32 ? static_cast<TI>(257) : static_cast<TI>(513)));
+
+  // Narrow to i16 so MulHigh is cheaper than i32/i64
+  const Rebind<int16_t, D> di16;
```
jan-wassenberg: Better to just Repartition and BitCast, we don't actually want to change the bits, just reinterpret so that Mul will use the 16-bit instruction, with results only valid in every second lane.

jan-wassenberg (changes requested):
LGTM with one minor improvement.
Looks like SVE2 and Arm are failing, but I didn't see any issue yet.

hwy/contrib/math/math-inl.h
RaviTriv: Thanks!

Good idea, updated to repartition and bitcast on multi lane targets.

jan-wassenberg:
Sorry about the unrelated build warnings/errors, these are being fixed in #3023.
Do you know what's the remaining armv7 issue?

RaviTriv:
no worries, gonna take a closer look at the armv7 issue today or 2mro.

will keep you posted if i run into issues / get stuck.

thanks!

RaviTriv:
image

on armv7 during newton we were entering subnormal range on extreme values and the architecture flushed to zero.

updated to reorder the multiplication to prevent value from becoming too small. 

pipeline green :)

hwy/contrib/math/math-inl.h
```diff
@@ -1457,6 +1472,142 @@ HWY_INLINE V Atanh(const D d, V x) {
...
+  const VI exp_shifted =
+      Add(ShiftRight < kIsF32 ? 23 : 52 > (x_int),
+          Set(di, kIsF32 ? static_cast<TI>(257) : static_cast<TI>(513)));
+
+  // Barrett reduction (n/3) via MulHigh by 0x5556
+  // Cannot Repartition to smaller lanes on a single-lane target
+  VI exp_shifted_div_3;
+  VI exp_mod_3;
+  if constexpr (HWY_MAX_LANES_D(D) > 1) {
+    if (Lanes(d) > 1) {
```
jan-wassenberg, line 1545: I think we're unduly penalizing SVE/RVV here with a runtime branch. It is vanishingly rare outside of tests for there to be a single-lane D. How about we shift the div/mod code into a function so we can more easily reason about this?
The happy path would be `if constexpr (detail::IsFull(d))`, then we know there are >1 lanes and no runtime check needed. Only if false should we check Lanes().

hwy/contrib/math/math-inl.h
```diff
@@ -1457,6 +1472,142 @@ HWY_INLINE V Atanh(const D d, V x) {
...
+          Set(di, kIsF32 ? static_cast<TI>(1) : static_cast<TI>(3)));
+  const VI idx = Add(ShiftLeft < kIsF32 ? 1 : 2 > (exp_mod_3), top_mant);
+
+  V r;
+  if constexpr (kIsF32) {
+    // (1/cbrt(lo) + 1/cbrt(hi))/2 over 6 bins of [1,8)
+    HWY_ALIGN static constexpr float initial_guess[8] = {
+        0.92807984f, 0.81504166f, 0.73603648f, 0.65004617f,
+        0.58375800f, 0.51406258f, 0.0f,        0.0f};
+    if constexpr (HWY_MAX_LANES_D(D) >= 4) {
```
jan-wassenberg: fast_math-inl has some example code for checking whether Lookup8 is safe:
`  if constexpr ((HWY_MAX_LANES_D(D) >= 4 && !HWY_HAVE_SCALABLE) ||
                (HWY_HAVE_SCALABLE && sizeof(T) == 4 && detail::IsFull(d))) {
`
In general, we want to minimize use of Lanes(), especially in conditionals.

jan-wassenberg (changes requested):
Congrats, good find :) Just two more updates on the Lanes checks, then we're good to go :D

hwy/contrib/math/math-inl.h
RaviTriv: good idea! updated

jan-wassenberg (approved):
Very nice! LGTM, thanks for updating.

RaviTriv:
> Very nice! LGTM, thanks for updating.

thank you : )

any ideas why the copybara job is failing?

jan-wassenberg:
:) It's an unrelated internal failure. We're retrying now.

PR #3000, opened 2026-04-21
Implement Dot Product examples

henrixapp:
Hi,

I have added a simple dot-product  and a more optimized unrolled example based on https://github.com/google/highway/blob/master/hwy/examples/sum_array_advanced.cc 
Instead of using FillBytes, I wrote a uniform distribution based fill (with a parameter to change the maximum value generated).
FillBytes in the above example leads to zero sums in floats, that's why I changed the behavior here.

Here are some results on a Zen4 mobile chip (with power supplied):

```sh
$ ./examples/dot_product                                                                                  
Float - Scalar dot: 2533.756348 (took 0.574623 seconds)
Float - SIMD dot:   2533.757080 (took 0.048921 seconds)
Float - Speedup:    11.745826x
Double - Scalar dot: 2533.756961 (took 0.582187 seconds)
Double - SIMD dot:   2533.756961 (took 0.096564 seconds)
Double - Speedup:    6.029003x
Int32 - Scalar dot: 205553.000000 (took 0.134996 seconds)
Int32 - SIMD dot:   205553.000000 (took 0.048379 seconds)
Int32 - Speedup:    2.790400x
All Validations PASSED
```
I have also tested to unroll the loops and use the more speciall MulAdd function, this did not  lead to  significant performance increase. 
```
$ ./examples/dot_product_unroll                                                                             
Float - Scalar sum: 2533.756348 (took 0.572441 seconds)
Float - SIMD Unrolled sum:   2533.756836 (took 0.048170 seconds)
Float - Speedup (Unrolled):    11.883801x
Double - Scalar sum: 2533.756961 (took 0.577112 seconds)
Double - SIMD Unrolled sum:   2533.756961 (took 0.097204 seconds)
Double - Speedup (Unrolled):    5.937099x
Int32 - Scalar sum: 205553.000000 (took 0.137622 seconds)
Int32 - SIMD Unrolled sum:   205553.000000 (took 0.048350 seconds)
Int32 - Speedup (Unrolled):    2.846388x
All Validations PASSED
```

Instead I saw that having both variants in the same test-harness, lead to the compiler not optimizing the scalar version (increasing its running time by a factor of two). So I opted to have two separate files.

Should I also add a bazel `BUILD` file?


hwy/examples/dot_product.cc
```diff
@@ -0,0 +1,170 @@
...
+
+// HWY_ONCE marks code that will be compiled once instead once per target
+// architecure
+#if HWY_ONCE
+namespace hwy {
+// Fills array with random values between 0 and max_v
+template <typename T>
+void FillRandom(const std::uint64_t seed, T* HWY_RESTRICT result,
+                const size_t size, T max_v) {
+  std::mt19937_64 engine(seed);
```
jan-wassenberg, line 88: Recommend using RandomFiniteValue from test_util.h instead, mt19937 is not something we recommend in general :)

hwy/examples/dot_product_unroll.cc
```diff
@@ -0,0 +1,203 @@
...
+}  // namespace hwy
+HWY_AFTER_NAMESPACE();
+
+// HWY_ONCE marks code that will be compiled once instead once per target
+// architecure
+#if HWY_ONCE
+namespace hwy {
+// Fills array with random values between 0 and max_v
+template <typename T>
+void FillRandom(const std::uint64_t seed, T* HWY_RESTRICT result,
```
jan-wassenberg, line 118: Here also RandomFiniteValue.

jan-wassenberg (changes requested):
Nice speedup, congrats :) I'm surprised that the unrolling didn't help, likely because the compiler managed it already. I'd suggest only keeping the unrolled version/file because this is still recommended for more complex code examples.

Yes, please also add to the root-directory BUILD file in the same way as sum_array_advanced.

hwy/examples/dot_product.cc
henrixapp: I tried:
```cpp
 RandomState rng{seed};
  for (size_t i = 0; i < size; i++) {
    result[i] = RandomFiniteValue<T>(&rng);
  }
```
but this leads to -nan values as a result and failing validation:
```
Float - Scalar sum: -nan (took 0.665675 seconds)
Float - SIMD Unrolled sum:   -nan (took 0.023965 seconds)
Float - Speedup (Unrolled):    27.776584x
Float Validation FAILED
```

jan-wassenberg: huh, that's very surprising. First, initializing rng with a seed is not necessary, the default is fine. But that should not break things.
What's the output if you printf the individual generated values, e.g. `fprintf(stderr, "%f\n", static_cast<double>(result[i])); `?

henrixapp: I set count to 5, the values look a bit too large
```
0.000000
-0.000000
-28613490490126368768.000000
-2662403224467430258390335488.000000
-0.000000
4593175659997029457063598050246656.000000
-6273333568570962737797393603887104.000000
19547652295783749948031855755264.000000
-107985210665647607109386240.000000
0.000000
Float - Scalar sum: -nan (took 0.000569 seconds)
Float - SIMD Unrolled sum:   -nan (took 0.000253 seconds)
```
When running with `feenableexcept(FE_INVALID | FE_DIVBYZERO | FE_OVERFLOW);` i get a exception at the i=2 entry, which makes sense looking at the magnitude of the numbers.

I am using the seed to have different values for the two arrays, but I can also switch to passing the RNG.

jan-wassenberg: Oh, I see, thanks. Those numbers are indeed finite but huge, enough so that multiplying them can overflow.

I'm very tempted to move AesCtrEngine, RngStream and Random(Normalized)Float from gemma.cpp into Highway, but after the upcoming 1.4 release. I suppose we could land this change for now with std::mt19937.

henrixapp: What about using https://github.com/google/highway/blob/master/hwy/contrib/random/random-inl.h#L247 and scaling the values with a smaller, finite constant?

jan-wassenberg: The VectorXoshiro class is overkill here and using internal::Xoshiro is also not great. Let's go with mt19937, there are anyway a couple of existing usages and we could replace them all later.

henrixapp:
Thank you for your feedback. I have added a cc_binary to the BUILD file. I noticed that running under bazel with `-c opt` has different optimizations level than Cmake with ReleaseType=Release.

```
$ bazel run -c opt :dot_product_unroll
Float - Scalar sum: 2533.756348 (took 1.052195 seconds)
Float - SIMD Unrolled sum:   2533.756836 (took 0.050711 seconds)
Float - Speedup (Unrolled):    20.748926x
Double - Scalar sum: 2533.756961 (took 1.527101 seconds)
Double - SIMD Unrolled sum:   2533.756961 (took 0.099519 seconds)
Double - Speedup (Unrolled):    15.344748x
Int32 - Scalar sum: 205553.000000 (took 0.506948 seconds)
Int32 - SIMD Unrolled sum:   205553.000000 (took 0.049204 seconds)
Int32 - Speedup (Unrolled):    10.303053x
All Validations PASSED
```
vs.
```
$ ./examples/dot_product_unroll
Float - Scalar sum: 2533.756348 (took 0.601308 seconds)
Float - SIMD Unrolled sum:   2533.756836 (took 0.052166 seconds)
Float - Speedup (Unrolled):    11.526715x
Double - Scalar sum: 2533.756961 (took 0.611199 seconds)
Double - SIMD Unrolled sum:   2533.756961 (took 0.101009 seconds)
Double - Speedup (Unrolled):    6.050953x
Int32 - Scalar sum: 205553.000000 (took 0.143256 seconds)
Int32 - SIMD Unrolled sum:   205553.000000 (took 0.051944 seconds)
Int32 - Speedup (Unrolled):    2.757870x
All Validations PASSED
```

leading to larger performance "improvements" caused by longer running time of the scalar code.

jan-wassenberg:
Ah, that matches my experience. Good to understand the cause of unrolling helping less, interesting that CMake has more aggressive optimization settings (automatically unrolling an fp accumulator requires something like -ffast-math which is dangerous).

One [build issue](https://github.com/google/highway/actions/runs/24766911837/job/72463289014?pr=3000): Armv7 lacks fp64, please wrap anything using `double` in `#if HWY_HAVE_FLOAT64`.

jan-wassenberg (approved):
Thanks for making the change. We will land this shortly, after the upcoming release.

Issue #2974, opened 2026-04-09
6/859 tests fail with gcc-16.0.1 on armv7 with: EMU128  # GetParam() = EMU128 (Subprocess aborted)


stefson:
```
The following tests FAILED:
        595 - HwyShiftTestGroup/HwyShiftTest.TestAllVariableRoundingShr/EMU128  # GetParam() = EMU128 (Subprocess aborted)
        822 - SortTestGroup/SortTest.TestAllSortIota/EMU128  # GetParam() = EMU128 (Subprocess aborted)
        824 - SortTestGroup/SortTest.TestAllSort/EMU128  # GetParam() = EMU128 (Subprocess aborted)
        826 - SortTestGroup/SortTest.TestAllSelect/EMU128  # GetParam() = EMU128 (Subprocess aborted)
        828 - SortTestGroup/SortTest.TestAllPartialSort/EMU128  # GetParam() = EMU128 (Subprocess aborted)
        832 - SortTestGroup/SortTest.TestAllFloatInf/EMU128  # GetParam() = EMU128 (Subprocess aborted)
Errors while running CTest
```
compressed build and tests logs: [logfiles.tar.gz](https://github.com/user-attachments/files/26609673/logfiles.tar.gz)

highway git checkout was: https://github.com/google/highway/commit/3b9166ffed76d3b40ffad97fc854d69c3ffdeca6

gcc-16 is a snapshot from last week: 

```
Using built-in specs.
COLLECT_GCC=gcc-16
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/armv7a-unknown-linux-musleabihf/16/lto-wrapper
Target: armv7a-unknown-linux-musleabihf
[...]

PR #3024, opened 2026-04-30
Enable testing on Fedora Rawhide

### LINKED ISSUE #3022: Test on leading edge linux distribution
Currently tests are not done with new releases of GCC compilers as Ubuntu stable releases take some time to get updated.
Consider using a container to run tests on a distribution such as Fedora rawhide, OpenSuse Tumbleweed or Arch to get newer compiler versions.  Happy to make a pull request that would do this.

bkmgit:
Uses Fedora infrastructure https://packit.dev

Closes #3022

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -1,0 +1,11 @@
+specfile_path: fedora/highway.spec
+
+notifications:
+  pull_request:
+    successful_build: true
+
+jobs:
+- job: copr_build
+  trigger: pull_request
+  targets:
+  - fedora-rawhide
--- a/fedora/highway.spec
+++ b/fedora/highway.spec
@@ -1,0 +1,75 @@
+%global common_description %{expand:
+Highway is a C++ library for SIMD (Single Instruction, Multiple Data), i.e.
+applying the same operation to 'lanes'.}
+ 
+Name:           highway
+Version:        1.4.0
+Release:        %autorelease
+Summary:        Efficient and performance-portable SIMD
+ 
+License:        (Apache-2.0 OR BSD-3-Clause) AND CC0-1.0
+URL:            https://github.com/google/highway
+Source:         %url/archive/%{version}/%{name}-%{version}.tar.gz
+ 
+BuildRequires:  cmake
+BuildRequires:  gcc-c++
+BuildRequires:  gtest-devel
+ 
+%description
+%common_description
+%package        devel
+Summary:        Development files for Highway
+Requires:       %{name}%{?_isa} = %{version}-%{release}
+ 
+%description devel
+%{common_description}
+ 
+Development files for Highway.
+ 
+%package        doc
+Summary:        Documentation for Highway
+BuildArch:      noarch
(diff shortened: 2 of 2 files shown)


bkmgit:
At present tests on ppc64 and build on aarch64 fail:
 https://koji.fedoraproject.org/koji/taskinfo?taskID=144992149

bkmgit:
For tests on Fedora to run, enable Packit integration. See:
https://packit.dev/docs/guide#github

jan-wassenberg (commented):
Nice. Probably good to wait a bit until we can remove CC0.

jan-wassenberg:
Had a quick look at the failing builds.

aarch64 Error: unknown pseudo-op: `.lbe73430'
Probably OOM leading to truncated assembly file. I wonder if we can influence the flags, e.g. remove -ffat-lto-objects?
Might also help to split compare_test and/or disable some targets, I can do that.

ppc:
```
[----------] 1 test from HwyMemoryTestGroup/HwyMemoryTest
[ RUN      ] HwyMemoryTestGroup/HwyMemoryTest.TestAllLoadDup128/EMU128
i16x8 expect [0+ ->]:
  0x0001,0x0002,0x0003,0x0004,0x0005,0x0006,0x0007,
i16x8 actual [0+ ->]:
  0x0110,0x0000,0x0000,0x0000,0x0001,0x0002,0x0003,
Abort at memory_test.cc:393221: EMU128, i16x8 lane 0 mismatch: expected '0x0001', got '0x0110'.
```
This kind of thing is usually a compiler mistake caused by incorrect constant folding. I'm adding hwy::Unpredictable1() there which might help.

bkmgit:
For aarch64 another issue is fixed by  https://github.com/google/highway/pull/3030
If  fewer cores are used to compile, the build succeeds. 

Failing build that resulted in #3030
 https://koji.fedoraproject.org/koji/taskinfo?taskID=145011276
Build with fix:
 https://koji.fedoraproject.org/koji/taskinfo?taskID=145019086

Disabling LTO also reduces memory requirements, but not using fat LTO objects might be sufficient in
most cases, see:
https://stackoverflow.com/questions/13799452/what-is-the-difference-in-gcc-between-lto-and-fat-lto-objects

bkmgit:
AArch64 on Copr seems to have another failing test:
https://copr.fedorainfracloud.org/coprs/fed500/highway/build/10415978/

The following tests FAILED:
	1546 - HwyReductionTestGroup/HwyReductionTest.TestAllMaskedReduceSum/SVE_256 (Subprocess aborted)
	1558 - HwyReductionTestGroup/HwyReductionTest.TestAllMaskedReduceMax/SVE_256 (Subprocess aborted)

The machines used on Copr are likely different than the ones used on the koji tests.  It may be helpful to log more information about the CPU microarchitecture being used and available RAM when doing tests.

Arm seems to have several other bugs:
https://github.com/google/highway/issues/2974

PR #3030, opened 2026-05-01
Allow bug workaround for GCC 16

bkmgit:
Bug has not been fixed in GCC 16 yet.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=115556


jan-wassenberg (approved):
Thanks :) Here's hoping for GCC 17, hah.

PR #3058, opened 2026-05-17
Add ERF

RaviTriv:
Implementing `erf` from [wishlist](https://github.com/google/highway/blob/master/g3doc/op_wishlist.md?plain=1#L41).

Thanks for time and effort reviewing :)

**Notes / Questions**
- Suggested target precision to aim for?
- Currently basing off the [cephes approach](https://github.com/jeremybarnes/cephes/blob/master/single/ndtrf.c)
  - Originally looked at [JXL](https://github.com/libjxl/libjxl/blob/main/lib/jxl/base/fast_math-inl.h), but think we may need piecewise approximation in this case to support larger range?

### DIFF
--- a/g3doc/op_wishlist.md
+++ b/g3doc/op_wishlist.md
@@ -38,7 +38,7 @@ _mm512_getmant (f32/f64)
 
 High-precision! Consider copying from SLEEF. See #1650.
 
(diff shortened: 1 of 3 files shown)


jan-wassenberg (approved):
Nice, LGTM :) Would you like to land this already?

RaviTriv:
> Nice, LGTM :) Would you like to land this already?

Yep, that would be awesome. Thank you!

PR #3063, opened 2026-05-26
Fixed PPC10 AndXor implementation

johnplatts:
Fixed compile errors with PPC10 AndXor implementation.


jan-wassenberg (approved):
Oops, thank you for fixing.

PR #3067, opened 2026-05-29
add contrib/dot benchmarks

mytsvy:
Add a dot product benchmark that measures both generic and kMultipleOfVector paths side-by-side across common sizes.
Dot::Compute has assumption-dependent fast paths; benchmarking only the generic path hides key perf behavior. This adds an explicit comparison in one run.

## Test:

```
[ RUN      ] DotBenchGroup/DotBench.BenchAllDot/AVX2
Dot same (f32x8):
256 elems: gen  163.04 GB/s (MAD 0.29%) | mult  160.82 GB/s (MAD 0.28%)
1024 elems: gen  209.07 GB/s (MAD 0.22%) | mult  193.11 GB/s (MAD 1.06%)
16384 elems: gen   95.18 GB/s (MAD 0.89%) | mult   95.75 GB/s (MAD 0.97%)
262144 elems: gen   54.48 GB/s (MAD 0.47%) | mult   54.50 GB/s (MAD 0.45%)
1048576 elems: gen   41.13 GB/s (MAD 5.64%) | mult   44.42 GB/s (MAD 1.92%)
[        OK ] DotBenchGroup/DotBench.BenchAllDot/AVX2

```

### DIFF
--- a/hwy/contrib/dot/dot_bench.cc
+++ b/hwy/contrib/dot/dot_bench.cc
@@ -1,0 +1,217 @@
+// Copyright 2026 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+
+#include "hwy/aligned_allocator.h"
+#include "hwy/nanobenchmark.h"
+#include "hwy/per_target.h"
+#include "hwy/timer.h"
+
+// clang-format off
+#undef HWY_TARGET_INCLUDE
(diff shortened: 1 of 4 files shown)


hwy/contrib/dot/dot_bench.cc
```diff
@@ -0,0 +1,217 @@
...
+  auto a = AllocateAligned<TA>(max_count);
+  auto b = AllocateAligned<TB>(max_count);
+  HWY_ASSERT(a && b);
+
+  RandomState rng(static_cast<uint64_t>(Unpredictable1()) *
+                  0x9E3779B97F4A7C15ULL);
+  FillArrays(a.get(), b.get(), max_count, rng);
+
+  const size_t N = Lanes(d);
+  (void)N;
```
jan-wassenberg, line 94: Let's use N when passing to TypeName, then no need for the void cast?

hwy/contrib/dot/dot_bench.cc
```diff
@@ -0,0 +1,217 @@
...
+  params.verbose = false;
+  Result results[1];
+  FuncInput input = Unpredictable1();
+
+  for (size_t count : kCounts) {
+    const double bytes =
+        static_cast<double>(count) * (sizeof(TA) + sizeof(TB));
+
+    const size_t num_results_generic = MeasureClosure(
+        [&](FuncInput) {
```
jan-wassenberg, line 107: To ensure the computation isn't elided, can we make Compute() depend on the FuncInput? For example, we could set input to count * Unpredictable1(), then pass the FuncInput as the count arg to Compute?

jan-wassenberg (changes requested):
Thanks for adding :) Two minor suggestions:

jan-wassenberg:
Also looks like we have some check failures: signed/unsigned casts.

PR #3076, opened 2026-06-02
Only include `hwy_list_targets` if top-level build; only build tests if `HWY_ENABLE_TESTS=ON`

phongn:
I'm working on integrating Highway with the [Apache Traffic Server](https://github.com/apache/trafficserver). I ran into a few small CMake-related build issues when including Highway as an in-tree vendor'ed copy with some things like examples, docs, tests, etc. removed:

* Tests build even if `HWY_ENABLE_TESTS=OFF`
* The `hwy_list_targets` always builds even if it isn't ultimately called (say, because it's included and not standalone). The Meson build system [does not](https://github.com/google/highway/blob/master/meson.build#L507-L513), and the Bazel build system [marks it as a test](https://github.com/google/highway/blob/master/BUILD#L598-L606).

This PR proposes to try and bring the CMake behavior in-line with Meson and Bazel. 

The current files declares a minimum version of CMake 3.10, so I did not use `PROJECT_IS_TOP_LEVEL`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -598,6 +598,7 @@ if(UNIX AND NOT APPLE)
 endif()
 endif()  # HWY_ENABLE_CONTRIB
 
+if (HWY_ENABLE_TESTS)
 add_library(hwy_test ${HWY_LIBRARY_TYPE} ${HWY_TEST_SOURCES})
 target_link_libraries(hwy_test PUBLIC hwy)
 target_compile_options(hwy_test PRIVATE ${HWY_FLAGS})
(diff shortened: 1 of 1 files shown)


phongn:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/google/highway/pull/3076/checks?check_run_id=79180956200) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

CLA issue resolved

jan-wassenberg (approved):
Seems reasonable. FYI you may still want to build hwy_list_targets to diagnose any issues with compiler flags or see what targets your CPU supports.

eustas:
We invoke `hwy_list_targets` in libjxl (CI) build, so that is easy to see what has gone wrong. With this change our CI become broken. 
Let's add something like `HWY_BUILD_LIST_TARGETS` that defaults to `CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR` (btw, that expression deserves its own variable for better readability and reuasbility)

jan-wassenberg:
I like your idea of adding a CMake variable. Want to send a PR or should I?

eustas:
Yup, will do when get spare cycles. Anyways, it seems that only libjxl needs that currently =)

phongn:
Apologies for breaking your workflow, @eustas. I've submitted #3121 with your suggestions.

eustas:
Thanks for fixing!

PR #3080, opened 2026-06-04
Added Pow op

johnplatts:
Added `Pow(d, a, b)` operation to math-inl.h as `Pow(d, a, b)` is usually more accurate than `Exp(Mul(Log(a), b))`, and `Pow(d, a, b)` is usually within 1 ulp of `std::pow(a[i], b[i])` (whereas the error of `exp(log(a) * b)` can exceed 500 ulp for some values).

`Pow(d, a, b)` also correctly handles special cases such as `pow(1, b)`, `pow(a, 0)`, and `pow(a, b)` for `a < 0`.

`Pow(d, a, b)` also computes `log2(abs(a))` to extra precision to ensure that the fractional part of `log2(|a|) * b` is sufficiently accurate if `|log2(|a|) * b| > 0.5`.

### DIFF
--- a/hwy/contrib/math/math-inl.h
+++ b/hwy/contrib/math/math-inl.h
@@ -384,6 +384,21 @@ HWY_NOINLINE V CallHypot(const D d, VecArg<V> a, VecArg<V> b) {
   return Hypot(d, a, b);
 }
 
+/**
+ * Highway SIMD version of Pow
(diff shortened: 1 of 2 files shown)


jan-wassenberg (approved):
Very nice work, thanks for adding!

PR #3083, opened 2026-06-04
Added Cosh op

johnplatts:
Added Cosh op to math_inl-h.


jan-wassenberg (approved):
LGTM, FYI I intend to later split up math_test into multiple files because it is getting lengthy.

PR #3056, opened 2026-05-15
Add ShiftRightAndDemoteTo / RoundingShiftRightAndDemoteTo

eyupcanakman:
Implements `ShiftRightAndDemoteTo` and `RoundingShiftRightAndDemoteTo` for #2628. NEON fuses both steps into one saturating shift-narrow (`vqshrn`/`vqshrun`/`vqrshrn`/`vqrshrun`); other targets fall back to `DemoteTo(d, [Rounding]ShiftRight<k>(v))`.

The `Demote2To` variants and SVE2/RVV/LSX/LASX overrides will follow in separate PRs per the issue thread.

Issue #2628, opened 2025-07-20
Adding [Rounding]ShiftRightAndDemoteTo ops


johnplatts:
While Clang 20 and later can optimize [Rounding]ShiftRight+DemoteTo down to a single instruction on NEON on AArch64, there are some other targets that can carry out [Rounding]ShiftRight+DemoteTo using fewer instructions than `DemoteTo(dn, [Rounding]ShiftRight<kShiftAmt>(v))` (due to missed optimizations by GCC/Clang as of GCC 15 and Clang 20), including NEON on Armv7, SVE2, RVV, LSX, and LASX.

ShiftRight+DemoteTo is equivalent to NEON `vqshr[u]n_n_*`, SVE2 `svqshr[u]n_n_`+`svuzp1`, RVV `__rvv_vnclip[u]` with the __RISCV_VXRM_RDN rounding mode, LSX `__lsx_vsrani_*`/`__lsx_vsrlni_*`, and LASX `__lasx_xvsrani_*`/`__lasx_xvsrlni_*`+`__lasx_xvpermi_d`.

RoundingShiftRight+DemoteTo is equivalent to NEON `vqrshr[u]n_n_*`, SVE2 `svqrshr[u]n_n_`+`svuzp1`, RVV `__rvv_vnclip[u]` with the __RISCV_VXRM_RNU rounding mode, LSX `__lsx_vsrarni_*`/`__lsx_vsrlrni_*`, and LASX `__lasx_xvsrarni_*`/`__lasx_xvsrlrni_*`+`__lasx_xvpermi_d`.

There are some real-world use cases for RoundingShiftRightAndDemoteTo, including in a portable, cross-platform implementation of the libaom convolve8 ops (whose NEON implementations in aom_dsp/arm/aom_convolve8_neon.c take advantage of vqrshrun_n_s16).

There is also plenty of image processing code that uses `vqrshrun_n_s16` according to a Sourcegraph search.

Should separate [Rounding]ShiftRightAndDemoteTo ops be added due to the better codegen that [Rounding]ShiftRightAndDemoteTo has over  `DemoteTo(dn, [Rounding]ShiftRight<kShiftAmt>(v))` on some targets and the use of such ops in image processing code and video codecs?

johnplatts:
[...]

PR #3087, opened 2026-06-07
Add ShiftRightAndReorderDemote2To / ShiftRightAndOrderedDemote2To

eyupcanakman:
Implements `ShiftRightAndReorderDemote2To` and `ShiftRightAndOrderedDemote2To`, with their rounding variants, for #2628. These are the `Demote2To` follow-up to #3056. NEON fuses the shift and the saturating narrow into one instruction (`vqshrn`/`vqshrun`/`vqrshrn`/`vqrshrun`, using the `_high` form for the second vector); other targets fall back to `ReorderDemote2To(d, [Rounding]ShiftRight<k>(a), ...)` and the Ordered equivalent.

SVE2/RVV/LSX/LASX overrides are left as a TODO per the issue thread.

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -2267,6 +2267,32 @@ or `ConcatOdd` followed by `PromoteLowerTo`:
     more efficient. Note that integer inputs are saturated to the destination
     range as with `DemoteTo`. Only available if `HWY_TARGET != HWY_SCALAR`.
 
+*   `V`,`D`: any `(V, D)` accepted by `ReorderDemote2To`, with `V` and
+    `TFromD<D>` integer \
+    <code>Vec&lt;D&gt; **ReorderShiftRightAndDemote2To**&lt;int kShiftAmt&gt;(D, V
+    a, V b)</code>: equivalent to `ReorderDemote2To(D, ShiftRight<kShiftAmt>(a),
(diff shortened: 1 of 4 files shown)


g3doc/quick_reference.md
```diff
@@ -2267,6 +2267,32 @@ or `ConcatOdd` followed by `PromoteLowerTo`:
     more efficient. Note that integer inputs are saturated to the destination
     range as with `DemoteTo`. Only available if `HWY_TARGET != HWY_SCALAR`.
 
+*   `V`,`D`: any `(V, D)` accepted by `ReorderDemote2To`, with `V` and
+    `TFromD<D>` integer \
+    <code>Vec&lt;D&gt; **ShiftRightAndReorderDemote2To**&lt;int kShiftAmt&gt;(D, V
```
jan-wassenberg: For consistency with other names, how about we make Reorder and Ordered a prefix, such that this op would be called ReorderShiftRightAndDemote2To?

jan-wassenberg (dismissed):
LGTM, just a naming change.

eyupcanakman:
Done. Put Reorder and Ordered first and kept RoundingShiftRight intact, so the rounding pair is ReorderRoundingShiftRightAndDemote2To and OrderedRoundingShiftRightAndDemote2To. Can reorder those two if you'd prefer.

jan-wassenberg (approved):
LGTM, thanks for making the change.

PR #3106, opened 2026-06-10
Fixes to NEON PerBlock2x2MatMul and matmul_op_test

johnplatts:
Added default implementation of PerBlock2x2MatMul in generic_ops-inl.h for targets other than HWY_SCALAR where `HWY_NATIVE_PER_BLOCK_2X2_MATMUL_BF16` is 0.

Also made fix to `d` argument type to PerBlock2x2MatMul implementation in arm_neon-inl.h as there are Simd tags other than `Simd<int32_t, 4, 0>` such as `Simd<int32_t, 2, 1>` and `Simd<int32_t, 1, 2>` that are tags for 16-byte SIMD vectors.

Also added matmul_op_test.cc to CMakeLists.txt.

Also enabled TestInt8PerBlock2x2MatMul and TestBf16PerBlock2x2MatMul for all targets except for HWY_SCALAR.

Also fixed compiler warnings in matmul_op_test.cc.

### DIFF
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -7746,55 +7746,6 @@ HWY_API V Per4LaneBlockShuffle(V v) {
 }
 #endif
 
-// ------------------------------ PerBlock2x2MatMul
-#if HWY_TARGET != HWY_SCALAR || HWY_IDE
-
-#if !HWY_NATIVE_PER_BLOCK_2X2_MATMUL_INT8
-template <class DI32, HWY_IF_I32_D(DI32), HWY_IF_V_SIZE_GT_D(DI32, 8)>
-HWY_API VFromD<DI32> PerBlock2x2MatMul(DI32 di32,
-                                       VFromD<Repartition<int8_t, DI32>> a,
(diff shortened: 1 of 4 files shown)


hwy/ops/generic_ops-inl.h
```diff
@@ -7746,6 +7746,55 @@ HWY_API V Per4LaneBlockShuffle(V v) {
 }
 #endif
 
+// ------------------------------ PerBlock2x2MatMul
+#if HWY_TARGET != HWY_SCALAR || HWY_IDE
```
jan-wassenberg: Thanks for the various fixes. I was thinking to avoid defining this op if not native, though.
It is exotic enough to be expensive to emulate, considerably worse than just using dot products directly.
And it's only supported on Arm. Generally we don't want such performance cliffs. I believe this really does require an `#ifdef` in user code, including different packing/data layout. Does that make sense?

jan-wassenberg (changes requested):

hwy/ops/generic_ops-inl.h
johnplatts: Yes.

jan-wassenberg (approved):
:) Thanks for updating.

PR #3107, opened 2026-06-10
Fixes to SVE BF16 MulEvenAdd and WidenMulEven

johnplatts:
Corrected BF16 WidenMulEven to fix compiler errors if SVE BF16 is enabled.

In addition, fixed compiler warning in BF16 MulEvenAdd to fix compiler warning if SVE BF16 is enabled.


jan-wassenberg (approved):
Thanks for fixing. We recently had another such issue, I've audited all HWY_SVE_HAVE_BF16_FEATURE and found no others.

PR #3109, opened 2026-06-11
Fix build on ppc64/s390x

miladfarca:
Fix is needed after 636d64d8 since said platforms do not support F16 and fail during compilation:
```
hwy/contrib/sort/sort_test.cc:298:39: error: no matching constructor for initialization of 'hwy::float16_t'
  298 |   std::iota(keys.begin(), keys.end(), T{0});
```

### DIFF
--- a/hwy/contrib/sort/sort_test.cc
+++ b/hwy/contrib/sort/sort_test.cc
@@ -354,9 +354,11 @@ void TestSelectWithNaN() {
   }
 #endif
   TestSelectWithNaNForType<float>();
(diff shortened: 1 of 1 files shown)


jan-wassenberg (changes requested):
Oops, my mistake, thanks for fixing. Would you mind adding HWY_HAVE_FLOAT64 around the usage of double, a few lines below?

miladfarca:
Not a problem, I've also added it around the usage of double.

jan-wassenberg (approved):
Thanks :)

PR #3111, opened 2026-06-11
Add IsSorted to hwy/contrib/algo

### LINKED ISSUE #3094: Add IsSorted to hwy/contrib/algo (op_wishlist item)
The op wishlist lists `IsSorted` under "Remaining STL functions for `hwy/contrib/algo`" ([g3doc/op_wishlist.md](https://github.com/google/highway/blob/master/g3doc/op_wishlist.md?plain=1#L51)). I'd like to implement it.

**Proposed semantics**
`IsSorted(d, in, count)` returns `true` iff `in[0, count)` is in non-decreasing order (`in[i] <= in[i + 1]` for all `i`), matching `std::is_sorted`. A range of fewer than two elements is trivially sorted.

**Plan**
- Portable, vectorized non-decreasing scan: compare each element to its successor via an offset `LoadU` + `Le` + `AllTrue`, with an early-exit on the first descent and a masked tail — mirroring `find-inl.h` / `minmax-inl.h`.
- Tested against a scalar reference across element types, lengths, and misalignments, consistent with the existing `contrib/algo` tests.

Does the name and semantics look right before I open a PR?

dcsid:
Adds `IsSorted` to `hwy/contrib/algo`, from the op wishlist. Closes #3094.

`IsSorted(d, in, count)` matches `std::is_sorted` (non-decreasing). Per #3094 there is also a comparator overload: `comp(d, a, b)` returns a mask, true where `a` is ordered strictly before `b` — e.g. passing `Gt` checks descending order. The default forwards `Lt`.

Implementation follows `find-inl.h`: 2x-unrolled loop comparing each vector with an overlapping load offset by one element, `AllFalse(Or(...))` early-exit, masked `LoadN` tail.

Tests check both overloads against scalar references: random and boundary lengths, 3 misalignments, random / sorted / all-equal / single-descent inputs (including descents at block boundaries and the last pair). Pass locally on NEON, NEON_BF16, NEON_WITHOUT_AES, EMU128.

Performance vs `std::is_sorted` (Apple M4 Pro, clang 21 `-O2`, 64Ki sorted elements = full scan; scalar baseline ~15 µs):

| type | lanes | speedup |
|------|------:|--------:|
| int8 | 16 | ~19x |
| int16 | 8 | ~10x |
| int32 / float | 4 | ~4.6x |
| double | 2 | ~2.4x |


dcsid:
Fix for the pre-existing CI breakage is up: https://github.com/google/highway/pull/3113. Happy to rebase this PR once it lands.

jan-wassenberg (approved):
Very nice, thanks for adding - and impressive speedups :) We'll probably be able to merge already because internal tests are green.

PR #3113, opened 2026-06-12
Fix sort_test build on toolchains without native float16

### LINKED ISSUE #3112: sort_test fails to build on toolchains without native float16 (since 636d64d8)
Since 636d64d8 ("Fix VQSelect sentinel/NaN handling", #3098), `sort_test.cc` fails to build on toolchains without native float16 support, where `hwy::float16_t` is the emulated struct:

- `sort_test.cc:298`: `std::iota(keys.begin(), keys.end(), T{0})` — no `float16_t` constructor from `int` (brace-init fails), and `std::iota` also requires `operator++`.
- `sort_test.cc:326`: `x == -kInf` — no unary `operator-` for the emulated `float16_t`.

Sample errors:

```
sort_test.cc:298:12: error: no matching function for call to 'hwy::float16_t::float16_t(<brace-enclosed initializer list>)'
sort_test.cc:326:34: error: no match for 'operator-' (operand type is 'const hwy::float16_t')
```

This currently fails on master (e.g. the CI run for 16933bd2) and on PRs based on it, in: GCC 9–12, Clang 11–15 (incl. x86_32), armv7, ppc64le, GCC-14 on LoongArch64, and the Windows LLVM (C++20) Meson build — i.e. everywhere `HWY_HAVE_FLOAT16` is 0. Newer compilers with native `_Float16` are unaffected.

Possibly `ConvertScalarTo<T>` for the iota seed and `ScalarNeg`/comparing via bit pattern (or `ConvertScalarTo<T>(-inf)`) for line 326 would keep it portable.

dcsid:
`sort_test.cc` fails to build where `hwy::float16_t` is emulated: `std::iota` with `T{0}` needs construction from `int`, and `-kInf` needs unary minus (GCC 9–12, Clang 11–15, armv7, ppc64le, LoongArch64, Windows LLVM). Fixes #3112.

- Replace the iota with `ConvertScalarTo<T>`, and `x == kInf || x == -kInf` with `ScalarAbs(x) == kInf`.
- Guard the f16 instantiation with `HWY_HAVE_FLOAT16` (as in `CallAllSortTraits`): f16 `GetLane` does not compile on LSX (noted in #3112).

Errors reproduce under `-DHWY_HAVE_SCALAR_F16_TYPE=0` and compile cleanly with the fix; `sort_test --gtest_filter='*SelectWithNaN*'` passes on AArch64.

### DIFF
--- a/hwy/contrib/sort/sort_test.cc
+++ b/hwy/contrib/sort/sort_test.cc
@@ -295,7 +295,10 @@ void TestSelectWithNaNForType() {
   const size_t k = num / 2;
 
   std::vector<T> keys(num);
-  std::iota(keys.begin(), keys.end(), T{0});
+  // Not std::iota: emulated float16_t lacks construction from int.
+  for (size_t i = 0; i < num; ++i) {
+    keys[i] = ConvertScalarTo<T>(i);
+  }
   std::mt19937_64 rng(123456789);
   std::shuffle(keys.begin(), keys.end(), rng);
 
@@ -323,7 +326,7 @@ void TestSelectWithNaNForType() {
   for (T x : keys) {
     if (x != x) {
       ++num_nan;
-    } else if (x == kInf || x == -kInf) {
+    } else if (ScalarAbs(x) == kInf) {  // emulated float16_t lacks unary minus
       ++num_inf;
     } else {
       finite_out.push_back(x);


jan-wassenberg (commented):
Thanks :) FYI #3109 fixes the missing #if. Let's wait for that one to land, then merge.

I'm curious which compiler you're targeting? HWY_HAVE_SCALAR_F16_OPERATORS is not universal, but ought to be widely available. Perhaps we can get that to 1, or maybe we should extend the emulation of hwy::float16_t.

jan-wassenberg (approved):

dcsid:
Rebased onto #3109, now just the `ConvertScalarTo`/`ScalarAbs` constructs.

No particular compiler, just hit it on the CI matrix. FWIW the emulated `float16_t` gaps were construction from `int` (`std::iota`'s `T{0}`) and unary minus (`-kInf`); covering those via `HWY_HAVE_SCALAR_F16_OPERATORS` or the emulation would drop the need for the guard.

PR #3117, opened 2026-06-14
Support value construction, unary minus, and increment on emulated float16_t

dcsid:
Extends the emulated `hwy::float16_t` (the `uint16_t`-backed type used when there is no native `_Float16`, e.g. PowerPC / older compilers) so generic code like `std::iota(v.begin(), v.end(), T{0})` and unary minus compile on those targets. Follows up the discussion in #3112.

Added for the emulated build only (mirroring the native operators):
- converting constructor from any value static-castable to `double` (rounds via `F16FromF64`)
- unary minus (exact sign-bit flip) and unary plus
- pre/post `++` / `--` (via an `F32` round-trip)

All `constexpr` where the converters are; the native path is untouched (everything is behind `#if !HWY_HAVE_SCALAR_F16_TYPE`).

Notes:
- Comparison operators already work for emulated `float16_t`, so this set is enough for `std::iota`, fill, negate, and sort-by-value.
- Binary `+ - * /` are intentionally omitted (Highway converts to `float` before scalar f16 arithmetic).
- `bfloat16_t` has the same gap — happy to mirror it in a follow-up, or fold it in here if you'd prefer parity.

Tested on AArch64 (native) and by forcing `-DHWY_HAVE_SCALAR_F16_TYPE=0` (the emulated path), where `std::iota` over `float16_t` now compiles and runs. `base_test` now covers construction / unary minus / increment for `float16_t` on all targets.

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -1292,6 +1292,24 @@ struct alignas(2) float16_t {
   static constexpr float16_t FromBits(uint16_t bits) {
     return float16_t(F16FromU16BitsTag(), bits);
   }
+
+  // Emulated scalar operators so generic code (e.g. std::iota, unary minus)
+  // also compiles when there is no native _Float16. Construction and
+  // increment/decrement use ConvertScalarTo, defined after the struct, so they
+  // are declared here and defined out-of-class below.
+  template <typename T, HWY_IF_NOT_F16(T)>
+  explicit constexpr float16_t(T arg);
+
+  // Unary minus is an exact sign-bit flip; no conversion needed.
+  constexpr float16_t operator-() const noexcept {
+    return FromBits(static_cast<uint16_t>(bits ^ uint16_t{0x8000}));
+  }
+  constexpr float16_t operator+() const noexcept { return *this; }
+
+  HWY_CXX14_CONSTEXPR float16_t& operator++() noexcept;
+  HWY_CXX14_CONSTEXPR float16_t operator++(int) noexcept;
+  HWY_CXX14_CONSTEXPR float16_t& operator--() noexcept;
+  HWY_CXX14_CONSTEXPR float16_t operator--(int) noexcept;
 #endif  // HWY_HAVE_SCALAR_F16_TYPE
 
   // When backed by a native type, ensure the wrapper behaves like the native
@@ -1701,11 +1719,13 @@ HWY_F16_CONSTEXPR inline std::partial_ordering operator<=>(
 #endif  // HWY_SSE2_HAVE_SCALAR_BF16_TYPE
 
 // Compiler supports __bf16, not necessarily with operators.
+#ifndef HWY_HAVE_SCALAR_BF16_TYPE  // allow override
 #if HWY_ARM_HAVE_SCALAR_BF16_TYPE || HWY_SSE2_HAVE_SCALAR_BF16_TYPE
 #define HWY_HAVE_SCALAR_BF16_TYPE 1
 #else
 #define HWY_HAVE_SCALAR_BF16_TYPE 0
 #endif
+#endif  // HWY_HAVE_SCALAR_BF16_TYPE
 
 #ifndef HWY_HAVE_SCALAR_BF16_OPERATORS
 // Recent enough compiler also has operators. aarch64 clang 18 hits internal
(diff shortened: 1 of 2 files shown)


hwy/base.h
```diff
@@ -1292,6 +1292,26 @@ struct alignas(2) float16_t {
   static constexpr float16_t FromBits(uint16_t bits) {
     return float16_t(F16FromU16BitsTag(), bits);
   }
+
+  // Emulated scalar operators so generic code (e.g. std::iota, unary minus)
+  // also compiles when there is no native _Float16. Construction and
+  // increment round through F16FromF64/F32FromF16, which are defined after the
```
jan-wassenberg: We can use ConvertScalarTo instead of having to specify the F32/F64 type, ConvertScalarTo can convert anything to anything.

hwy/base.h
```diff
@@ -1292,6 +1292,26 @@ struct alignas(2) float16_t {
...
+  // struct, so they are declared here and defined out-of-class below.
+  template <typename T,
+            hwy::EnableIf<!IsSame<RemoveCvRef<T>, float16_t>() &&
```
jan-wassenberg: The first condition can be simplified to `HWY_IF_F16(T)`.

hwy/base.h
```diff
@@ -1623,6 +1643,40 @@ HWY_API HWY_F16_CONSTEXPR float16_t F16FromF64(double f64) {
 #endif
 }
 
+#if !HWY_HAVE_SCALAR_F16_TYPE
+// Out-of-class definitions for the emulated float16_t operators declared in the
+// struct, placed after F16FromF32/F16FromF64/F32FromF16 so those converters are
+// visible. Construction routes through F64 (a single rounding for any input);
```
jan-wassenberg: We could just use ConvertScalarTo so no need to route through F64, which is unavailable on Armv7.

hwy/base.h
```diff
@@ -1623,6 +1643,40 @@ HWY_API HWY_F16_CONSTEXPR float16_t F16FromF64(double f64) {
...
+// increment/decrement round-trip through F32.
+template <typename T, hwy::EnableIf<!IsSame<RemoveCvRef<T>, float16_t>() &&
```
jan-wassenberg: Also HWY_IF_F16(T)?

hwy/base_test.cc
```diff
@@ -846,6 +846,26 @@ struct TestSpecialFloat {
             hwy::EnableIf<!EnableSpecialFloatArithOpTest<T>()>* = nullptr>
   static HWY_INLINE void TestSpecialFloatArithOperators(T /*unused*/) {}
 
+  // Value construction, unary minus, and increment/decrement work for float16_t
+  // on every target, including those that emulate it with a uint16_t (where
+  // these previously failed to compile). Restricted to float16_t because the
+  // emulated bfloat16_t does not yet provide them (left as a follow-up).
+  template <class T, hwy::EnableIf<IsSame<T, float16_t>()>* = nullptr>
```
jan-wassenberg: Also HWY_IF_F16?

hwy/base_test.cc
```diff
@@ -846,6 +846,26 @@ struct TestSpecialFloat {
...
+  static HWY_NOINLINE void TestFloat16ScalarOps(T /*unused*/) {
+    HWY_ASSERT_EQ(2.0f, ConvertScalarTo<float>(T(2)));    // value ctor from int
+    HWY_ASSERT_EQ(2.5f, ConvertScalarTo<float>(T(2.5)));  // value ctor from double
+    HWY_ASSERT_EQ(-3.0f, ConvertScalarTo<float>(-T(3)));  // unary minus
+    T inc = T(0);
```
jan-wassenberg: I'd recommend starting with T(Unpredictable1()), then PreventElision(inc) at the end to ensure the compiler does not optimize this out, and basically only test its elided version, which is not always the same as the code it would/should have generated. Same for dec.

jan-wassenberg (changes requested):
Nice, thanks for adding this!
Would be great to also have this for BF16 for symmetry if you're up for it?

dcsid:
Implemented all of your suggestions: `ConvertScalarTo` for construction and `++`/`--`, `HWY_IF_NOT_F16`/`HWY_IF_NOT_BF16`, `Unpredictable1()`/`PreventElision` in the test, and the same set for `bfloat16_t`.

Thanks for the review :)

hwy/base.h
```diff
@@ -2719,6 +2753,61 @@ HWY_API constexpr TTo ConvertScalarTo(TFrom in) {
           static_cast<TFrom&&>(in)));
 }
 
+#if !HWY_HAVE_SCALAR_F16_TYPE
+// Emulated float16_t operators (declared in the struct). ConvertScalarTo
+// converts any scalar to/from f16 without spelling out the F32/F64 converter.
+template <typename T,
+          hwy::EnableIf<!hwy::IsSame<hwy::RemoveCvRef<T>, hwy::float16_t>()>*>
```
jan-wassenberg, line 2760: Also HWY_IF_NOT_F16?

hwy/base.h
```diff
@@ -2719,6 +2753,61 @@ HWY_API constexpr TTo ConvertScalarTo(TFrom in) {
...
+  const float16_t result = *this;
+  --*this;
+  return result;
+}
+#endif  // !HWY_HAVE_SCALAR_F16_TYPE
+
+#if !HWY_HAVE_SCALAR_BF16_TYPE
+// Emulated bfloat16_t operators (mirrors float16_t).
+template <typename T,
+          hwy::EnableIf<!hwy::IsSame<hwy::RemoveCvRef<T>, hwy::bfloat16_t>()>*>
```
jan-wassenberg, line 2787: And HWY_IF_NOT_BF16?

jan-wassenberg (commented):
Nice, tests look great, thanks for updating! One more nit if you don't mind?

dcsid:
Good call to check, I actually tried that. The definitions can't take the macro though: it expands to `... = nullptr`, and that default is already on the in-class declaration, so repeating it isn't allowed.

I also tried moving them in-class so the macro only appears once, but that breaks `constexpr`: `ConvertScalarTo` is defined after the struct, and a `constexpr` body can't call a function that's only forward-declared at that point.

So I kept the macro on the declarations and spelled out the matching `EnableIf` on the definitions, the same way the out-of-class comparison operators just above already do. Happy to revisit if you'd prefer a different structure!

jan-wassenberg (approved):
Oh, tricky, thanks for explaining. Seems worth a comment for future such cases :)

PR #3118, opened 2026-06-14
ci: pin actions to full commit SHAs in meson and multiarch workflows

XananasX7:
## Summary

Two workflow files (`meson_build_test.yml`, `multiarch.yml`) used mutable action tag references that can be silently re-pointed to different—potentially malicious—commits.

### Changes

| Workflow | Action | Before | After |
|----------|--------|--------|-------|
| both | `actions/checkout` | `@v6.0.3` | `@9f698171ed81b15d1823a05fc7211befd50c8ae0` |
| both | `uraimo/run-on-arch-action` | `@v3.1.0` | `@f9b26e3a1a408d5fd530d20c17b9f3f4428ff8d9` |

Original tags kept as comments. The other two workflows (`build_test.yml`, `docs_pages_workflow.yml`) already use pinned SHAs and are unchanged.

Qualifies under the [Google Open Source Patch Rewards Program](https://bughunters.google.com/open-source-security/patch-rewards) for proactive CI/supply-chain security hardening.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -130,7 +130,7 @@ jobs:
         with:
           egress-policy: audit  # cannot be block - runner does git checkout
 
-      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v4.0.0
+      - uses: actions/checkout@9f698171ed81b15d1823a05fc7211befd50c8ae0 # v6.0.3
 
       - name: Install deps
         run: sudo apt-get update && sudo apt-get install ${{ matrix.extra_deps }}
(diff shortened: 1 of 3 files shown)


jan-wassenberg (commented):
Thanks, pinning makes sense. 
It's interesting that #3074 set df4cb1c069e1874edd31b4311f1884172cec0e10 instead of the new 9f698171ed81b15d1823a05fc7211befd50c8ae0, and also didn't update some of the version comments from 4.0.0 to 6.0.3.
Would you mind harmonizing those (using the same hash everywhere - your choice of which) and updating also the incorrect previous 4.0.0 comments, in the hope that dependabot then updates them in future?

XananasX7:
Done! I've pushed a follow-up commit that harmonizes everything:

- Replaced all remaining `df4cb1c069e1874edd31b4311f1884172cec0e10 # v4.0.0` references with `9f698171ed81b15d1823a05fc7211befd50c8ae0 # v6.0.3` across `build_test.yml`, `meson_build_test.yml`, and `multiarch.yml`
- All 11 `actions/checkout` references across all four workflow files now use the same SHA (`9f698171`) with the correct `# v6.0.3` annotation

(For context: `df4cb1c` was actually the v6.0.3 release commit too, just mislabelled — probably a copy-paste from another PR. Either way, everything is consistent now and Dependabot should be able to pick them up cleanly.)

Thanks for catching that!

jan-wassenberg (approved):
Nice, thanks for updating!

PR #3121, opened 2026-06-16
Add `HWY_ENABLE_LIST_TARGETS` to control building `hwy_list_targets`

phongn:
In #3076, the `hwy_list_targets` target in CMake was altered to only build if it were a top-level project. Some CI systems, notably libjxl, relied on the output of that target automatically running even if not top-level. 

This PR introduces a `HWY_ENABLE_LIST_TARGETS` option to override this new default behavior. For CMake < 3.21, it also sets `PROJECT_IS_TOP_LEVEL` and `<PROJECT-NAME>_IS_TOP_LEVEL` for convenience. 

Tested on:
* EL9's CMake-3.26.5
* Kitware's static-distribution CMake-3.10.3
* Kitware's static-distribution CMake-4.3.3


jan-wassenberg (approved):
LGTM, thanks for adding! @eustas FYI

PR #3123, opened 2026-06-16
Add macOS support to `GetCpuString`

phongn:
While testing #3121 on a few different platforms I noticed that no CPU name was printed by `GetCpuString` on macOS. It might be useful for some folk to see it in various testing or CI harnesses. 

Unfortunately, the broader aarch64 ISA does not include a brand lookup like x64 does; it requires lookup tables or reading system-specific APIs (e.g. parsing `CIM_Processor` or registry entries on Windows; parsing `/proc/cpuinfo` on Linux, etc.).

### DIFF
--- a/hwy/timer.cc
+++ b/hwy/timer.cc
@@ -20,6 +20,10 @@
 #include <chrono>  // NOLINT
 #include <ratio>   // NOLINT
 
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Thanks for adding :)

Issue #3098, opened 2026-06-09
VQSelect corrupts float arrays containing NaN


dmatth1:
`VQSelect` on a float/double array that contains NaN returns a non-permutation of
  the input because the `+inf` sentinel leaks in and real values are lost. `VQSort` on the
  same data is correct, and `b[k]` is usually still correct so it hides from spot-checks.

  Cause: `Select()` (vqsort-inl.h) replaces NaN with the `+inf` sentinel, runs the
  *partial* select (which doesn't move sentinels to the back like a full sort would),
  then blindly overwrites the last `num_nan` slots with NaN — so sentinels leak and
  tail values are destroyed.

  Failing regression test + standalone repro (branch off current master, reproduces on aarch64/NEON):
  https://github.com/dmatth1/highway/tree/vqselect-nan-corruption

  ```
  cmake -B cmbuild -DHWY_ENABLE_TESTS=ON -DHWY_TEST_STANDALONE=ON
  cmake --build cmbuild --target sort_test
  ./cmbuild/tests/sort_test --gtest_filter='*TestSelectWithNaN*'
  # -> Abort: VQSelect leaked 10 inf into a NaN array
  ```

Discovered while working on wiring `VQSelect` into Numpy for np.partition - https://github.com/numpy/numpy/pull/31506

dmatth1:
Thanks for the quick fix! 

One follow-up edge case I think remains: ReplaceSentinelWithNaN identifies sentinels by value (Eq(v, st.LastValue)), but ascending sorts the sentinel is +inf, which is also a legal input value. So when an array contains real +inf together with NaN, the scan can't tell a sentinel from a genuine infinity. After a partial select that leaves them interleaved, the wrong slots get turned into NaN.

Concrete failure on current master (n=20000, double, 16 real +inf + 8 NaN, ascending):

k=19999: selected element returns +inf, expected NaN   <-- FAIL
k=19996: selected element returns +inf, expected NaN   <-- FAIL

The NaN count and the finite multiset are preserved, so a count/multiset check passes — it only shows up in keys[k] when k falls in the NaN region. (Arrays without real infinities are unaffected, so the merged fix is correct for the common case.)

One way to sidestep the value-collision entirely: move NaNs to the back up front, detecting them with IsNaN (which never confuses +inf with NaN) rather than by sentinel value, then run the partial op on the non-NaN prefix — k in the NaN region then needs no work.

jan-wassenberg:
:) Good catch, this is tricky.
[...]

PR #3129, opened 2026-06-18
Fix VQSelect/PartialSort NaN handling for arrays containing real inf

dmatth1:
Follow-up to #3104 (fixes the remaining case in #3098).
  
  **Bug.** #3104 restores NaN after a partial select/sort by scanning for the `st.LastValue` sentinel (`Eq(v, sentinel)`). But that sentinel is `+inf` for ascending so arrays containing real `+inf` alongside NaN still
  corrupt and the wrong slots become NaN. E.g. n=20000 with real `+inf` and `k` in the NaN-rank region, `VQSelect` returns `+inf` where the NaN-last order requires NaN.

  **Fix.** Replace the sentinel scheme on the partial paths with `detail::PartitionNaNToBack` (per @jan-wassenberg's suggestion) - one pass that `CompressStore`s the non-NaN lanes to the front, counts NaN via the mask, fills the tail with
  canonical NaN, then runs the partial op on the prefix `[0, num - num_nan)` (clamping `k`). Detection is by `IsNaN` so there's no sentinel and no `+inf` collision. `Sort()` is unchanged, `ReplaceSentinelWithNaN` is removed.

  **Tests.** `TestSelectWithNaN` now covers `VQSelect` and `VQPartialSort` × ascending/descending × float/double/float16 (runtime-gated) with NaN and real `+inf` present, checking `keys[k]`/sorted prefix and the non-NaN multiset. f16-safe
  via `ConvertScalarTo`/`ScalarIsNaN`/`ScalarIsInf`.

  Verified on aarch64 (NEON, NEON_WITHOUT_AES, EMU128, `HaveFloat16()=1`). The new test fails against the #3104 code and passes here

### DIFF
--- a/hwy/contrib/sort/sort_test.cc
+++ b/hwy/contrib/sort/sort_test.cc
@@ -286,80 +286,160 @@ void TestPartialSortKEqualsN() {
   }
 }
 
-// VQSelect on a floating-point array containing NaN must still leave a
-// permutation of the input (NaNs allowed anywhere): no value may be created or
-// lost.
+// Shuffled finite values (within float16_t's exact range to avoid overflow to
+// inf), with a few NaN and a few real +inf at spread-out positions. The +inf
+// is the case a by-value sentinel scan confuses with NaN. Returns the NaN count.
 template <typename T>
-void TestSelectWithNaNForType() {
-  const size_t num = AdjustedReps(100000);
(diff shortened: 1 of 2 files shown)


jan-wassenberg (approved):
Nice work, thanks for the thorough test and fix!

PR #3140, opened 2026-06-19
Add RVV overrides for [Rounding]ShiftRightAndDemoteTo

eyupcanakman:
Implements the RVV override for `ShiftRightAndDemoteTo` and `RoundingShiftRightAndDemoteTo`, following #3056 for #2628. Same-sign narrowing uses one fused `vnclip`/`vnclipu` (RDN, or RNU for the rounding variant) instead of a separate shift then narrow; other cases fall back to `DemoteTo(d, [Rounding]ShiftRight<k>(v))`.

CI has no RVV target, so I checked under qemu (rv64gcv, clang-20) and results match the generic path at VLEN 128 and 256; gcc-14 builds clean. SVE2/LSX/LASX stay TODO.


jan-wassenberg (approved):
Nice :) Thanks for adding.

PR #3141, opened 2026-06-20
Add tgamma

RaviTriv:
Implementing `tgamma` from [wishlist](https://github.com/google/highway/blob/master/g3doc/op_wishlist.md?plain=1#L42).

Thanks for time and effort reviewing :)

### DIFF
--- a/g3doc/op_wishlist.md
+++ b/g3doc/op_wishlist.md
@@ -37,7 +37,7 @@ _mm512_getmant (f32/f64)
 
 High-precision! Consider copying from SLEEF. See #1650.
 
-fmod, ilogb, lgamma, logb, modf, nextafter, nexttoward, scalbn, tgamma
+fmod, ilogb, lgamma, logb, modf, nextafter, nexttoward, scalbn
 
 ### Remaining STL functions for hwy/contrib/algo
 
@@ -194,3 +194,4 @@ For SVE (svld1sb_u32)+WASM? Compiler can probably already fuse.
 *   ~~tan~~
 *   ~~pow~~
 *   ~~Lookup32~~
+*   ~~tgamaa~~
--- a/g3doc/op_wishlist.md
+++ b/g3doc/op_wishlist.md
@@ -194,4 +194,4 @@ For SVE (svld1sb_u32)+WASM? Compiler can probably already fuse.
 *   ~~tan~~
 *   ~~pow~~
 *   ~~Lookup32~~
-*   ~~tgamaa~~
+*   ~~tgamma~~
--- a/hwy/contrib/math/fp_arith-inl.h
+++ b/hwy/contrib/math/fp_arith-inl.h
@@ -192,7 +192,7 @@ HWY_INLINE HWY_MAYBE_UNUSED VF DDAdd(DF df, VF a_hi, VF a_lo, VF b_hi, VF b_lo,
 
 // Returns hi = (a_hi + a_lo) * b, r_lo = low part
 template <class DF, HWY_IF_FLOAT3264_D(DF), class VF = Vec<DF>>
-HWY_INLINE HWY_MAYBE_UNUSED VF DDMulV(DF df, VF a_hi, VF a_lo, VF b, VF& r_lo) {
+HWY_INLINE HWY_MAYBE_UNUSED VF DDMul1(DF df, VF a_hi, VF a_lo, VF b, VF& r_lo) {
   VF p_lo;
   const VF p_hi = TwoProducts(df, a_hi, b, p_lo);
   p_lo = MulAdd(a_lo, b, p_lo);
@@ -201,8 +201,8 @@ HWY_INLINE HWY_MAYBE_UNUSED VF DDMulV(DF df, VF a_hi, VF a_lo, VF b, VF& r_lo) {
 
 // Returns hi = (a_hi + a_lo) * (b_hi + b_lo), r_lo = low part
 template <class DF, HWY_IF_FLOAT3264_D(DF), class VF = Vec<DF>>
-HWY_INLINE HWY_MAYBE_UNUSED VF DDMul(DF df, VF a_hi, VF a_lo, VF b_hi, VF b_lo,
-                                     VF& r_lo) {
+HWY_INLINE HWY_MAYBE_UNUSED VF DDMul2(DF df, VF a_hi, VF a_lo, VF b_hi, VF b_lo,
+                                      VF& r_lo) {
   VF p_lo;
   const VF p_hi = TwoProducts(df, a_hi, b_hi, p_lo);
   p_lo = MulAdd(a_hi, b_lo, MulAdd(a_lo, b_hi, p_lo));
@@ -215,7 +215,7 @@ HWY_INLINE HWY_MAYBE_UNUSED VF DDDiv(DF df, VF a_hi, VF a_lo, VF b_hi, VF b_lo,
                                      VF& r_lo) {
   const VF q1 = Div(a_hi, b_hi);
   VF qb_lo;
-  const VF qb_hi = DDMulV(df, b_hi, b_lo, q1, qb_lo);
+  const VF qb_hi = DDMul1(df, b_hi, b_lo, q1, qb_lo);
   VF r_hi_lo;
   const VF r_hi = DDAdd(df, a_hi, a_lo, Neg(qb_hi), Neg(qb_lo), r_hi_lo);
   const VF q2 = Div(r_hi, b_hi);
--- a/hwy/contrib/math/fp_arith-inl.h
+++ b/hwy/contrib/math/fp_arith-inl.h
@@ -30,6 +30,13 @@
 
 #include "hwy/highway.h"
 
+// Disable FMA contraction so each mul/add in the error-free transforms rounds
+// on its own.
+#if HWY_COMPILER_GCC && !HWY_COMPILER_CLANG
+#pragma GCC push_options
+#pragma GCC optimize("fp-contract=off")
+#endif
+
 HWY_BEFORE_NAMESPACE();
 namespace hwy {
 namespace HWY_NAMESPACE {
@@ -226,5 +233,9 @@ HWY_INLINE HWY_MAYBE_UNUSED VF DDDiv(DF df, VF a_hi, VF a_lo, VF b_hi, VF b_lo,
 }  // namespace HWY_NAMESPACE
 }  // namespace hwy
 HWY_AFTER_NAMESPACE();
+
+#if HWY_COMPILER_GCC && !HWY_COMPILER_CLANG
+#pragma GCC pop_options
+#endif
 
 #endif  // NOLINT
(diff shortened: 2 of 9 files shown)


hwy/contrib/math/math-inl.h
```diff
@@ -705,6 +720,74 @@ HWY_INLINE HWY_MAYBE_UNUSED V InRangeFloatDivRem(V a, V b, V q) {
 #endif
 }
 
+// Double-double helpers
+// One value stored in two doubles (hi + lo) for ~2x precision, lo is hi's
+// rounding error.
+
+// Returns s = a + b, lo = exact rounding error
```
jan-wassenberg: We have very similar code in https://github.com/google/gemma.cpp/blob/main/ops/fp_arith-inl.h .
Want to copy that code exactly? In gemma.cpp we could then also use the new copy which resides in Highway. Maybe put it in contrib/math/fp_arith-inl.h?

hwy/contrib/math/math-inl.h
```diff
@@ -1591,6 +1761,117 @@ HWY_INLINE V ExtPrecLog2ForPow(D d, V x, V& log2_x_lo) {
...
+  hi = DDAdd(hi, lo, Neg(w), kZero, lo);
+  hi = DDAdd(hi, lo, kHalfLn2PiHi, kHalfLn2PiLo, lo);
+  const V series = Mul(inv_w, impl.StirlingPoly(d, u));
+  return DDAdd(hi, lo, series, kZero, lo);
+}
+
+// Computes signed Gamma(a). For a < 0.5 uses w = 1 - a; then a [2, 3)
+// polynomial below kStirlingLimit, Stirling above.
+template <class D, class V = VFromD<D>>
+HWY_INLINE V gammak(D d, V a) {
```
jan-wassenberg: Name is unusual, any reason not to just call it Gamma? Or at least in PascalCase.

hwy/contrib/math/math-inl.h
```diff
@@ -1591,6 +1761,117 @@ HWY_INLINE V ExtPrecLog2ForPow(D d, V x, V& log2_x_lo) {
...
+  V num_hi = kOne, num_lo = kZero;
+  V den_hi = kOne, den_lo = kZero;
+
+  for (int i = 0; i < 2; ++i) {
+    const auto up = Lt(wa, kTwo);
+    V d_lo;
+    const V d_hi = DDMulV(den_hi, den_lo, wa, d_lo);
+    den_hi = IfThenElse(up, d_hi, den_hi);
+    den_lo = IfThenElse(up, d_lo, den_lo);
+    wa = IfThenElse(up, Add(wa, kOne), wa);
```
jan-wassenberg: Can be MaskedAddOr(wa, up, wa, kOne).

hwy/contrib/math/math-inl.h
```diff
@@ -1591,6 +1761,117 @@ HWY_INLINE V ExtPrecLog2ForPow(D d, V x, V& log2_x_lo) {
...
+  const auto neg = Lt(a, kHalf);
+  V w = IfThenElse(neg, Sub(kOne, a), a);
+
+  // Shift w into [2, 3) via Gamma(x+1) = x*Gamma(x).
+  V wa = w;
+  V num_hi = kOne, num_lo = kZero;
+  V den_hi = kOne, den_lo = kZero;
+
+  for (int i = 0; i < 2; ++i) {
+    const auto up = Lt(wa, kTwo);
```
jan-wassenberg: Prefer to minimize auto. We can set up an M typedef like we do for V: class M = MFromD<D>.

hwy/contrib/math/math-inl.h
```diff
@@ -1591,6 +1761,117 @@ HWY_INLINE V ExtPrecLog2ForPow(D d, V x, V& log2_x_lo) {
...
+  const V series = Mul(inv_w, impl.StirlingPoly(d, u));
+  return DDAdd(hi, lo, series, kZero, lo);
+}
+
+// Computes signed Gamma(a). For a < 0.5 uses w = 1 - a; then a [2, 3)
+// polynomial below kStirlingLimit, Stirling above.
+template <class D, class V = VFromD<D>>
+HWY_INLINE V gammak(D d, V a) {
+  using T = TFromD<D>;
+  constexpr bool kIsF32 = (sizeof(T) == 4);
```
jan-wassenberg: We should also static_assert IsFloat<T>.

hwy/contrib/math/math-inl.h
```diff
@@ -1591,6 +1761,117 @@ HWY_INLINE V ExtPrecLog2ForPow(D d, V x, V& log2_x_lo) {
...
+  const V gamma_w = IfThenElse(Lt(w, kStirlingLimit), gamma_a, gamma_b);
+
+  // For a < 0.5: Gamma(a) = pi / (sin(pi*a) * Gamma(w)).
+  const V ra = Round(a);
+  const V frac = Sub(a, ra);
+  const V s_mag = Sin(d, Mul(kPi, frac));
+  const RebindToSigned<decltype(d)> di;
+  const auto odd =
+      RebindMask(d, Ne(And(ConvertTo(di, ra), Set(di, 1)), Zero(di)));
+  const V s_signed = IfThenElse(odd, Neg(s_mag), s_mag);
```
jan-wassenberg, line 1801: We will soon land MaskedXorOr which could save an instruction here: s_signed = MaskedXorOr(s_mag, odd, s_mag, kSign);

jan-wassenberg (changes requested):
Nice work :) A few minor suggestions, the core is solid:

hwy/contrib/math/math-inl.h
RaviTriv: Nope, updated to `Gamma`

hwy/contrib/math/math-inl.h
RaviTriv: cool! definitely, taking a closer look and will update :)

hwy/contrib/math/math-inl.h
RaviTriv: Nice, will update this shortly

g3doc/op_wishlist.md
```diff
@@ -195,3 +195,4 @@ For SVE (svld1sb_u32)+WASM? Compiler can probably already fuse.
 *   ~~cosh~~
 *   ~~tan~~
 *   ~~pow~~
+*   ~~tgamaa~~
```
jan-wassenberg: Typo: tgamma :)

hwy/contrib/math/fp_arith-inl.h
```diff
@@ -0,0 +1,230 @@
...
+                                     VF& r_lo) {
+  VF e;
+  const VF s = TwoSums(df, a_hi, b_hi, e);
+  e = Add(e, Add(a_lo, b_lo));
+  return FastTwoSums(df, s, e, r_lo);
+}
+
+// Returns hi = (a_hi + a_lo) * b, r_lo = low part
+template <class DF, HWY_IF_FLOAT3264_D(DF), class VF = Vec<DF>>
+HWY_INLINE HWY_MAYBE_UNUSED VF DDMulV(DF df, VF a_hi, VF a_lo, VF b, VF& r_lo) {
```
jan-wassenberg: The V in the name isn't clear to me. Maybe DDMul1 and DDMul2 would be more clear?

jan-wassenberg (commented):
Nice, thanks for updating. Just two more nits:

hwy/contrib/math/math-inl.h
jan-wassenberg: We can also do this in a follow-up.

jan-wassenberg (approved):

jan-wassenberg (changes requested):
Looks like tests are still failing :)

hwy/contrib/math/fp_arith-inl.h
```diff
@@ -0,0 +1,223 @@
...
+    const VF b_hi = BitCast(
+        df, ShiftLeft<kHalf>(RoundingShiftRight<kHalf>(BitCast(du, b))));
+    const VF a_lo = Sub(a, a_hi);
+    const VF b_lo = Sub(b, b_hi);
+    // Sum the exact partial products into the low part
+    err = MulAdd(
+        a_lo, b_lo,
+        Add(MulAdd(a_hi, b_lo, Mul(a_lo, b_hi)), MulSub(a_hi, b_hi, prod)));
+  }
+  return prod;
```
RaviTriv, line 81: There was an issue with the fp contraction optimization on targets without FMA. In the original `TwoProducts` the `mul` + `sub` was getting fused so the intermediate rounding was being skipped. Made an update to the non fma path to use bit shifts to split each input into half-mantissa hi/lo parts so that every partial product is exact.

RaviTriv:
> Looks like tests are still failing :)

Thanks for taking a look! Pushed some updates, should be almost green except for `f64` on LoongArch.
- Loosened to 6 ULP on `f32` and 8 ULP on `f64`
- LoongArch is hitting 32 ULP on `f64` 😅, investigating some more.

jan-wassenberg (dismissed):
Nice work :)

jan-wassenberg:
CI is flagging an issue: this conflicts with functions of the same name in gemma.cpp, in particular UpdateCascadedSums etc. Let's rename them to avoid the conflict (namespace is insufficient because some of the args are in the Highway namespace, hence trigger ADL). Any preference? Could be CascadedSums_Update/Reduce etc, or better yet a CascadedSums class with static member functions, or do you have some other idea?

RaviTriv:
> CI is flagging an issue: this conflicts with functions of the same name in gemma.cpp, in particular UpdateCascadedSums etc. Let's rename them to avoid the conflict (namespace is insufficient because some of the args are in the Highway namespace, hence trigger ADL). Any preference? Could be CascadedSums_Update/Reduce etc, or better yet a CascadedSums class with static member functions, or do you have some other idea?

Good catch! Maybe a nested namespace wrapping the whole file so everything in the file is included, something like `compensated::UpdateCascadedSums`?
- Open to whatever name makes sense :) 
- Also good with the static class approach if that is preferable

jan-wassenberg:
hm, not everything in that file is "compensated" - that's just the summation, the file also defines TwoProducts/TwoSums. The class would still be reasonable but we'd also have to rename the other functions (TwoSums). Instead I will update gemma.cpp so this problem goes away, then we can land this as-is.

jan-wassenberg:
OK, the change has landed. Would you mind rebasing to fix the conflict?

RaviTriv:
> hm, not everything in that file is "compensated" - that's just the summation, the file also defines TwoProducts/TwoSums. The class would still be reasonable but we'd also have to rename the other functions (TwoSums). Instead I will update gemma.cpp so this problem goes away, then we can land this as-is.

good point, thank you for handling this!

jan-wassenberg (approved):

jan-wassenberg:
Almost there, additional fixes were required to gemma.cpp to enable this, which are landing soon.

PR #3143, opened 2026-06-20
Fix DivCeil overflow and reject zero divisors

fallintoplace:
Fixes two edge cases in `hwy/base.h`:

- `DivCeil` used `(a + b - 1) / b`, which can overflow for large `a`. This switches it to the equivalent `a == 0 ? 0 : 1 + (a - 1) / b` form.
- `Divisor` and `Divisor64` accepted zero through the `<= 1` fast path. They now assert on zero and keep the `1` fast path explicit. The `Divisor64` fallback path gets the same check.

Tests:

- `./cmake-out/tests/base_test`
- C++11 and C++17 header compile probes for `DivCeil`
- Header-only `HWY_HAVE_DIV128=0` probe for the `Divisor64` fallback path

### DIFF
--- a/hwy/base.h
+++ b/hwy/base.h
@@ -2822,9 +2822,9 @@ HWY_API uint32_t LemireMod(uint32_t in, uint32_t range) {
 template <typename T1, typename T2>
 constexpr T1 DivCeil(T1 a, T2 b) {
 #if HWY_CXX_LANG >= 201703L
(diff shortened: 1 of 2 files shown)


jan-wassenberg (approved):
Thanks for fixing. I agree it makes sense to assert in DivCeil, it should not be called in time-sensitive code anyway.

PR #3144, opened 2026-06-22
Added GCC 15, GCC 16, Clang 20, Clang 21, and Clang 22 to GitHub workflow

johnplatts:
Resolves one of the issues raised by issue #3022 as the Ubuntu 26.04 GitHub Actions Runner Image was recently released.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -26,40 +26,39 @@ jobs:
     strategy:
       matrix:
         include:
-          - name: Clang-11
-            extra_deps: clang-11
-            c_compiler: clang-11
-            cxx_compiler: clang++-11
-            cxx_standard: 17
-
           - name: Clang-12
             extra_deps: clang-12
             c_compiler: clang-12
             cxx_compiler: clang++-12
+            extra_cmake_flags: -DHWY_ENABLE_CONTRIB=OFF
             cxx_standard: 17
 
           - name: Clang-13
             extra_deps: clang-13
             c_compiler: clang-13
             cxx_compiler: clang++-13
+            extra_cmake_flags: -DHWY_ENABLE_CONTRIB=OFF
             cxx_standard: 17
 
           - name: Clang-14
             extra_deps: clang-14
             c_compiler: clang-14
             cxx_compiler: clang++-14
+            extra_cmake_flags: -DHWY_ENABLE_CONTRIB=OFF
             cxx_standard: 17
 
           - name: Clang-15
             extra_deps: clang-15
             c_compiler: clang-15
             cxx_compiler: clang++-15
+            extra_cmake_flags: -DHWY_ENABLE_CONTRIB=OFF
             cxx_standard: 17
 
           - name: Clang-15 (C++20)
             extra_deps: clang-15
             c_compiler: clang-15
             cxx_compiler: clang++-15
+            extra_cmake_flags: -DHWY_ENABLE_CONTRIB=OFF
             cxx_standard: 20
 
           - name: Clang-15 (x86_32)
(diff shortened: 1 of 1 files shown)


jan-wassenberg (changes requested):
Thanks for updating! I've recently been surprised to see our spend on CI. Would you please suggest some of the other/older matrix elements to retire/remove?

johnplatts:
> Thanks for updating! I've recently been surprised to see our spend on CI. Would you please suggest some of the other/older matrix elements to retire/remove?

Possibly Clang 11 through Clang 13 and GCC 9, but there were some changes in Clang 15 and GCC 12 that added support for the _Float16 type on x86 as long as at least SSE2 is enabled.

In addition, GCC 11 is the default version of GCC for Ubuntu 22.04 and we really want to ensure that Google Highway successfully compiles with GCC 11 on x86_64 as there are still likely some packages with Google Highway dependencies that are still being built for Ubuntu 22.04 with GCC 11.

@jan-wassenberg - Are you using infrastructure other than GitHub workflows to check that Google Highway can successfully be built with some of these earlier GCC/Clang versions?

Should I disable building hwy_contrib (and the associated hwy_contrib_tests) for some of these GCC/Clang compiler versions in the GitHub workflow (which would speed up builds), but still compile hwy_contrib (and the hwy_contrib tests) for at least a few of these GCC/Clang compiler versions?

johnplatts:
> Thanks for updating! I've recently been surprised to see our spend on CI. Would you please suggest some of the other/older matrix elements to retire/remove?

Another change that could be made (which would likely significantly speed up build times and reduce spend on CI) is update the multiarch workflow to use cross compilers targeting armv7 and powerpc64le that run natively on x86-64 (instead of using uraimo/run-on-arch-action, which runs native armv7-linux-gnueabihf or powerpc64le-linux-gnu GCC compilers on x86-64 through QEMU emulation).

bkmgit:
Running on GCP might reduce costs:
- https://github.com/Cyclenerd/google-cloud-github-runner
- https://docs.cloud.google.com/run/docs/tutorials/github-runner

jan-wassenberg:
Clang 11 is 6 years old - I could imagine removing that one, yes.
Agree we should keep GCC 11. GCC 9 is 7 years old and also seems good to remove.

> Are you using infrastructure other than GitHub workflows to check that Google Highway can successfully be built with some of these earlier GCC/Clang versions?

No, our internal CI only uses something close to Clang head.

> Should I disable building hwy_contrib

Yes, please :) Good idea to include it in just one action.

> update the multiarch workflow to use cross compilers

This also sounds good. You mentioned in #3148 that Ubuntu includes cross compilers for many targets - good news.
If we avoid run-on-arch-action, would that work for both GCC and Clang? I believe clang also includes support for all targets.

Would welcome a PR with your proposals.

> Running on GCP might reduce costs:

Given the accounting status quo, we are only able to use default runners :/

johnplatts:
> This also sounds good. You mentioned in #3148 that Ubuntu includes cross compilers for many targets - good news. If we avoid run-on-arch-action, would that work for both GCC and Clang? I believe clang also includes support for all targets.

Yes, the Clang packages in Ubuntu are capable of cross-compilation for LoongArch, Power ISA, RISC-V, and ZSeries (as long as the g++ cross-compiler package for the target is installed).

.github/workflows/build_test.yml
```diff
@@ -202,55 +202,134 @@ jobs:
...
+    name: Build and test ${{ matrix.name }}
+    runs-on: ubuntu-26.04
+    strategy:
+      matrix:
+        include:
+          - name: Clang-20
+            extra_deps: clang-20
+            c_compiler: clang-20
+            cxx_compiler: clang++-20
+            # Disable AVX3_SPR/AVX3_ZEN4 targets in GitHub workflow build with
```
jan-wassenberg: hm, does this issue affect even clang 22? Is this something that we should report instead of just disabling the targets?

.github/workflows/build_test.yml
```diff
@@ -202,55 +202,134 @@ jobs:
...
+            c_compiler: clang-22
+            cxx_compiler: clang++-22
+            # Disable AVX3_SPR/AVX3_ZEN4 targets in GitHub workflow build with
+            # Clang 16 or later to work around crashes that occur with Clang 16
+            # or later when some of the Google Highway tests are compiled with
+            # all x86 targets enabled
+            cxx_flags: -DHWY_DISABLED_TARGETS=0x58
+            cxx_standard: 17
+
+          - name: Clang-22 (VQSort disabled)
```
jan-wassenberg: Maybe instead of clang21 + clang22 + clang22-no-vqsort we can just have clang21-no-vqsort and clang22?

jan-wassenberg (approved):
👍

.github/workflows/build_test.yml
johnplatts: @jan-wassenberg - I am now able to successfully build all non-broken targets (including AVX3_SPR/AVX3_ZEN4) with Clang 20, Clang 21, and Clang 22 without having to disable any targets.

I have updated the build/test workflow to build all non-broken targets with Clang 20, Clang 21, and Clang 22.

Issue #2738, opened 2025-09-29
-march  rv64gcv1p0 does not make sense on RVA23 platforms


xypron:
Ubuntu 25.10 is compiled for rva23u64. This includes many other extensions besides vector that could add up to speed-ups. Therefore it does not make sense to enforce -march  rv64gcv1p0 in file BUILD (cf. #1740).

Building with different parts of highway using different settings of -march currently leads to errors in link time optimization (LTO) on GCC 15.2. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=110812. Unfortunately the GCC patch [a7fe8f83bf89 ("RISC-V: Always register vector built-in functions during LTO [PR110812]"](a7fe8f83bf89704e278c1db30f51d6bb26bf0e69).

Please, provide a solution to compile all of highway using the default GCC profile.

xypron:
Linked Ubuntu bug report: [2121375 FTBFS with GCC 15.2](https://bugs.launchpad.net/ubuntu/+source/highway/+bug/2121375).

jan-wassenberg:
Hi, I agree it's not ideal to hard-code `march`. Highway supports using target attributes and runtime dispatch, which would avoid this problem.

Unfortunately we are not able to enable it due to compiler bugs/limitations:

```
// The riscv_vector.h in Clang 16-18 requires compiler flags, and 19 still has
// some missing intrinsics, see
// https://github.com/llvm/llvm-project/issues/56592. GCC 13.3 also has an
// #error check, whereas 14.1 fails with "argument type 'vuint16m8_t' requires
// the V ISA extension": https://gcc.gnu.org/bugzilla/show_bug.cgi?id=115325.
#if HWY_ARCH_RISCV && HWY_COMPILER_CLANG >= 1900 && 0
#define HWY_HAVE_RUNTIME_DISPATCH_RVV 1
```
Would you like to try enabling that with GCC 15.2 and see if it works there?

xypron:
Building origin/master natively on a rva23s64 VM using GCC 15.2 with rva23u64 extensions enabled by default succeeded using this diff:

```
diff --git a/hwy/detect_targets.h b/hwy/detect_targets.h
index 5671253e..58dd7c0d 100644
--- a/hwy/detect_targets.h
+++ b/hwy/detect_targets.h
@@ -757,7 +757,7 @@
 // https://github.com/llvm/llvm-project/issues/56592. GCC 13.3 also has an
[...]

PR #3148, opened 2026-06-23
ci: add riscv64 to multiarch CI workflow

rajeshgangam:
## Summary

- Add RISC-V 64-bit (`riscv64`) to the foreign architectures CI matrix using the same `run-on-arch-action` pattern as `armv7` and `ppc64le`

Highway has 6,903 lines of RVV code in `hwy/ops/rvv-inl.h` with 448+ operations and 5 RVV-related commits in the last month, but zero CI coverage for riscv64. This means regressions can go undetected.

The existing `install` and `run` steps work unmodified:
- `build-essential` on riscv64 Ubuntu provides GCC 13+ (satisfies `HWY_BROKEN_RVV`)
- CMakeLists.txt auto-detects RISC-V and adds `-march=rv64gcv1p0`
- `libgtest-dev`, `cmake`, `ninja-build` are available in the distro

## Notes

- If build times under QEMU are prohibitive, we can add `cmake_flags: -DHWY_ENABLE_CONTRIB=OFF` to skip vqsort
- The 3 known test failures from issue #2738 should not trigger here since `HWY_HAVE_RUNTIME_DISPATCH_RVV` requires Clang 19+ (this CI uses GCC)

### DIFF
--- a/.github/workflows/multiarch.yml
+++ b/.github/workflows/multiarch.yml
@@ -123,3 +123,36 @@ jobs:
           cmake -DCMAKE_C_COMPILER_TARGET="loongarch64-unknown-linux-gnu" -DCMAKE_CXX_COMPILER_TARGET="loongarch64-unknown-linux-gnu" -DCMAKE_CROSSCOMPILING=true -DCMAKE_CROSSCOMPILING_EMULATOR="qemu-loongarch64;-cpu;max;-L;/opt/cross-tools/target" -DCMAKE_SYSTEM_NAME=Linux -DCMAKE_SYSTEM_PROCESSOR=loongarch64 -B out .
           cmake --build out
           ctest --test-dir out
+
+  riscv64_cmake:
+    name: Build and test GCC on RISC-V 64
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+        with:
+          egress-policy: audit  # cannot be block - runner does git checkout
+
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Install deps
+        run: |
+          sudo apt-get update && sudo apt-get install qemu-user \
+            g++-14-riscv64-linux-gnu
+
+      - name: Build and test
+        run: |
+          export CMAKE_BUILD_PARALLEL_LEVEL=2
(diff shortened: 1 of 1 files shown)


johnplatts:
@jan-wassenberg and @rajeshgangam - I have run builds using GitHub workflows that use GCC cross compilers running natively on x86_64 that target non-x86 ISA's, and using GCC cross compilers running natively on x86_64 are much faster than using run-on-arch-action.

Here is a link to a GitHub workflow run using a workflow that builds Google Highway for non-x86 ISA targets using GCC cross compilers targeting non-x86 ISA's that are running natively on x86_64 (which are included in Ubuntu 26.04):
https://github.com/johnplatts/jep_google_highway/actions/runs/28099376891 (and the workflow used for this build can be found at https://raw.githubusercontent.com/johnplatts/jep_google_highway/e096a6ebb300d58b471cf1ca0d16e26828f9a129/.github/workflows/multiarch.yml).

Here is a link to a GitHub workflow run using a workflow that builds Google Highway for non-x86 ISA targets using run-on-arch-action (that builds Google Highway using GCC compilers that are running through QEMU emulation on an x86_64 system):
https://github.com/johnplatts/jep_google_highway/actions/runs/28104240990/job/83214006485 (and the workflow used for this build can be found at https://raw.githubusercontent.com/johnplatts/jep_google_highway/refs/heads/hwy_multiarch_only_workflow_062426_2/.github/workflows/multiarch.yml).

Using a GCC cross compiler running natively on x86_64 (which are included in Ubuntu for Armv7, AArch64, Alpha, ARC, LoongArch, Motorola 68k, Power ISA (including big-endian 32-bit, little-endian 64-bit, and big-endian 64-bit), RISC-V, SH4, SPARC, and ZSeries) to build Google Highway for non-x86 targets is much faster than using run-on-arch-action.

In addition, run-on-arch-action only supports Armv6, Armv7, AArch64, 64-bit RISC-V, 64-bit little-endian Power ISA, and ZSeries, whereas GCC cross compilers support more architectures.

jan-wassenberg:
I'm curious why we closed the issue? Seems like we could still move ahead with this?

johnplatts:
> I'm curious why we closed the issue? Seems like we could still move ahead with this?

I updated the workflow in pull request #3157 to build for Armv7, big-endian Power ISA, little-endian Power ISA, LoongArch, ZSeries, and RISC-V using both GCC 16 and Clang 22.

jan-wassenberg:
Got it, both are GCC (though 14 here and 16 in yours), thanks.

PR #3151, opened 2026-06-24
Fixes to Cuckoo2x2, PHAST, and ctf_aes example on big-endian

johnplatts:
Made fixes to Cuckoo2x2 and PHAST implementations to fix test failures on big-endian targets.

Also updated hwy/examples/ctf_aes.cc to allow the ctf_aes example to successfully compile on big-endian targets and the HWY_SCALAR target.

All tests pass on big-endian PPC10 with the changes in this pull request.


jan-wassenberg (approved):
Thanks for fixing :)
It's regrettable that every DupEven/Odd will likely require checking HWY_IS_BIG_ENDIAN. Should we deprecate it and replace with DupBottom/Top?

PR #3152, opened 2026-06-24
Made fixes if HWY_TEST_STANDALONE is 1

johnplatts:
Added support for HWY_EXPORT_AND_TEST_BEST_P if HWY_TEST_STANDALONE is 1.

Updated GTestFilterPattern constructor to set min_test_name_len_ to the minimum number of characters that must be matched for a string to match the filter pattern.

Updated GTestFilterPattern::Matches to fix a bug that occurs when a 1-character subpattern is being matched.

Also added HWY_TEST_MAIN to cuckoo2x2_test.cc, cuckoo_test.cc, hash_test.cc, and phast_test.cc to fix build errors if HWY_TEST_STANDALONE is 1.


jan-wassenberg (approved):
Thanks for fixing these :)

PR #3153, opened 2026-06-24
Fix LoadInterleaved3 for Vec64<uint16_t> on LoongArch LSX

wszqkzqk:
LoadInterleaved3 on 64-bit uint16_t vectors produced swapped channels 0 and 2 on LSX, corrupting any caller that relied on correct de-interleaving. This is the root cause of https://github.com/libjxl/libjxl/issues/4709.

Fix the index arrays so the expected lanes are returned.

Tested on Arch Linux for Loong64.


wszqkzqk:
CC @HecaiYuan @jan-wassenberg

jan-wassenberg (approved):
Thanks for fixing, makes sense.

PR #3156, opened 2026-06-25
Added GCC 16, Clang 22, and ppc64le GCC 16 to Meson build/test workflow

johnplatts:
Added GCC 16, Clang 22, and powerpc64le-linux-gnu-g++-16 to Meson build/test workflow.

Also removed run_on_arch from Meson build/test workflow as the powerpc64le-linux-gnu-g++-16 cross compiler running on Ubuntu 26.04 is now used to build for ppc64le in the Meson build/test workflow.


jan-wassenberg (approved):
LGTM, thanks!

PR #3157, opened 2026-06-25
Updated multiarch build/test workflow to use cross compilers

johnplatts:
Updated multiarch build/test workflow to build for Armv7, big-endian Power ISA, little-endian Power ISA, LoongArch, ZSeries, and RISC-V using cross compilers running on x86_64 instead of using run-on-arch-action (which runs compilers built for the target architecture under QEMU emulation) as using cross compilers running on x86_64 is much faster than using run-on-arch-action.

In addition, run-on-arch-action does not support big-endian Power ISA or LoongArch, whereas Ubuntu 26.04 includes cross compilers and QEMU emulators for both big-endian Power ISA and LoongArch.

### DIFF
--- a/.github/workflows/multiarch.yml
+++ b/.github/workflows/multiarch.yml
@@ -13,7 +13,7 @@ jobs:
     name: Multiarch build and test for ${{ matrix.arch }} with GCC 16
     runs-on: ubuntu-26.04
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix:
         include:
           - arch: armv7
@@ -72,7 +72,7 @@ jobs:
         sudo apt-get update && sudo apt-get install qemu-user \
         "g++-16-${{ matrix.compiler_tgt_cpu_arch }}-linux-gnu${{ matrix.compiler_tgt_abi_suffix }}"
 
-    - name: Build and test
+    - name: Build
       run: |
         export CMAKE_BUILD_PARALLEL_LEVEL=2
         export CTEST_PARALLEL_LEVEL=2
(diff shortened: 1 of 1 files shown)


.github/workflows/multiarch.yml
```diff
@@ -11,58 +10,178 @@ permissions:
 
 jobs:
   multiarch:
-    runs-on: ubuntu-22.04
+    name: Multiarch build and test for ${{ matrix.arch }} with GCC 16
+    runs-on: ubuntu-26.04
     strategy:
-      fail-fast: true
+      fail-fast: false
```
jan-wassenberg: hm, this is going to be expensive. If there is a job that often fails (armv7?) and causes the others to cancel, maybe instead of disabling fail-fast we should mark that one continue-on-error?

.github/workflows/multiarch.yml
```diff
@@ -11,58 +10,178 @@ permissions:
...
+          -DCMAKE_SYSTEM_NAME=Linux -DCMAKE_SYSTEM_PROCESSOR="${{ matrix.compiler_tgt_cpu_arch }}" \
+          ${{ matrix.cmake_flags }} -B out .
+        cmake --build out
+        ctest --test-dir out
   aarch64_cmake:
     name: Build and test ${{ matrix.name }} on AArch64
-    runs-on: ubuntu-24.04-arm
+    runs-on: ubuntu-26.04-arm
     strategy:
+      fail-fast: false
```
jan-wassenberg: Here also.

.github/workflows/multiarch.yml
johnplatts: @jan-wassenberg - I adjusted the workflow by separating out the test step from the build step and marking the test step with continue-on-error.

jan-wassenberg (approved):
Thanks for updating, LGTM.

PR #3159, opened 2026-06-26
Add RVV overrides for Reorder/OrderedShiftRightAndDemote2To

eyupcanakman:
Implements the RVV override for `ReorderShiftRightAndDemote2To` and `OrderedShiftRightAndDemote2To` (and rounding variants), following #3140 and #3087 for #2628. Single-step narrowing reuses the fused `vnclip`/`vnclipu` from #3140, exactly as RVV `ReorderDemote2To` reuses `DemoteTo`; multi-step falls back to `ReorderDemote2To(d, [Rounding]ShiftRight<k>(a), ...)`.

CI has no RVV target, so I checked under qemu (rv64gcv, clang 19) and results match the generic path at VLEN 128 and 256; gcc-14 builds clean. SVE2/LSX/LASX stay TODO.


jan-wassenberg (approved):
LGTM :)

PR #3160, opened 2026-06-26
Add LSX overrides for [Rounding]ShiftRightAndDemoteTo

eyupcanakman:
Implements the LSX override for `ShiftRightAndDemoteTo` and `RoundingShiftRightAndDemoteTo`, following #3056 for #2628. Narrowing reuses the fused `vssrani`/`vssrlni` (`vssrarni`/`vssrlrni` for the rounding variant) from the existing `DemoteTo`, passing the shift amount instead of 0; multi-step cases fall back to `DemoteTo(d, [Rounding]ShiftRight<k>(v))`.

LoongArch runs in CI, so the `shift_test` differential covers this on LSX; I also checked under qemu (gcc-14, `-cpu max`) and results match the generic path. SVE2/LASX and the `Demote2To` variants stay TODO.


jan-wassenberg (approved):
Nice 👍

jan-wassenberg:
CI failed, checking why..

PR #3165, opened 2026-06-29
random: fix out-of-bounds Store in VectorXoshiro when count is not a multiple of Lanes

EylonKrause:
## What

`VectorXoshiro`'s bulk generators fill an `n`-element result with a loop strided by `Lanes()` that always does a full-vector `Store()`:

```cpp
for (std::uint64_t i = 0; i < n; i += Lanes(tag)) {
  const auto next = Update(s0, s1, s2, s3);
  Store(next, tag, result.data() + i);   // writes Lanes() lanes
}
```

When `n` is not a multiple of `Lanes()`, the final iteration writes up to `Lanes() - 1` elements past the end of `result`. This affects all four bulk paths:

- `operator()(size_t n)` and `Uniform(size_t n)` — `result` is an `AlignedVector` sized exactly `n`. The aligned allocator adds trailing slack, so the overrun usually lands in padding (which is why it isn't observed through the public API every time), but the slack is only `address % kAlias` bytes and can be smaller than the overwrite — a real heap overflow.
- `operator()<N>()` and `Uniform<N>()` — `result` is a stack `std::array<T, N>`; for `N` not a multiple of `Lanes()` this is a stack overwrite.

`random_test` only exercises power-of-two sizes (the file notes "tests do not check for vector remainders"), so CI does not catch it.

## Fix

Use `StoreN(v, d, p, HWY_MIN(Lanes(), n - i))` for the bounded store, matching the existing remainder-handling convention in `contrib` (`intdiv-inl.h`, `algo/transform-inl.h`, `hash/hash_eval.cc`). `StoreN` with a full count is equivalent to `Store`, so the multiple-of-`Lanes()` path is unchanged.

## Testing (`-fsanitize=address,undefined`)

- A minimal harness reproducing the loop over an exactly-sized `std::vector` (`n = Lanes() + 1`) reports, **before** the change:
  ```
  ERROR: AddressSanitizer: heap-buffer-overflow ... WRITE of size 16
      #1 Store<...> hwy/ops/x86_128-inl.h
  0x...058 is located 0 bytes after 24-byte region
  ```
  and exits cleanly with the `StoreN` form.
- `contrib/random/random_test` passes **128/128** under ASan with the fix applied (SSSE3/SSE2/SCALAR targets), confirming no regression.

### DIFF
--- a/hwy/contrib/random/random-inl.h
+++ b/hwy/contrib/random/random-inl.h
@@ -210,9 +210,17 @@ class VectorXoshiro {
     auto s1 = Load(tag, state_[{1}].data());
     auto s2 = Load(tag, state_[{2}].data());
     auto s3 = Load(tag, state_[{3}].data());
-    for (std::uint64_t i = 0; i < n; i += Lanes(tag)) {
+    const size_t lanes = Lanes(tag);
+    // Full vectors use the cheap Store; StoreN (which can be expensive on some
+    // targets) is only used once, for the final partial vector.
+    size_t i = 0;
+    for (; i + lanes <= n; i += lanes) {
       const auto next = Update(s0, s1, s2, s3);
       Store(next, tag, result.data() + i);
+    }
+    if (i < n) {
+      const auto next = Update(s0, s1, s2, s3);
+      StoreN(next, tag, result.data() + i, n - i);
     }
     Store(s0, tag, state_[{0}].data());
     Store(s1, tag, state_[{1}].data());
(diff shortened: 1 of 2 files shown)


EylonKrause:
@googlebot rescan

hwy/contrib/random/random-inl.h
```diff
@@ -212,7 +212,7 @@ class VectorXoshiro {
     auto s3 = Load(tag, state_[{3}].data());
     for (std::uint64_t i = 0; i < n; i += Lanes(tag)) {
       const auto next = Update(s0, s1, s2, s3);
-      Store(next, tag, result.data() + i);
+      StoreN(next, tag, result.data() + i, HWY_MIN(Lanes(tag), n - i));
```
jan-wassenberg: This is actually expensive on some targets and we do not want to include it in the main loop.
When writing loops, we recommend the clunky but efficient:
```
  size_t i = 0;
  if (HWY_LIKELY(count >= N)) {
    for (; i <= count - 2 * N; i += N) {
      // loop body
    }
  }

  size_t remaining = count - i;
  // same loop body, but LoadN/StoreN with remaining
```

jan-wassenberg (changes requested):
Thanks for the fix. I agree assuming vector multiple is not ideal.

hwy/contrib/random/random-inl.h
EylonKrause: Thanks! Updated all four loops (`operator()(n)`, `operator()<N>()`, `Uniform(n)`, `Uniform<N>()`) to keep the cheap full-width `Store` in the main loop and use `StoreN` only once, for the final partial vector, per your suggestion — the number of `Update()` calls (and therefore the generated values) is unchanged.

Also added `TestNextNRemainder` covering non-multiple-of-`Lanes` sizes for all four overloads, since the existing tests only exercised power-of-two sizes (which is why the original out-of-bounds `Store` slipped through).

jan-wassenberg (approved):
Thanks for updating. FYI the reason for the if() + i <= N - lanes form is that that avoids static analysis warnings which I have seen from some GCC. i + lanes could theoretically overflow, according to the static analyzer. We can land as-is because I currently do not see these warnings in the checks.

PR #3166, opened 2026-06-30
Updated implementation of Compress ops

johnplatts:
Resolves one of the issues raised in issue #3142.

Moved the default implementation of the Compress ops for vectors with 16-bit, 32-bit, or 64-bit lane types into generic_ops-inl.h.

Added new `Compress(d, v, m)` and `CompressNot(d, v, m)` overloads that take an additional `d` parameter as the additional `d` parameter allows the SVE implementation of I8/U8/I16/U16/F16/BF16 `Compress(d, v, m)` to avoid having to split `v` into halves if `Lanes(d) <= Lanes(RepartitionToWide<RebindToUnsigned<DFromV<decltype(v)>>>())` is known to be true at compile time.

Added new `CompressBits(d, v, bits)` overload that takes an additional `d` parameter, which allows the `CompressBits(d, v, bits)` overload to avoid having to load more than `HWY_MAX(Lanes(d), 8)` mask bits if `Lanes(d) < Lanes(DFromV<decltype(v)>())` is true, which is possible on RVV and SVE targets.

In addition, it is also possible for `Lanes(DFromV<decltype(v)>())` to be greater than 64 on RVV and SVE targets (other than SVE_256 or SVE2_128), and LoadMaskBits only requires that 8 bytes of `bits` be readable if `Lanes(d) <= 64` is true.

Also reimplemented I8/U8/I16/U16/F16/BF16 `Compress(d, v, m)` for SVE targets to avoid unnecessary splitting of `v` if `Lanes(d) <= Lanes(RepartitionToWide<RebindToUnsigned<DFromV<decltype(v)>>>())` is known to be true at compile time.

The new I8/U8 Compress implementation in arm_sve-inl.h also avoids using temporary arrays to store `v` and the bits of `m` as SVE can efficiently combine the results of compressing the two halves of `v` using detail::Splice.

Also updated emu128-inl.h to use the implementation of Compress in emu128-inl.h for I8/U8 vectors as the implementation in emu128-inl.h is more efficient than the default implementation in generic_ops-inl.h for I8/U8 vectors.

Also updated `CompressIsPartition<T>` to make CompressIsPartition<T>::value equal to 1 for all lane types on the EMU128 target as the EMU128 implementation of Compress is known to be a partitioning compress.


hwy/ops/arm_sve-inl.h
```diff
@@ -4540,65 +4550,103 @@ HWY_API V Compress(V v, svbool_t mask) {
...
-  const V evenL = BitCast(d16, compressedL);
-  const V evenH = BitCast(d16, compressedH);
-  const V v16L = detail::ConcatEvenFull(evenL, evenL);  // lower half
-  const V v16H = detail::ConcatEvenFull(evenH, evenH);
+namespace detail {
+
+template <class D>
+static constexpr bool NeedToSplitForCompress8Or16() {
+  return (HWY_MAX_LANES_D(D) * sizeof(TFromD<D>)) >=
+             ((HWY_TARGET == HWY_SVE_256) ? 32 : 16) &&
```
jan-wassenberg: Should this be HWY_MIN_BYTES? It is possible we later add a 512-bit SVE target.

hwy/ops/arm_sve-inl.h
```diff
@@ -4540,65 +4550,103 @@ HWY_API V Compress(V v, svbool_t mask) {
...
+    hwy::EnableIf<!detail::NeedToSplitForCompress8Or16<RemoveCvRef<D>>()>* =
+        nullptr>
+HWY_API VFromD<D> Compress(D d, VFromD<D> v, const svbool_t mask) {
+  using T = TFromD<D>;
+  using TU = MakeUnsigned<T>;
+
+  using TW =
+      If<(sizeof(T) == 1 &&
+          ((HWY_POW2_D(D) <= -2) ||
+           (HWY_MAX_LANES_D(D) <= ((HWY_TARGET == HWY_SVE_256) ? 8 : 4)))),
```
jan-wassenberg: HWY_MIN_BYTES/4?

hwy/ops/arm_sve-inl.h
```diff
@@ -4540,65 +4550,103 @@ HWY_API V Compress(V v, svbool_t mask) {
...
 
+template <class D, HWY_IF_NOT_T_SIZE_D(D, 8)>
+HWY_API VFromD<D> CompressNot(D d, VFromD<D> v, const svbool_t mask) {
+  return Compress(d, v, Not(mask));
+}
+
 template <class V, HWY_IF_T_SIZE_V(V, 8)>
 HWY_API V CompressNot(V v, svbool_t mask) {
-#if HWY_TARGET == HWY_SVE2_128 || HWY_IDE
+#if HWY_TARGET == HWY_SVE2_128
```
jan-wassenberg: FYI the HWY_IDE is to prevent our IDE from greying out the code. Not a big deal for short code sections, but still nice to keep. The endif/if instead of elif is also for this purpose, so that multiple blocks are not greyed out.

hwy/ops/generic_ops-inl.h
```diff
@@ -6388,13 +6403,988 @@ HWY_API V CompressBits(V v, const uint8_t* HWY_RESTRICT bits) {
...
+  alignas(16) static constexpr uint8_t kCompressNot64x2Table[64] = {
+      // PrintCompressNot64x2Tables
+      0, 1, 2,  3,  4,  5,  6,  7,  8, 9, 10, 11, 12, 13, 14, 15,
+      8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2,  3,  4,  5,  6,  7,
+      0, 1, 2,  3,  4,  5,  6,  7,  8, 9, 10, 11, 12, 13, 14, 15,
+      0, 1, 2,  3,  4,  5,  6,  7,  8, 9, 10, 11, 12, 13, 14, 15};
+  return kCompressNot64x2Table;
+}
+
+#if HWY_CAP_GE256
```
jan-wassenberg: This one is deprecated, how about HWY_MIN_BYTES >= 32?

hwy/ops/generic_ops-inl.h
```diff
@@ -6388,13 +6403,988 @@ HWY_API V CompressBits(V v, const uint8_t* HWY_RESTRICT bits) {
...
+  // byte indices for PSHUFB (one vector's worth for each of 256 combinations of
+  // 8 mask bits). Loading them directly would require 4 KiB. We can instead
+  // store lane indices and convert to byte indices (2*lane + 0..1), with the
+  // doubling baked into the table. AVX2 Compress32 stores eight 4-bit lane
+  // indices (total 1 KiB), broadcasts them into each 32-bit lane and shifts.
+  // Here, 16-bit lanes are too narrow to hold all bits, and unpacking nibbles
+  // is likely more costly than the higher cache footprint from storing bytes.
+  const uint8_t (&table)[2048] = detail::GetCompress16x8Table();
+
+#if HWY_IS_BIG_ENDIAN
```
jan-wassenberg, line 6912: Would you like to add/use a constexpr NativeFromLittleEndian (or similar name) function to reduce #if?

hwy/ops/generic_ops-inl.h
```diff
@@ -6388,13 +6403,988 @@ HWY_API V CompressBits(V v, const uint8_t* HWY_RESTRICT bits) {
...
+  // BitsFromMask clears the upper bits.
+  HWY_IF_CONSTEXPR(N < 16 / sizeof(T)) {
+    return detail::CompressBits(v, BitsFromMask(d, Not(mask)));
+  }
+  else {
+    return detail::CompressNotBits(v, BitsFromMask(d, mask));
+  }
+}
+
+#if HWY_CAP_GE256
```
jan-wassenberg: Also HWY_MIN_BYTES?

hwy/ops/generic_ops-inl.h
```diff
@@ -6388,13 +6403,988 @@ HWY_API V CompressBits(V v, const uint8_t* HWY_RESTRICT bits) {
...
+static HWY_INLINE Vec256<T> CompressNotBits(Vec256<T> v,
+                                            const uint64_t mask_bits) {
+  const DFromV<decltype(v)> d;
+  const Repartition<uint32_t, decltype(d)> du32;
+
+  HWY_DASSERT(mask_bits < (1ull << Lanes(d)));
+  // 32-bit indices because we only have _mm256_permutevar8x32_epi32 (there is
+  // no instruction for 4x64).
+  const auto indices_vec = CompressIndicesFromNotBits256<T>(mask_bits);
+#if HWY_TARGET == HWY_WASM_EMU256
```
jan-wassenberg, line 7230: Comment on why this target specifically? AVX2 and AVX3 already wrap around. Not sure about LASX?

hwy/ops/generic_ops-inl.h
```diff
@@ -6388,13 +6403,988 @@ HWY_API V CompressBits(V v, const uint8_t* HWY_RESTRICT bits) {
...
+
+template <class D, HWY_IF_NOT_T_SIZE_D(D, 1)>
+HWY_API size_t CompressBlendedStore(VFromD<D> v, MFromD<D> m, D d,
+                                    TFromD<D>* HWY_RESTRICT unaligned) {
+  const uint64_t mask_bits = BitsFromMask(d, m);
+  HWY_DASSERT(mask_bits < (1ull << MaxLanes(d)));
+  const size_t count = PopCount(mask_bits);
+
+  const auto compressed = Compress(v, m);
+#if (HWY_ARCH_S390X && HWY_TARGET <= HWY_Z14) ||           \
```
jan-wassenberg: Should we also add a HWY_NATIVE_STOREN (like HWY_NATIVE_MASK) for this?

jan-wassenberg (changes requested):
Excellent, thanks for tackling this duplication.
Some suggestions, happy to discuss:

hwy/ops/set_macros-inl.h
```diff
@@ -777,6 +796,7 @@
 #define HWY_NATIVE_FMA 1
 #define HWY_NATIVE_DOT_BF16 0
 #define HWY_NATIVE_MASK 1
+#define HWY_HAVE_NATIVE_STORE_N 1
```
jan-wassenberg: One more thing: how about we rename HWY_HAVE_NATIVE_STORE_N to HWY_NATIVE_STORE_N ? Currently, HWY_HAVE_ is for "are these available at all, e.g. float64", whereas HWY_NATIVE_MASK etc mean "it's available everywhere, but is it efficient"? This new flag belongs to the latter category, so good to keep it consistent with naming.

jan-wassenberg (dismissed):

jan-wassenberg:
We had an internal user calling the detail:: functions, causing internal CI to fail. I will fix that soon.

johnplatts:
> We had an internal user calling the detail:: functions, causing internal CI to fail. I will fix that soon.

@jan-wassenberg What is the status of the fix that is causing the internal CI to fail?

jan-wassenberg:
Thanks for following up. Sorry, no progress yet - I am backlogged due to heavy work on gemma.cpp. Is this PR urgent for your use case/workflow?

jan-wassenberg:
Some progress. I've implemented the fix and am waiting for internal CI.

jan-wassenberg:
This has now landed. @johnplatts , would you mind rebasing so we can try again?

jan-wassenberg:
Thanks for updating. FYI we're hitting another internal-only usage (of detail::Compress) which I have fixed, waiting for that to land.

jan-wassenberg:
Done :)

PR #3169, opened 2026-06-30
Enable RVV runtime dispatch for GCC 15+

### LINKED ISSUE #2738: -march  rv64gcv1p0 does not make sense on RVA23 platforms

rajeshgangam:
## Summary

Fixes #2738. Enables RISC-V Vector runtime dispatch for GCC 15+, which fixes build failures when GCC does LTO across translation units compiled with different `-march` settings (e.g. Ubuntu 25.10 targeting RVA23).

**Changes:**
- Enable `HWY_HAVE_RUNTIME_DISPATCH_RVV` for `HWY_COMPILER_GCC_ACTUAL >= 1500`
- Add `HWY_TARGET_STR "arch=+v"` for GCC 15+
- Fix float16 type mismatches when `HWY_RVV_HAVE_F16C=1` but `HWY_HAVE_FLOAT16=0`:
  - `BitCastFromByte`: route through uint16 delegation path when `!HWY_HAVE_FLOAT16`
  - `PromoteTo` (f16->f32): add wrapper macro reinterpreting `vuint16` to `vfloat16` before calling the conversion intrinsic
  - `DemoteTo` (f32->f16): add wrapper macro reinterpreting `vfloat16` back to `vuint16` after conversion

## Test plan

- Cross-compiled with GCC 16.1 (baseline `rv64gc`, runtime dispatch via target attributes)
- Tested under QEMU with `rv64,v=true,zvfhmin=true`
- 886/896 tests pass; 6 failures are pre-existing GCC codegen issues (interleaved loads, tuple ops, integer division — unrelated to this change)
- Runtime dispatch confirmed working via `hwy_list_targets`

### DIFF
--- a/hwy/detect_targets.h
+++ b/hwy/detect_targets.h
@@ -792,7 +792,8 @@
 // Clang 19+ supports target attributes for RVV intrinsics (resolved in
 // https://github.com/llvm/llvm-project/issues/56592 and
 // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=115325).
-#if HWY_ARCH_RISCV && HWY_COMPILER_CLANG >= 1900
+#if HWY_ARCH_RISCV && \
+    (HWY_COMPILER_CLANG >= 1900 || HWY_COMPILER_GCC_ACTUAL >= 1500)
 #define HWY_HAVE_RUNTIME_DISPATCH_RVV 1
 #else
 #define HWY_HAVE_RUNTIME_DISPATCH_RVV 0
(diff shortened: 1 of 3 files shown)


jan-wassenberg (approved):
Nice, thanks for sending the PR!

PR #3179, opened 2026-07-02
Add opt-in APX to AVX10_2

### LINKED ISSUE #3114: add APX to AVX10_2
Looks like all targets with AVX10.2 support also support APX.

Andarwinux:
fixes #3114


Andarwinux:
@jan-wassenberg Could you commit this instead? I don't want to make my email public.

jan-wassenberg:
Yes, I can land something similar.

PR #3191, opened 2026-07-03
Mixed precision dot product example

bkmgit:


hwy/examples/dot_product_mixed_precision.cc
```diff
@@ -0,0 +1,120 @@
...
+  V sum = hn::Zero(du32);
+  const size_t N8 = hn::Lanes(du8);
+  size_t i = 0;
+  if (count >= N8) {
+    for (; i <= count - N8; i += N8) {
+      sum = hn::SumOfMulQuadAccumulate(du32, hn::LoadU(du8, a + i), hn::LoadU(du8, b + i), sum);
+    }
+  }
+  uint32_t total = hn::ReduceSum(du32, sum);
+  // Simple scalar remainder handling
```
jan-wassenberg: How about having one final 'partial/remainder' loop body using LoadN for a and b? The remainders can be up to 63, which is a lot.

hwy/examples/dot_product_mixed_precision.cc
```diff
@@ -0,0 +1,120 @@
...
+uint8_t and the output is uint32_t.
+
+*/
+
+HWY_BEFORE_NAMESPACE();
+namespace hwy {
+namespace HWY_NAMESPACE {
+namespace hn = hwy::HWY_NAMESPACE;
+
+uint32_t DotProductScalar(const uint8_t* HWY_RESTRICT a, const uint8_t* HWY_RESTRICT b, size_t count) {
```
jan-wassenberg: This one doesn't have to dispatch and can even be outside HWY_NAMESPACE if we like - as in dot_product_unroll.cc.

hwy/examples/dot_product_mixed_precision.cc
```diff
@@ -0,0 +1,120 @@
...
+}  // namespace hwy
+
+int main() {
+  const size_t count = 10000025;
+  std::vector<uint8_t> a(count);
+  std::vector<uint8_t> b(count);
+  std::iota(a.begin(), a.end(), 1);
+  std::iota(b.begin(), b.end(), 1);
+  //start timer
+  auto startScalar = std::chrono::high_resolution_clock::now();
```
jan-wassenberg: Prefer hwy::platform::Now from hwy/timer.h, that returns seconds as fp64.

hwy/examples/README.md
```diff
@@ -27,6 +27,12 @@ Similar to sum_array_advanced, but for dot products. Adds:
 -   Duff's device for vector remainders
 -   Horizontal sum (reduction/fold) using `ReduceSum`.
 
+### `dot_product_mixed_precision.cc
+
+Similar to array simple but uses automatic type casting to allow
```
jan-wassenberg: Rather than automatic type casting, should we say something like "the instructions promote from u8 to u32 to prevent overflow"?

jan-wassenberg (changes requested):
Thanks for adding - a few minor suggestions :)

hwy/examples/dot_product_mixed_precision.cc
bkmgit: Done.

hwy/examples/dot_product_mixed_precision.cc
bkmgit: Done.

hwy/examples/README.md
bkmgit: Done.

bkmgit:
There is a function `SafeCopyN` in contrib, should this be used for copying values from a non-aligned array to an aligned array?

jan-wassenberg:
Thanks for updating :) For copying to aligned, I'd recommend just hwy::CopyBytes, no need to write a new function. But it's likely unnecessary: as long as pointers are aligned, LoadU will be as efficient as Load. Load just ensures the binary will crash if there is misalignment. So no need for a separate function, we can just remove the extra aligned path.

bkmgit:
Removed the aligned path.

jan-wassenberg (dismissed):
👍

bkmgit:
Removed some unnecessary includes and updated the description in the README

jan-wassenberg (approved):
Typo on promtion but let's land as-is before the meeting :)

PR #3193, opened 2026-07-06
Added ScalarByteSwap, NativeFromLittleEndian, and IsConsteval

johnplatts:
Added ScalarByteSwap, which is equivalent to C++23 std::byteswap, except that ScalarByteSwap is available in C++11, C++14, C++17, and C++20 modes and ScalarByteSwap has no dependencies on the C++ standard library `<bit>` header.

Added NativeFromLittleEndian, which returns `HWY_IS_BIG_ENDIAN ? ScalarByteSwap(val) : val`.

Added IsConsteval, which returns true if it is known that the call is being made in a constant-evaluated context and false otherwise.

HWY_HAVE_IS_CONSTEVAL is 1 if support for C++23 if consteval or __builtin_is_constant_evaluated() is available, and 0 otherwise.

If HWY_HAVE_IS_CONSTEVAL is 1, `hwy::IsConsteval()` is equivalent to `std::is_constant_evaluated()`, except that `hwy::IsConsteval()` has no dependencies on the `<type_traits>` header and `hwy::IsConsteval()` is also available in C++11, C++14, and C++17 modes.

`hwy::IsConsteval()` allows the implementation of ScalarByteSwap for MSVC cl 19.26 or later to use the _byteswap_ushort, _byteswap_ulong, or _byteswap_uint64 compiler intrinsics if called from a non-constant-evaluated context and to use constant expression fallbacks if called from a constant-evaluated context.


hwy/base.h
```diff
@@ -400,6 +400,51 @@ HWY_DLLEXPORT HWY_NORETURN void HWY_FORMAT(3, 4)
...
 
+//------------------------------------------------------------------------------
+// IsKnownToBeConstantEvaluated
+
+// If C++23 if consteval support is available or the
+// __builtin_is_constant_evaluated() builtin is available,
+// IsKnownToBeConstantEvaluated() returns true if the call is being evaluated
+// within a constant-evaluated context.
+
+// Otherwise, if C++23 if consteval support is not available and the
```
jan-wassenberg: We could simplify this condition by reference to `HWY_HAVE_IS_CONSTEVAL[_BUILTIN]` and just say that IsConsteval (IsKnownToBeConstantEvaluated) returns false if `HWY_HAVE_IS_CONSTEVAL[_BUILTIN]` is false?

hwy/base.h
```diff
@@ -400,6 +400,51 @@ HWY_DLLEXPORT HWY_NORETURN void HWY_FORMAT(3, 4)
...
+// hwy::IsKnownToBeConstantEvaluated() does not have any dependencies on the
+// <type_traits> header and hwy::IsKnownToBeConstantEvaluated() is also
+// available in C++11, C++14, and C++17 modes.
+
+#if defined(__cpp_if_consteval) && __cpp_if_consteval >= 202106L
+#define HWY_HAVE_IS_CONSTEVAL_BUILTIN 1
+
+HWY_API constexpr bool IsKnownToBeConstantEvaluated() {
+  if consteval {
+    return false;
```
jan-wassenberg: Should it be the other way around?

jan-wassenberg (changes requested):
Thanks for adding, these all seem useful. I wonder about the naming - they are a bit long. Would it help to shorten IsKnownToBeConstantEvaluated to IsConsteval, also bringing it closer to the name of HWY_HAVE_IS_CONSTEVAL_BUILTIN?

For HWY_HAVE_IS_CONSTEVAL_BUILTIN, we generally don't mention BUILTIN in the macro names, could drop that suffix unless you have strong feelings about keeping it?

Finally, ScalarByteSwapIfBigEndian expresses the mechanics more than the intent - wouldn't NativeFromLittleEndian(val) be more clear? Or even FromLittleEndian?

jan-wassenberg (approved):
Thanks for updating :)

Issue #3192, opened 2026-07-05
highway/hwy/ops/set_macros-inl.h:73:3: error: expected unqualified-id after #3173


Andarwinux:
```
highway/hwy/ops/set_macros-inl.h:73:3: error: expected unqualified-id
   73 |   || HWY_TARGET == HWY_SVE2_128
      |   ^
```

Andarwinux:
A `\` should be added or merged into the same line.

johnplatts:
@Andarwinux and @jan-wassenberg - I have fixed this compiler error in pull request #3194.

PR #3194, opened 2026-07-06
Fixed compiler errors on AArch64, resolves issue #3192

johnplatts:
Updated set_macros-inl.h to fix compiler error on AArch64.


jan-wassenberg (approved):
Thanks for fixing, unfortunately I didn't spot this during review.

PR #3195, opened 2026-07-06
Fixed errors in hash_test and phast_test if HWY_TEST_STANDALONE is 1

johnplatts:
Added new `HWY_GTEST_SKIP()` macro, which is defined as `GTEST_SKIP()` if HWY_TEST_STANDALONE is 0 and `return` if HWY_TEST_STANDALONE is 1. The `HWY_GTEST_SKIP()` macro does not support `operator <<` if HWY_TEST_STANDALONE is 1.

Fixed compiler error in hash_test.cc if HWY_TEST_STANDALONE is 1 by replacing GTEST_SKIP() with HWY_GTEST_SKIP().

Also replaced EXPECT_EQ with HWY_ASSERT_EQ in phast_test.cc to fix compiler error if HWY_TEST_STANDALONE is 1.


jan-wassenberg (approved):
Thanks for fixing :)

PR #3196, opened 2026-07-06
Made fixes and enhancements to ExpandLoop on SVE

johnplatts:
Updated ExpandLoop in arm_sve-inl.h to use `Iota(du, 0)` instead of `Iota(d, 0)` for `iota` to avoid issues if `TFromV<decltype(v)>` is a floating-point type.

Also updated ExpandLoop in arm_sve-inl.h to use SlideUpLanesOr to avoid having to store the lanes of `v` in a temporary vector and having to reload the result of `ExpandLoop`.


jan-wassenberg (approved):
👍

PR #3199, opened 2026-07-06
Fix TestShardMul on bi-endian

miladfarca:
Commit 4ecb74a has caused the following test failures on s390x:
```
ShardMulTestGroup/ShardMulTest.TestShardMulTwoVec
ShardMulTestGroup/ShardMulTest.TestShardMulCollisionFree
```
On big-endian, BitCast places the upper 32 bits in even lanes, so swap Even/Odd to match scalar.


jan-wassenberg (approved):
Thanks for fixing. I think we're going to want to have ConcatTop/Bottom helpers that eval to Odd/Even depending on byte order. Added to my TODO.

miladfarca:
Sounds good, thank you!

PR #3200, opened 2026-07-06
Mandelbrot example

bkmgit:
Demonstrates interleaving and working with structures of arrays

This style of example is used in several other places:
- https://www.mikekohn.net/software/mandelbrots_simd.php
- https://accelerated-computing.academy/fall25/labs/lab1/
- https://ispc.github.io/example.html
- https://web-docs.gsi.de/~mkretz/egrd24-simd/0000-01-08-mandelbrot.html

### DIFF
--- a/hwy/examples/mandelbrot.cc
+++ b/hwy/examples/mandelbrot.cc
@@ -1,0 +1,638 @@
+// Copyright 2026 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include <stddef.h>
+
+#include <bitset>
+#include <cmath>
+#include <fstream>
+#include <iostream>
+#include <set>
+#include <string>
+
+#undef HWY_TARGET_INCLUDE
+#define HWY_TARGET_INCLUDE "hwy/examples/mandelbrot.cc"
+#include "hwy/aligned_allocator.h"
+#include "hwy/foreach_target.h"  // IWYU pragma: keep
+#include "hwy/highway.h"
+#include "hwy/timer.h"
+
+/*
+Highway SIMD Tutorial: Compute and visualize a Mandelbrot set
+
+This example demonstrates how to compute the Mandelbrot[0] set over a given
+domain for a choosen function.  The output is saved in portable pixmap[1] file
+format.  Visualization uses a technique inspired by Buddhabrot[2], for each RGB
+color, a different maximum iteration is choosen.
+
+Consider the iteration
+z_(n+1) = (1-2(z_n)^3)/3(z_n)^2
+where z_n is the complex number x_n + i*y_n.
+This iteration can also be written as
+
+x_(n+1) = ((x_n^2-y_n^2)*(1-2*(x_n^3-3*x_n*y_n^2))
+           -4*x_n*y_n*(3*x_n^2*y_n-y_n^3))
+          /(3*(y_n^2+x_n^2)^2)
+
+y_(n+1) = -2*((x_n^2-y_n^2)*(3*x_n^2*y_n-y_n^3)
+              +x_n*y_n*(1-2*(x_n^3-3*x_n*y_n^2)))
+          /(3*(y_n^2+x_n^2)^2)
+
+A Mandelbrot set is the set of initial starting points which for the value of
+x_n^2 + y_n^2 does not tend to infinity.  In a computer implementation, we
+choose a maximum iteration number and to avoid overflows, a maximum value after
+which we denote the value as tending to infinity.
+
+To visualize the set, three different choices of the maximum iteration
+will be choosen for the same predefined maximum value. The resulting arrays
+for Mandelbrot sets will be assigned to the red, blue and green colors to create
+a color image.  If the point has not escaped, it will be assigned a value of 0,
+otherwise it will be assigned a value of
+
+255 * ( maximum number of iterations - iteration of escape) / (maximum number
+interations)
+
+rounded up to the nearest integer.
+
+References:
+0) https://en.wikipedia.org/wiki/Mandelbrot_set
+1) https://en.wikipedia.org/wiki/Netpbm
+2) https://en.wikipedia.org/wiki/Buddhabrot
+3) https://paulbourke.net/fractals/
+
+*/
+
+HWY_BEFORE_NAMESPACE();
+namespace hwy {
+namespace HWY_NAMESPACE {
+namespace hn = hwy::HWY_NAMESPACE;
+
+using DU8 = hn::ScalableTag<uint8_t>;
+const DU8 du8;
+using MU8 = hn::Mask<decltype(du8)>;
+using VU8 = hn::Vec<DU8>;
+using DU16 = hn::ScalableTag<uint16_t>;
+const DU16 du16;
+using MU16 = hn::Mask<decltype(du16)>;
+using VU16 = hn::Vec<DU16>;
+using DU32 = hn::ScalableTag<uint32_t>;
+const DU32 du32;
+using MU32 = hn::Mask<decltype(du32)>;
+using VU32 = hn::Vec<DU32>;
+using DF = hn::ScalableTag<float>;
+const DF df;
+using MF = hn::Mask<decltype(df)>;
+using VF = hn::Vec<DF>;
+
+void MandelbrotScalarSetup(uint8_t* HWY_RESTRICT r, uint8_t* HWY_RESTRICT g,
+                           uint8_t* HWY_RESTRICT b, float* HWY_RESTRICT x,
+                           float* HWY_RESTRICT y, const size_t x_points,
+                           const size_t y_points) {
+  float x_max = static_cast<float>(x_points / 2);
+  float y_max = static_cast<float>(y_points / 2);
+  for (size_t ij = 0; ij < x_points * y_points; ij++) {
+    float itemp = static_cast<float>(ij % y_points);
+    float i = itemp - x_max;
+    float j =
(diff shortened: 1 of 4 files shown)


hwy/examples/mandelbrot.cc
```diff
@@ -0,0 +1,285 @@
...
+5) https://paulbourke.net/fractals/
+*/
+
+HWY_BEFORE_NAMESPACE();
+namespace hwy {
+namespace HWY_NAMESPACE {
+namespace hn = hwy::HWY_NAMESPACE;
+
+float next_x(float x, float y) {
+  return ((std::pow(x, 2) - std::pow(y, 2)) *
```
jan-wassenberg: Rather than pow with exponent 2..4, let's instead write Pow2..4 helper functions that just return `x*x` and `x*x*x*x`?

hwy/examples/mandelbrot.cc
```diff
@@ -0,0 +1,285 @@
...
+          float x_temp = x[x_ind + y_ind * x_points];
+          float y_temp = y[x_ind + y_ind * x_points];
+          float x_new = next_x(x_temp, y_temp);
+          float y_new = next_y(x_temp, y_temp);
+          float norm_new = std::sqrt(std::pow(x_new, 2) + std::pow(y_new, 2));
+          x[x_ind + y_ind * x_points] = x_new;
+          y[x_ind + y_ind * x_points] = y_new;
+          if (norm_new > escape_value) {
+            r[x_ind + y_ind * x_points] = static_cast<uint8_t>(
+                std::ceil(255.0 * static_cast<float>(iter_max_r - n) /
```
jan-wassenberg: 0.f suffix, otherwise will promote to double.

jan-wassenberg (changes requested):
Nice example idea :)

hwy/examples/mandelbrot.cc
bkmgit: Done.

hwy/examples/mandelbrot.cc
bkmgit: Done.

hwy/examples/mandelbrot.cc
```diff
@@ -0,0 +1,637 @@
...
+}
+
+void CreatePPMSimd(const uint32_t* HWY_RESTRICT r,
+                   const uint32_t* HWY_RESTRICT g,
+                   const uint32_t* HWY_RESTRICT b,
+                   uint32_t* HWY_RESTRICT ppm,
+                   const size_t x_points, const size_t y_points) {
+
+  size_t n = 0;
+/* Seems to order output incorrectly
```
bkmgit: `StoreInterleaved3` seems not to work correctly here. Not sure why.

hwy/examples/mandelbrot.cc
```diff
@@ -0,0 +1,637 @@
...
+  	   	  du32,
+		  r + ij);
+        hn::Store(hn::MaskedSetOr(hn::Load(du32, g + ij), mask_newly_escaped, g_new),
+		  du32,
+		  g + ij);
+        hn::Store(hn::MaskedSetOr(hn::Load(du32, b + ij), mask_newly_escaped, b_new),
+		  du32,
+		  b + ij);
+        if (hn::AllTrue(du32, mask_escaped)) {
+          not_escaped[ij] = 0;
```
bkmgit: Seem to have trouble getting masks to work with `uint8_t`, not sure if this functionality can be added and if so, if it is worth adding.  Maybe it is better to use `uint16_t` and `float16_t`?  Computations could probably also be done with `bfloat16`.

hwy/examples/mandelbrot.cc
```diff
@@ -0,0 +1,637 @@
...
+}
+
+V pow4_simd(const V x) {
+  return hn::Mul(hn::Mul(x,x),hn::Mul(x,x));
+}
+
+
+V next_x_simd(const V x, const V y) {
+    return hn::Div(hn::Sub((hn::Mul(hn::Sub(pow2_simd(x), pow2_simd(y)),
+                                    (hn::Sub(hn::Set(df, 1.0f),
```
jan-wassenberg: This is getting very difficult to read. Would recommend making constants V k2 = Set(df, 2.0f), and also some lvalues / temporary variables so that each line/expression isn't so long and deeply nested.
This may also reveal some opportunities for fusing a Mul+Add into MulAdd - I see at least one.

hwy/examples/mandelbrot.cc
jan-wassenberg: What's the issue with uint8_t?
Would advise staying away from float16_t, it's only available if HWY_HAVE_FLOAT16, which is annoying. And bfloat16_t has an extremely limited set of ops available, which does not include normal arithmetic.
float32 is probably fine.

hwy/examples/mandelbrot.cc
jan-wassenberg: I think the issue is that we are only incrementing by NU32, but we are actually writing 3*NU32 values to the output.
BTW PPM files usually are 8-bit, no? Shouldn't we TruncateTo u8 (assuming values are already 0..255)?

hwy/examples/mandelbrot.cc
bkmgit: Have refactored the code.

hwy/examples/mandelbrot.cc
bkmgit: Thanks, factor of 3 resolved it.

hwy/examples/mandelbrot.cc
bkmgit: Ok, can use floats.

hwy/examples/mandelbrot.cc
```diff
@@ -0,0 +1,658 @@
...
+  const size_t NU32 = Lanes(du32);
+  const size_t NU8 = Lanes(du8);
+  uint32_t sn32[NU32*4];
+  uint8_t sn8[NU8];
+  hn::StoreU(scaled_norm_0, du32, sn32);
+  hn::StoreU(scaled_norm_1, du32, sn32 + NU32);
+  hn::StoreU(scaled_norm_2, du32, sn32 + 2 * NU32);
+  hn::StoreU(scaled_norm_3, du32, sn32 + 3 * NU32);
+  for(int i = 0; i < 4 * NU32; i++) sn8[i] = static_cast<uint8_t>(sn32[i]);
+  const VU8 scaled_norm_u8 = hn::LoadU(du8, sn8);
```
bkmgit: At the moment, compute a threshold value using floats, then convert it to U32 as this preserves size of vectors.  How should this be converted to U8 using Simd?  It is not clear to me how to combine outputs produced by U8FromU32, though maybe there is a better way to create a mask.

hwy/examples/mandelbrot.cc
```diff
@@ -0,0 +1,658 @@
...
+    hn::StoreN(hn::Mul(hn::Set(df, (3.0f / x_max_f)), i), df, x + ij, remainder);
+    hn::StoreN(hn::Mul(hn::Set(df, (3.0f / y_max_f)), j), df, y + ij, remainder);
+  }  
+  ij = 0;
+  for (; ij < x_points * y_points; ij += NU8) {
+    hn::Store(hn::Zero(du8), du8, r + ij);
+    hn::Store(hn::Zero(du8), du8, g + ij);
+    hn::Store(hn::Zero(du8), du8, b + ij);
+  }
+  // Use StoreN for remainder, does not values beyond remainder
```
jan-wassenberg: does not _access_ values..? (Also above)

hwy/examples/mandelbrot.cc
jan-wassenberg: Ah, given the type tag `du32` for VU32 (we almost always pass that around as a template param, alongside the vectors), we can do:
```
RepartitionToNarrow<decltype(du32)> du16;
RepartitionToNarrow<decltype(du16)> du8;
vu16_10 = OrderedDemote2To(du16, scaled_norm0, scaled_norm1);
vu16_32 = OrderedDemote2To(du16, scaled_norm2, scaled_norm3);
vu8 = OrderedDemote2To(du8, vu16_10, vu16_32);
```

If you don't have pairs of vectors,
```
Rebind<uint8_t, decltype(du32)> du8;
Rebind<uint16_t, decltype(du32)> du16;
vu16h = DemoteTo(du16, scaled_norm0);
vu8q = DemoteTo(du8, vu16h);
```
Halving the size always works but there is actually a shortcut, you can just say `vu8q = DemoteTo(du8, scaled_norm0);`.

hwy/examples/mandelbrot.cc
```diff
@@ -0,0 +1,658 @@
...
+  const size_t NU8 = Lanes(du8);
+  for(; n < x_points * y_points; n +=NU8) {
+    StoreInterleaved3(hn::Load(du8, r + n),
+                      hn::Load(du8, g + n),
+                      hn::Load(du8, b + n),
+                      du8,
+                      ppm + 3 * n);
+  }
+  // remainder
+  for (; n < x_points * y_points; ++n) {
```
jan-wassenberg: Would also recommend LoadN here, though it's probably not time-critical.

jan-wassenberg (commented):
Nice, much easier to see the MulAdd fusions this way :)

hwy/examples/mandelbrot.cc
bkmgit: Done.

hwy/examples/mandelbrot.cc
bkmgit: The error was in using `Demote2To` instead of `Truncate2To`.  `Truncate2To` matches the behavior of `static_cast<uint8_t>` and so gives an difference of less than 1% between the scalar and simd implementations, whereas Demote2To results in a difference of about 4%.

hwy/examples/mandelbrot.cc
bkmgit: Done.

jan-wassenberg (dismissed):
Nice, LGTM :)

bkmgit:
Some spelling corrections, but otherwise unchanged.

jan-wassenberg (approved):
hm, the way this PR is being updated, I only see the entire change, not a delta vs the last state. Anything we can do about that to prevent requiring a full re-review on each update?

bkmgit:
Usually try to squash things when ready to merge as some of the intermediate commits are unlikely to be useful to have in the main project history.  Can leave previous history until everything is resolved, though this would probably require a squash before merging. Let me know what would be preferred.

jan-wassenberg:
I think github does fine without the squash, we can try skipping that :)

PR #3204, opened 2026-07-07
Made fixes to float_distribution example

johnplatts:
Replaced std::bit_cast with hwy::BitCastScalar in float_distribution.cc as std::bit_cast requires C++20.

Also updated FloatDistributionSIMD in float_distribution.cc to work with the HWY_SCALAR target.

Also replaced HWY_MAX_BYTES with `hwy::VectorBytes()` in Verify in float_distribution.cc, as it is possible for `hwy::VectorBytes() > HWY_MAX_BYTES` to be true if `HWY_TARGET != (HWY_TARGET & (-HWY_TARGET))` is true (or in other words, there is more than 1 target enabled).

Also fixed a compiler warning with the sizes list in RunTests() in float_distribution.cc.

Also updated CMakeLists.txt to build and compile float_distribution.

### DIFF
--- a/hwy/examples/float_distribution.cc
+++ b/hwy/examples/float_distribution.cc
@@ -68,15 +68,19 @@ static HWY_INLINE size_t FloatDistributionSIMD(const float* HWY_RESTRICT in,
                                                size_t count,
                                                uint8_t threshold_val,
                                                uint8_t* HWY_RESTRICT out) {
+#if HWY_TARGET == HWY_SCALAR
+  // FloatDistributionSIMD is not supported on the HWY_SCALAR target
+
+  (void)in;
+  (void)count;
+  (void)threshold_val;
+  (void)out;
+  return 0;
+#else
   using DF32 = hn::ScalableTag<float>;
   using DU32 = hn::ScalableTag<uint32_t>;
-#if HWY_TARGET == HWY_SCALAR
-  using DU16 = hn::Rebind<uint16_t, DU32>;
-  using DU8 = hn::Rebind<uint8_t, DU32>;
-#else
   using DU16 = hn::Repartition<uint16_t, DU32>;
   using DU8 = hn::Repartition<uint8_t, DU32>;
-#endif
 
   const DF32 df32;
   const DU32 du32;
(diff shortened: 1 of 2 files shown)


hwy/examples/float_distribution.cc
```diff
@@ -99,10 +108,26 @@ static HWY_INLINE size_t FloatDistributionSIMD(const float* HWY_RESTRICT in,
...
                               VU32 exp3) HWY_ATTR {
-    // Demote compresses i32 -> i16 -> i8
-    // It performs also clamping of values.
-    // Reorder version dosen't assume any order of output elements.
-    // If you want to preserve order you should use OrderedDemote2To.
+  // Demote compresses i32 -> i16 -> i8
+  // It performs also clamping of values.
+  // Reorder version dosen't assume any order of output elements.
+  // If you want to preserve order you should use OrderedDemote2To.
+#if HWY_TARGET == HWY_SCALAR
```
jan-wassenberg: I'd prefer not to complicate/clutter examples with HWY_SCALAR workarounds; HWY_SCALAR is deprecated. How about we check #if HWY_TARGET != HWY_SCALAR and otherwise have an empty test function?

hwy/examples/float_distribution.cc
```diff
@@ -186,7 +213,7 @@ bool Verify(const float* HWY_RESTRICT in, size_t count, float threshold) {
   expected.resize(expected_size);
 
   const uint8_t threshold_val = static_cast<uint8_t>(GetExponent(threshold));
-  std::vector<uint8_t> actual(count + HWY_MAX_BYTES);
```
jan-wassenberg, line 189: Good catch, thanks. This is outside of HWY_NAMESPACE and hence not an upper bound. I'll add a note to documentation to clarify this.

jan-wassenberg (changes requested):

hwy/examples/float_distribution.cc
johnplatts: @jan-wassenberg - Already made the change to have FloatDistributionSIMD simply return 0 on HWY_SCALAR

jan-wassenberg (approved):
Thanks for changing :)

PR #3205, opened 2026-07-07
Fixed build errors with Meson build

johnplatts:
Added hwy/contrib/sort/vqsort_have.cc to hwy_contrib_sources list in meson.build to fix build errors with Meson.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -170,6 +170,7 @@ hwy_contrib_sources = files(
     'hwy/contrib/hash/phast.cc',
     'hwy/contrib/hash/shardmul.cc',
     'hwy/contrib/sort/vqsort.cc',
+    'hwy/contrib/sort/vqsort_have.cc',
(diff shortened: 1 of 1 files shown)


jan-wassenberg (changes requested):
Thanks for fixing. I see that this file is also not yet in CMakeLists.txt, would you like to add it there as well?

johnplatts:
> Thanks for fixing. I see that this file is also not yet in CMakeLists.txt, would you like to add it there as well?

vqsort_have.cc is already included in CMakeLists.txt if HWY_ENABLE_CONTRIB is ON through `file(GLOB HWY_CONTRIB_SOURCES "hwy/contrib/sort/vqsort_*.cc")`.

jan-wassenberg:
Oh, right. I thought I also saw this mentioned in a failing CMake build, but seems not; might have also been the Meson build.

jan-wassenberg (approved):

PR #3207, opened 2026-07-08
Add runtime dispatch on FreeBSD and OpenBSD

brad0:


hwy/detect_targets.h
```diff
@@ -827,6 +827,26 @@
 #endif
 #endif  // HWY_HAVE_RUNTIME_DISPATCH_LINUX
 
+#ifndef HWY_HAVE_RUNTIME_DISPATCH_FREEBSD  // allow override
```
jan-wassenberg: Is there any reason to keep these separate? Or should we merge into HWY_HAVE_RUNTIME_DISPATCH_FREEBSD_OPENBSD?

We could even have one shared default:

```
#undef HWY_HAVE_RUNTIME_DISPATCH_LINUX_BSD
#if (HWY_ARCH_ARM || HWY_ARCH_PPC || HWY_ARCH_S390X) && HWY_OS_LINUX && \
    (HWY_COMPILER_GCC_ACTUAL || HWY_COMPILER_CLANG >= 1700) && HWY_HAVE_AUXV
#define HWY_HAVE_RUNTIME_DISPATCH_LINUX_BSD 1
#else
#define HWY_HAVE_RUNTIME_DISPATCH_LINUX_BSD 0
#endif
```

And then just per platform (maybe again shared among the BSD):
```
#ifndef HWY_HAVE_RUNTIME_DISPATCH_LINUX  // allow override
#define HWY_HAVE_RUNTIME_DISPATCH_LINUX HWY_HAVE_RUNTIME_DISPATCH_LINUX_BSD
#endif  // HWY_HAVE_RUNTIME_DISPATCH_LINUX
```

What do you think?

hwy/targets.cc
```diff
@@ -503,7 +512,12 @@ static int64_t DetectTargets() {
     bits |= HWY_NEON;
 
 #if defined(HWCAP_ASIMDHP) && defined(HWCAP_ASIMDDP) && defined(HWCAP2_BF16)
+#if HWY_HAVE_ELF_AUX_INFO
```
jan-wassenberg: To avoid duplication and reduce #if, how about we add a GetHwCap2 function that wraps this? Same for GetHwCap.

jan-wassenberg (changes requested):
Nice, thanks for adding.

hwy/detect_targets.h
brad0: Not really. I figured it would be best to just keep them separated.

jan-wassenberg: Sure, that's a safe default. I don't mind separate macros as long as they share the same initialization - we don't want to duplicate the compiler version check logic, that risks diverging. BTW I notice that HWY_ARCH_S390X is allowed for Linux dispatch but not BSD, is that intentional?

brad0: Neither OS has S390X support nor Loongarch. Maybe Loongarch might happen for either.

I only added the bits for the architectures we do have, ARM/AArch64/PowerPC/RISC-V.

brad0:
How about something like this?

jan-wassenberg (dismissed):
LGTM :) I suppose this could break if getauxval() is added to either OS, but this works for now.

brad0:
> LGTM :) I suppose this could break if getauxval() is added to either OS, but this works for now.

That's not going to happen.

brad0:
Just remembered to check meson.

brad0:
@jan-wassenberg Ping.

jan-wassenberg (approved):
Thanks for pinging, didn't see the additional push.

jan-wassenberg:
(Retrying)

brad0:
@jan-wassenberg Is there anything I need to do?

jan-wassenberg:
Thanks for following up. Merges are automated but sometimes our pipeline gets stuck. I fixed it already.

brad0:
> Thanks for following up. Merges are automated but sometimes our pipeline gets stuck. I fixed it already.

@jan-wassenberg Thanks. I noticed it seems to be mostly flowing and then seemingly was stuck. I have noticed this with other projects bots and sometimes they need a bit of a kick.

PR #3208, opened 2026-07-08
Game of Life example

bkmgit:
Game of life example to demonstrate stencil computations and computations using arrays of bits.

std::vector(bool) cannot be used and so hw::AlignedVector(bool) cannot also be used.

Could possibly output results as uncompressed animated [gif](https://en.wikipedia.org/wiki/GIF) or animated [Lottie json](https://docs.lottiefiles.com/en/format/lottie-json/specification) files

### DIFF
--- a/hwy/examples/game_of_life.cc
+++ b/hwy/examples/game_of_life.cc
@@ -1,0 +1,357 @@
+// Copyright 2026 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include <stddef.h>
+#include <stdint.h>
+
+#include <iostream>
+#include <vector>
+
+#undef HWY_TARGET_INCLUDE
+#define HWY_TARGET_INCLUDE "hwy/examples/game_of_life.cc"
+
+#include "hwy/foreach_target.h"  // IWYU pragma: keep
+// Put after foreach_target.h to avoid redefinition errors
+#include "hwy/aligned_allocator.h"
+#include "hwy/contrib/random/random-inl.h"
+#include "hwy/highway.h"
+#include "hwy/timer.h"
+
+/*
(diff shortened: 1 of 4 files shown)


hwy/examples/game_of_life.cc
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  return;
+}
+
+static int Run() {
+  const size_t nx = 60;
+  const size_t ny = 60;
+  const size_t iterations = 100;
+  bool* a = (bool*)malloc(nx*ny*sizeof(bool));
```
jan-wassenberg: Prefer AlignedFreeUniquePtr<uint8_t[]> ptr = AllocateAligned<uint8_t>(num); - this avoids leaks, and is vector-aligned.

jan-wassenberg (changes requested):
Looks like a good framework :)
If you'd like to start vectorizing already, I think SlideUpLanesOr is a good way to get right neighbors, for up/down it's just loading previous row, and for left we can use LoadU(d, ptr-1).

bkmgit:
A straight forward SIMD implementation is described at:
 https://lemire.me/blog/2018/07/18/accelerating-conways-game-of-life-with-simd-instructions/

However, bit operations seem to be the way to go:
https://binary-banter.github.io/game-of-life/

This maybe a good extended example, but might not be quite so helpful for demonstrating stencil computations.

hwy/examples/game_of_life.cc
bkmgit: Done.

bkmgit:
> Looks like a good framework :) If you'd like to start vectorizing already, I think SlideUpLanesOr is a good way to get right neighbors, for up/down it's just loading previous row, and for left we can use LoadU(d, ptr-1).

This works. Can cycle through the rows to reduce number of loads further.  GCC optimizes the scalar code better than Clang, but Clang optimizes the vectorized code better than GCC.

Still need to optimize bit based implementation.

jan-wassenberg:
:) I'd suggest focusing on the stencil - the goal is to demonstrate that, rather than the best possible Game of life sim.

bkmgit:
Maybe it is better to put the bit version in a separate example?  Working directly with bits is useful for cryptography, graph traversal ([SlimSell: A Vectorized Graph Representation for Breadth-First Search](https://htor.inf.ethz.ch/publications/index.php?pub=266) as well as for binarized neural networks (managing memory accesses is of course more challenging):
- https://jonathanmarek1.github.io/binarynet-tensorflow/
- https://arxiv.org/abs/2509.07025
- https://github.com/ittuann/Binarized-Neural-Networks
- https://jonathanmarek1.github.io/binarynet-tensorflow/
- https://docs.larq.dev/larq/

jan-wassenberg (approved):
Nice usage of the new Slide1*Or ops 👍

PR #3212, opened 2026-07-11
Stream triad example

bkmgit:


bkmgit:
The code is bandwidth bound, so SIMD will not give a performance boost on most architectures.  SIMD would likely be more energy efficient as fewer instructions are needed for long vectors.  The compiler would likely optimize the scalar code.

hwy/examples/stream_triad.cc
```diff
@@ -0,0 +1,284 @@
...
+
+void StreamTriadSIMD(float* HWY_RESTRICT a, const float* HWY_RESTRICT b,
+                     const float* HWY_RESTRICT c, const float q,
+                     const size_t count) {
+
+  const size_t NF = hn::Lanes(df);
+  size_t i = 0;
+  if (count >= NF) {
+    for (; i <= count - NF; i += NF) {
+      hn::Store(
```
jan-wassenberg: Would recommend unrolling this loop 2 or 4x, might help a bit. We can also hoist the Set(df, q) out of the loop just for clarity.

hwy/examples/stream_triad.cc
```diff
@@ -0,0 +1,284 @@
...
+void StreamTriadSIMDThreads(float* HWY_RESTRICT a, const float* HWY_RESTRICT b,
+                            const float* HWY_RESTRICT c, const float q,
+                            const size_t count) {
+
+  const size_t NF = hn::Lanes(df);
+  ThreadPool pool(ThreadPool::MaxThreads());
+  size_t workers = pool.NumWorkers();
+  pool.Run(0, workers, [=](uint64_t task, HWY_MAYBE_UNUSED size_t thread) {
+    // Use work chunks that are multiples of the number of lanes
+    size_t low = NF * static_cast<size_t>(task) * (count / (NF * workers));
```
jan-wassenberg: How about using IndexRangePartition from contrib/thread_pool/index_range.h? Typically used as:
```
  IndexRangePartition StaticPartition(size_t num_items) {
    size_t size = RoundUpTo(hwy::DivCeil(num_items, pool.NumWorkers()), NF);
    return IndexRangePartition(IndexRange(0, num_items), HWY_MIN(size, num_items));
  }
```
Then partition.Range(task) gives the begin/end for each worker.

hwy/examples/stream_triad.cc
```diff
@@ -0,0 +1,284 @@
...
+                                               c.data(), q, count);
+  // Record stop timer and print time
+  const double t_simd_threads_1 = hwy::platform::Now();
+  const double dt_simd_threads = 1000.0 * (t_simd_threads_1 - t_simd_threads_0);
+  const double simd_threads_bandwidth =
+      3.0 * sizeof(float) * count / (1000.0 * dt_simd_threads);
+  bool simd_threads_validate = false;
+  simd_threads_validate = HWY_DYNAMIC_DISPATCH(TriadValidate)(
+      a_scalar.data(), a_simd_threads.data(), count, threshold);
+  if (simd_threads_validate) {
```
jan-wassenberg: Consider pulling this printing into a function to reduce duplication, either with an extra caption arg "SIMD", or just print a "SIMD:" line beforehand?

jan-wassenberg (changes requested):
Nice example :)

hwy/examples/stream_triad.cc
bkmgit: Done.

hwy/examples/stream_triad.cc
bkmgit: Done.

hwy/examples/stream_triad.cc
bkmgit: Done.

hwy/examples/stream_triad.cc
```diff
@@ -0,0 +1,290 @@
...
+}
+
+void StreamTriadSIMD(float* HWY_RESTRICT a, const float* HWY_RESTRICT b,
+                     const float* HWY_RESTRICT c, const float q,
+                     const size_t count) {
+  const size_t NF = hn::Lanes(df);
+  size_t i = 0;
+  const VF qv = hn::Set(df, q);
+  // Unroll by 4 to mask latency
+  for (; i <= count - 4 * NF; i += 4 * NF) {
```
jan-wassenberg, line 70: Here we are assuming count >= 4 * NF. Probably legit in such a benchmark, but let's still HWY_DASSERT() that.

hwy/examples/stream_triad.cc
```diff
@@ -0,0 +1,290 @@
...
+    hn::Store(MulAdd(hn::Load(df, c + i + NF), qv, hn::Load(df, b + i + NF)),
+              df, a + i + NF);
+    hn::Store(
+        MulAdd(hn::Load(df, c + i + 2 * NF), qv, hn::Load(df, b + i + 2 * NF)),
+        df, a + i + 2 * NF);
+    hn::Store(
+        MulAdd(hn::Load(df, c + i + 3 * NF), qv, hn::Load(df, b + i + 3 * NF)),
+        df, a + i + 3 * NF);
+  }
+  for (; i <= count; i += NF) {
```
jan-wassenberg: i <= count - NF, right?

jan-wassenberg (commented):
Nice :) One minor update:

hwy/examples/stream_triad.cc
bkmgit: Done

hwy/examples/stream_triad.cc
bkmgit: Done.

jan-wassenberg (dismissed):
LGTM :)

bkmgit:
Resolved merge conflict

jan-wassenberg (approved):

PR #3213, opened 2026-07-12
Specify types to masks example easier to read

bkmgit:


jan-wassenberg (approved):
Thanks, it's good to reduce auto and be explicit.

PR #3214, opened 2026-07-12
Add missing argument

bkmgit:


bkmgit:
libvips uses this notation:
https://github.com/libvips/libvips/blob/master/libvips/resample/shrinkv_hwy.cpp#L62

Repartitioned types also use the same names:
https://github.com/google/highway/blob/master/hwy/examples/float_distribution.cc#L82

Maybe it is worth doing the same in the source code as libvips ?

jan-wassenberg:
Thanks for catching. I think this notation is potentially confusing: Rust SIMD and others have adopted a u8x16 notation meaning exactly 16 u8, not "as many u8 as u16". I think it would make more sense to have a h(alf) or q(uarter) suffix. For example, Rebind from u32 to u8 means 1/4 vector.

bkmgit:
Ok. If one rebinds from u32 to u8, one cannot put the contents directly into u8, so u8quarter is clearer but is a longer expression. Perhaps u8q?

jan-wassenberg (approved):
q is great :) Thanks for updating!

PR #3215, opened 2026-07-13
Add lagmma

RaviTriv:
Implementing `lgamma` from [wishlist](https://github.com/google/highway/blob/master/g3doc/op_wishlist.md?plain=1#L40).
- Also addressing [following up](https://github.com/google/highway/pull/3141#discussion_r3450712496) from `tgamma` PR to use `MaskedXorOr`.

Thanks for taking a look!

### DIFF
--- a/hwy/contrib/math/math-inl.h
+++ b/hwy/contrib/math/math-inl.h
@@ -393,10 +393,10 @@ HWY_NOINLINE V CallTgamma(const D d, VecArg<V> x) {
  * @return natural log of the absolute value of the gamma function of 'x'
  */
 template <class D, class V>
-HWY_INLINE V Lgamma(D d, V x);
+HWY_INLINE V LogGamma(D d, V x);
 template <class D, class V>
-HWY_NOINLINE V CallLgamma(const D d, VecArg<V> x) {
-  return Lgamma(d, x);
+HWY_NOINLINE V CallLogGamma(const D d, VecArg<V> x) {
+  return LogGamma(d, x);
 }
 
 /**
@@ -1927,7 +1927,7 @@ HWY_INLINE V Gamma(D d, V a) {
 }
 
 template <class D, class V = VFromD<D>, class M = MFromD<D>>
-HWY_INLINE V LogGamma(D d, V a) {
+HWY_INLINE V Lgamma(D d, V a) {
   using T = TFromD<D>;
   static_assert(IsFloat<T>(), "Only makes sense for floating-point");
   LgammaImpl<T> impl;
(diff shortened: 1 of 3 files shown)


hwy/contrib/math/math-inl.h
```diff
@@ -1798,12 +1920,68 @@ HWY_INLINE V Gamma(D d, V a) {
...
+  V acc = kZero;
+  for (int i = 0; i < LgammaImpl<T>::kReduceSteps; ++i) {
+    const M down = Ge(y, kTwo);
+    y = MaskedSubOr(y, down, y, kOne);
+    acc = MaskedAddOr(acc, down, acc, impl::Log(d, y));
+  }
+
+  // [1, 2): logGamma = (y-1)*(y-2)*MidPoly(y - 1.5) + acc.
+  const V t = Sub(y, kMidCenter);
+  const V zero_factors = Mul(Sub(y, kOne), Sub(y, kTwo));
```
jan-wassenberg: One of the sub could be fused away into MulSub, or preferably MulAdd(f, m, -a) which is more widely natively supported.

jan-wassenberg (changes requested):
Very nice! Just one tiny optimization, and a question: should we maybe call the user-facing function LogGamma, and the implementation Lgamma or LogGamma+Suffix? This might be more obvious. The comment can still mention that it corresponds to lgamma.

hwy/contrib/math/math-inl.h
RaviTriv: good catch!
- updated to use `MulAdd` on FMA targets
  - used fallback on non FMA targets as the intermediate rounding error was relatively large where lgamma is around 0

hwy/contrib/math/math-inl.h
```diff
@@ -1947,7 +1947,13 @@ HWY_INLINE V LogGamma(D d, V a) {
   const V w = MaskedSubOr(a, neg, kOne, a);
 
   // [0.5, 1): logGamma = (w-1)*LowPoly(w - 0.75).
-  const V low = Mul(Sub(w, kOne), impl.LowPoly(d, Sub(w, kLowCenter)));
+  const V low_poly = impl.LowPoly(d, Sub(w, kLowCenter));
+  V low;
+  if constexpr (HWY_NATIVE_FMA) {
+    low = MulAdd(w, low_poly, Neg(low_poly));
```
jan-wassenberg: Rather than MulAdd + Neg, we are better off using MulSub because that is actually native on some platforms (x86). Same below.
Even better, we can change to q = Sub(kTwo, y), then NegMulAdd(y, q, q) which is native on more platforms.

jan-wassenberg (changes requested):
Nice :) Got one more optimization suggestion:

hwy/contrib/math/math-inl.h
RaviTriv: thanks, updated, the Sub trick is neat!

jan-wassenberg (approved):
👍

PR #3223, opened 2026-07-20
Ensure type is specified

bkmgit:
Closes #3222

PR #3225, opened 2026-07-20
Fix build on ppc64/s390x

miladfarca:
Getting the following errors after 451144b:
```
error: implicit conversion loses integer precision: 'int' to 'const uint8_t' (aka 'const unsigned char')
```


jan-wassenberg:
Thanks for fixing :) #3223 also includes this change.

miladfarca:
Ah! I'll close this PR than, thank you.

Issue #2682, opened 2025-08-21
Any plan for float16_t vector math function(hwy/contrib/math/math-inl.h) support?


ixgbe:
Any plan for float16_t vector math function(hwy/contrib/math/math-inl.h) support? 

This could be used by numpy(numpy/_core/src/umath/loops_half.dispatch.c.src).

jan-wassenberg:
Hi, no plans for this at the moment but would welcome some patches to add it.

johnplatts:
One of the issues is that the Hypot implementation in math-inl.h requires that `MaxExponentField<T>() / 4 + MantissaBits<T>() > MaxExponentField<T>() / 2` which is true for F32 and F64 but not F16.

The best way to implement F16 Hypot is to promote both of the F16 vector operands to 1 or 2 F32 vectors, compute Hypot using F32 vectors, and then demote the result (which is in 1 or 2 F32 vectors) back to a F16.

dcsid:
I'd like to take this on, per @johnplatts's suggestion: add F16 overloads that promote to F32, reuse the existing F32 kernels, and demote the result.  This sidesteps the `Hypot` exponent-field constraint and stays within 1 ULP in F16. Native F16 polynomials could be a later optimization if the NumPy use case needs the throughput.

My plan is to start with a first PR with the promote/compute/demote adapter plus the `Exp`/`Log` family, with `math_test` coverage including `HWY_HAVE_FLOAT16=0` targets (as in #3113/#3117). Follow-up PRs batch in the remaining functions, with `Hypot` and other special cases last.

Sound reasonable? If so I'll start on the first PR.

jan-wassenberg:
Sounds good :) Unfortunately we currently lack an overload of OrderedDemote2To for f32->f16, so the demotions would probably use DemoteTo + Combine.

math-inl.h is quite big - what do you think of moving this into f16_math-inl.h or a similar name?

dcsid:
Works for me 👍  I'll put it in `f16_math-inl.h` and use `DemoteTo` + `Combine` for the demotion. Starting on the first PR.

PR #3228, opened 2026-07-22
Add float16 exp/log math functions via promotion to float32

dcsid:
Towards #2682 — first PR of the series discussed there: adds `float16_t` overloads of `Exp`, `Exp2`, `Expm1`, `Log`, `Log2`, `Log10` and `Log1p` in a new `hwy/contrib/math/f16_math-inl.h`, which promote to float32, evaluate the existing float32 kernels, and demote the result via `DemoteTo` + `Combine` as suggested (no `OrderedDemote2To` for f32->f16).

Notes:
- The overloads take `Simd<float16_t, N, kPow2>` so they are more specialized than the unconstrained templates in math-inl.h; a SFINAE-constrained overload with the same signature would be ambiguous. No changes to math-inl.h are needed.
- Works on all targets, including when `HWY_HAVE_FLOAT16` is 0 — only ops from the always-supported list in quick_reference.md are used. Vectors that fit in half of the widest vector use a single promotion; wider vectors (and RVV tags already at `HWY_MAX_POW2`, where `Rebind` to float would be illegal) take a per-half path.
- The new `f16_math_test` exhaustively tests all float16 bit patterns in each function's documented range, in vector batches so that lane placement is also verified. Max observed error is 1 ULP against a single-rounded double-precision reference, with infinities required to match exactly.

Follow-up PRs will add the remaining functions in batches, with `Hypot` and other special cases last.


hwy/contrib/math/f16_math-inl.h
```diff
@@ -0,0 +1,239 @@
...
+struct Log2Kernel {
+  template <class DF, class VF>
+  HWY_INLINE VF operator()(DF df, VF x) const {
+    return Log2(df, x);
+  }
+};
+
+}  // namespace f16_impl
+
+// These overloads take Simd<float16_t, N, kPow2> rather than a generic D with
```
jan-wassenberg: hm, this works, but what do you think of adding a HWY_IF_NOT_SPECIAL_FLOAT(T) constraint to the math-inl.h versions?

jan-wassenberg (commented):
Nice work, generally LGTM, just two quick questions.
What's the test runtime like for exhaustive enumeration? Should we also parallelize that?

dcsid:
Done in 3b32355d. Added `HWY_IF_NOT_SPECIAL_FLOAT_D` to the seven declarations in math-inl.h and switched the float16 overloads to plain `HWY_IF_F16_D`, dropping the `Simd<float16_t, N, kPow2>` specialization.

One wrinkle: the definitions are separated from the declarations, and a default template argument may not be restated, so the definitions repeat the `EnableIf` without the `= nullptr`.

dcsid:
Runtime is negligible: the full 65,536-pattern sweep is ~7 ms per function per target (~96 ms for the whole binary), so parallelizing doesn't seem worth it.

The red jobs look like infra: ppc64le (Power10) failed in `gtest_discover_tests` with empty output from `convert_test` under QEMU (a contrib-off job), which fail-fast then cancelled the sibling multiarch jobs; both AArch64 jobs ended with "runner has received a shutdown signal". Happy to have them re-run.

jan-wassenberg (dismissed):
Thanks for updating! A bit unfortunate about the default arg, ideally we wouldn't declare functions beforehand, but would define CallX after X(). I'll take a TODO to clean that up at some point.

jan-wassenberg:
Internal CI is failing on Arm:
f16x32: Exp(-13.882812) expected 9.536743E-07 actual 1.072884E-06 ulp 2 max ulp 1

Please update the ulp tolerance to 2 :)

jan-wassenberg (dismissed):
Thanks for updating :)

jan-wassenberg:
Still/again failing: HwyF16MathTestGroup/HwyF16MathTest.TestAllF16Exp/SVE2 (this is QEMU running on x86)
```
f16x32: Exp(-0.126953) expected 8.808594E-01 actual 8.823242E-01 ulp 3 max ulp 2
f16x32: Exp(9.703125) expected 1.636800E+04 actual 1.444800E+04 ulp 240 max ulp 2
f16x32: Exp(10.992188) expected 5.942400E+04 actual 5.241600E+04 ulp 219 max ulp 2
```
These are surprisingly large differences :/ We are comparing against an f64 reference, but surely f32 isn't quite that bad by comparison. Any thoughts on how this is happening?

dcsid:
Sorry, my subnormal explanation was wrong. Reverted the 2 ULP bump and fixed the real cause in 69b2c1a0.

The single-promotion path selected itself by MaxLanes, which on scalable targets is only an upper bound: for `CappedTag<float16_t, 32>` at VL=512, `Lanes(d)` is 32 but `Rebind<float, D>` yields 16 lanes. The kernel saw only the lower half, and DemoteTo duplicates its result into both register halves, so `out[i] = f(in[i-16])` , matching all your samples, all at lanes >= 16. Dispatch now checks that Rebind preserves `Lanes(d)`, with HWY_DASSERTs, and failures now print MaxLanes/pow2 since three tags here all print as f16x32.

I can't run SVE locally, so I'd appreciate another internal run.

jan-wassenberg:
Oh, good catch. Scary that completely-wrong result are only 240 ulps away - this can definitely happen with some approximations, though not what we are doing here, which is promoting.

So basically we are imposing the requirement that input f16 vectors be at most half-vectors. Usually what I'd suggest is starting with the largest type, f32, and then deriving the other tags from it. This means if we start with a legal f32 tag, then we can always derive an f16 tag from it (unless it was already pow=-3, but that would not be sensible anyway). WDYT?

dcsid:
Thanks, that's a much better way to think about it! I checked all SVE/RVV tags and vector lengths: `Rebind<float16_t, Df32>` always matches `Lanes(Df32)`, whereas the reverse is exactly what needed the predicate.

The public overloads take an f16 tag, so the adapter would derive `RepartitionToWide<D16>`, which is the per-half tag. Taken fully, that means always processing in halves and dropping the single-promotion path: simpler, lane-safe by construction, one extra kernel call for small vectors. Happy to do that in the trig PR (next up) and retire the predicate.

jan-wassenberg:
:) Yeah, always going with the PromoteLowerHalf/Combine SGTM, thanks!

Issue #3229, opened 2026-07-23
Proposal: add Lookup64 for efficient 64-entry byte tables


KimYannn:
## Motivation

Some byte-oriented codecs need to map SIMD lanes through a 64-entry table. Base64 encoding is a common example: after reshaping 48 input bytes into four vectors of 6-bit indices, each output vector indexes the same 64-byte alphabet.

Highway currently provides `Lookup8`, `Lookup16`, and `Lookup32`. For a 64-entry byte table, application code must instead classify ranges and add offsets, or compose multiple smaller lookups. On AArch64 NEON this leaves the native four-register `TBL` instruction (`vqtbl4q_u8`) inaccessible through Highway.

## Proposed API

Extend the existing memory-table lookup family with:

```cpp
template <class D, class VI>
Vec<D> Lookup64(D d, const TFromD<D>* table, VI indices);

template <class D>
constexpr bool CanLookup64(D d);
```

Proposed semantics match `Lookup32`:

- `TFromD<D>` and index lanes are 8-bit integers.
- `table` contains 64 elements and is 64-element aligned.
- Each result lane is `table[indices[i]]`.
- Results are undefined for indices greater than or equal to 64.
- `CanLookup64` reports whether the tag has an efficient supported implementation.

## Implementation sketch

- **AArch64 NEON, 128-bit vectors:** load four 16-byte table vectors and lower directly to `vqtbl4q_u8`. This supports full NEON vectors even though they contain only 16 byte lanes.
- **Fixed vectors with at least 32 byte lanes:** use one table vector when 64 lanes are available, otherwise use `TwoTablesLookupLanes` over two 32-byte halves.
- **Scalable vectors with at least 32 byte lanes:** load two 32-entry halves, perform two lookups, and select according to `index >= 32`.

A prototype compiles the AArch64 path to one four-register lookup per result vector:

```asm
tbl.16b v21, { v3, v4, v5, v6 }, v7
```

The prototype includes tests for indices spanning all four 16-byte quarters and documentation alongside `Lookup8/16/32`. The table tests pass for NEON_BF16, NEON, NEON_WITHOUT_AES, and the existing targets that do not support the operation safely skip it through target-aware test instantiation. AVX2 and AVX-512 compile probes also succeed through the generic implementation.

## Base64 validation

I tested the proposal in a portable Highway Base64 encoder on an Apple M4 Pro. All implementations used the same 16 MiB input, `std::string` allocation, warmup, and median of nine samples.

[...]

PR #3232, opened 2026-07-24
Add efficient 64-entry byte table lookups

KimYannn:
## Summary

- add a portable `Lookup64` API so byte-oriented codecs can use 64-entry tables without architecture-specific code
- lower 128-bit AArch64 lookups to four-register `TBL`, with safe fixed-width and scalable fallbacks plus capability checks, tests, and documentation

## Motivation

Base64 encoding maps 6-bit values through a 64-byte alphabet. Without this operation, Highway code must classify ranges and add offsets, leaving AArch64's `vqtbl4q_u8` inaccessible.

On an Apple M4 Pro with a 16 MiB input, replacing the offset-based translation with `Lookup64` improved Highway Base64 encoding from 17.97 to 20.99 GiB/s, matching an architecture-specific NEON implementation at 20.78 GiB/s.

## Validation

- all 24 `table_test` cases pass for NEON_BF16, NEON, NEON_WITHOUT_AES, and EMU128
- warnings-as-errors build passes for the affected target
- generated AArch64 code contains a single `tbl.16b {v1, v2, v3, v4}` per lookup
- AVX2 and AVX-512 compile probes pass through the portable implementation

Design discussion: #3229

💘 Generated with Crush

### DIFF
--- a/g3doc/quick_reference.md
+++ b/g3doc/quick_reference.md
@@ -2645,15 +2645,17 @@ The following `ReverseN` must not be called if `Lanes(D()) < N`:
     `TwoTablesLookupLanes(a, b, indices)` on RVV/SVE if `Lanes(d) <
     Lanes(DFromV<V>())`.
 
-Each of the `Lookup8`, `Lookup16`, `Lookup32` (let $X denote the 8/16/32) ops
-below return `GatherIndex(D(), tbl, indices)`, but are much more efficient, and
-are limited to $X elements. Results are undefined if any indices are >= $X. They
-are implemented using `TableLookupLanes` or `TwoTablesLookupLanes`. Let `T`
-denote `TFromD<D>`. These ops are guaranteed to work if `D` is a full vector,
-`HWY_TARGET != HWY_SCALAR` and `HWY_MIN_BYTES / sizeof(T) >= $X/2`. Use the
-constexpr function `CanLookup$X(D())` to verify this. Even if it returns false,
-the ops are still safe to call if `Lanes(D())` >= $X/2. Note that `tbl` must be
-$X-element aligned!
+Each of the `Lookup8`, `Lookup16`, `Lookup32`, `Lookup64` (let $X denote the
+8/16/32/64) ops below return `GatherIndex(D(), tbl, indices)`, but are much more
+efficient, and are limited to $X elements. Results are undefined if any indices
+are >= $X. They are implemented using `TableLookupLanes` or
+`TwoTablesLookupLanes`. Let `T` denote `TFromD<D>`. These ops are guaranteed to
+work if `D` is a full vector, `HWY_TARGET != HWY_SCALAR` and
+`HWY_MIN_BYTES / sizeof(T) >= $X/2`. `Lookup64` is also guaranteed for 128-bit
+AArch64 NEON vectors. Use the constexpr function `CanLookup$X(D())` to verify
+this. Even if it returns false, the ops are still safe to call if
+`Lanes(D()) >= $X/2`, except that `Lookup64` also supports full AArch64 NEON
+vectors. Note that `tbl` must be $X-element aligned!
 
 *   `D`: {u,i,f}{16,32,64} \
     <code>Vec&lt;D&gt; **Lookup8**(D, const TFromD&lt;D&gt;* tbl, VI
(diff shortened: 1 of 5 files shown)


KimYannn:
Updated `CanLookup64` to match the maintainer-recommended predicate from #3229, with corresponding test and documentation updates. All 24 table tests pass locally.

g3doc/quick_reference.md
```diff
@@ -2645,15 +2645,17 @@ The following `ReverseN` must not be called if `Lanes(D()) < N`:
...
-the ops are still safe to call if `Lanes(D())` >= $X/2. Note that `tbl` must be
-$X-element aligned!
+Each of the `Lookup8`, `Lookup16`, `Lookup32`, `Lookup64` (let $X denote the
+8/16/32/64) ops below return `GatherIndex(D(), tbl, indices)`, but are much more
+efficient, and are limited to $X elements. Results are undefined if any indices
+are >= $X. They are implemented using `TableLookupLanes` or
+`TwoTablesLookupLanes`. Let `T` denote `TFromD<D>`. These ops are guaranteed to
+work if `D` is a full vector, `HWY_TARGET != HWY_SCALAR` and
+`HWY_MIN_BYTES / sizeof(T) >= $X/2`. `Lookup64` is also guaranteed for 128-bit
+AArch64 NEON vectors. Use the constexpr function `CanLookup$X(D())` to verify
```
jan-wassenberg: Nit: NEON vectors.. if T is byte-sized.

g3doc/quick_reference.md
```diff
@@ -2645,15 +2645,17 @@ The following `ReverseN` must not be called if `Lanes(D()) < N`:
...
+this. Even if it returns false, the ops are still safe to call if
+`Lanes(D()) >= $X/2`, except that `Lookup64` also supports full AArch64 NEON
```
jan-wassenberg: The ", except that `Lookup64` also supports full AArch64 NEON vectors" could maybe be replaced with "or HWY_TARGET_IS_NEON && sizeof(T) == 1. But note that CanLookup64 would have returned true in this case, so maybe we should just remove this note?

hwy/ops/arm_neon-inl.h
```diff
@@ -6428,6 +6428,39 @@ HWY_API Vec128<T> TwoTablesLookupLanes(Vec128<T> a, Vec128<T> b,
...
+  }
+
+  const Full128<uint8_t> du8;
+  const auto t0 = Load(du8, reinterpret_cast<const uint8_t*>(table));
+  const auto t1 = Load(du8, reinterpret_cast<const uint8_t*>(table) + 16);
+  const auto t2 = Load(du8, reinterpret_cast<const uint8_t*>(table) + 32);
+  const auto t3 = Load(du8, reinterpret_cast<const uint8_t*>(table) + 48);
+  detail::Tuple4<uint8_t, 16> tables = {{{t0.raw, t1.raw, t2.raw, t3.raw}}};
+  const auto idx_u8 = BitCast(du8, indices);
+  return BitCast(d, Vec128<uint8_t>{vqtbl4q_u8(tables.raw, idx_u8.raw)});
```
jan-wassenberg: If d is a half-vector, I think the final BitCast would have to be ResizeBitCast, but no harm in always doing that, can just replace BitCast with ResizeBitCast.

hwy/ops/shared-inl.h
```diff
@@ -596,6 +596,14 @@ HWY_API constexpr bool CanLookup32(D d) {
          (HWY_HAVE_SCALABLE && detail::IsFull(d) && (sizeof(T) == 1));
 }
 
+// Returns whether `Lookup64` can definitely be used for vectors created from
+// tag `d`. May return a false negative for large scalable vectors.
+template <class D, typename T = TFromD<D>>
+HWY_API constexpr bool CanLookup64(D d) {
+  return (!HWY_HAVE_SCALABLE && MaxLanes(d) >= 32) ||
+         (HWY_TARGET_IS_NEON && detail::IsFull(d) && (sizeof(T) == 1));
```
jan-wassenberg: After the above fix, I think we can remove the detail::IsFull(d) condition.

jan-wassenberg (changes requested):
Nice! Some minor suggestions for updating:

g3doc/quick_reference.md
KimYannn: Updated the documentation to specify that the NEON exception applies when `T` is byte-sized.

g3doc/quick_reference.md
KimYannn: Removed the redundant exception note and now rely on `CanLookup64` for capability guidance.

hwy/ops/arm_neon-inl.h
KimYannn: Updated the native NEON implementation to use `ResizeBitCast` for both indices and results, removed the full-vector overload restriction, and added partial-vector coverage.

hwy/ops/shared-inl.h
KimYannn: Removed `detail::IsFull(d)` after adding partial-vector support. The NEON capability is additionally AArch64-gated because the native four-register lookup is AArch64-only.

jan-wassenberg (approved):
Nice work, great that we are also testing partial vectors.

PR #3235, opened 2026-07-27
Add  Base64 codec

KimYannn:
## Summary

- add strict RFC 4648 Base64 encode and decode helpers under `hwy/contrib/base64`
- encode 48-byte NEON64 blocks to 64 characters with `Lookup64`
- decode 64-character NEON64 blocks to 48 bytes with dual lookup tables, `TBL`/`TBX`, and `SLI` packing
- retain scalar tails and fallbacks, including canonical padding-bit validation
- integrate the implementation and tests with Bazel, CMake, and Meson

## Motivation

Base64 maps naturally to fixed-size SIMD blocks, but Highway does not currently provide a contrib codec. The NEON64 path reduces character classification, packing, and validation overhead while preserving strict decoding behavior.

On an Apple M4 Pro with Apple Clang, the optimized decode path reached about 16.2 GiB/s for 1 MiB and 16 MiB inputs. In the same local benchmark, simdutf 9.0.0 reached about 14.2 and 13.9 GiB/s, respectively.

## Validation

- RFC 4648 known vectors
- randomized round trips for input lengths 0 through 257
- invalid length, character, padding, canonical-bit, and non-ASCII cases
- invalid characters in every block of the 256-character validation batch
- all 12 target-parameterized tests passed on NEON_BF16, NEON, NEON_WITHOUT_AES, and EMU128
- `git diff --check`

### DIFF
--- a/hwy/contrib/base64/base64-inl.h
+++ b/hwy/contrib/base64/base64-inl.h
@@ -71,15 +71,15 @@ HWY_INLINE void EncodeBase64Tail(const uint8_t* HWY_RESTRICT input,
   output[3] = kAlphabet[b2 & 63];
 }
 
-#if HWY_ARCH_ARM_A64 && HWY_TARGET_IS_NEON
+#if (HWY_ARCH_ARM_A64 && HWY_TARGET_IS_NEON) || HWY_TARGET <= HWY_AVX3_DL
 
-// Encodes 48 input bytes to 64 output characters.
-HWY_INLINE void EncodeBase64Block(const uint8_t* HWY_RESTRICT input,
+// Encodes 3 * Lanes(d) input bytes to 4 * Lanes(d) output characters.
+template <class D>
+HWY_INLINE void EncodeBase64Block(D d, const uint8_t* HWY_RESTRICT input,
                                   char* HWY_RESTRICT output) {
-  const Full128<uint8_t> d;
-  Vec<decltype(d)> b0;
-  Vec<decltype(d)> b1;
-  Vec<decltype(d)> b2;
+  VFromD<D> b0;
+  VFromD<D> b1;
+  VFromD<D> b2;
   LoadInterleaved3(d, input, b0, b1, b2);
 
   const auto mask63 = Set(d, uint8_t{63});
@@ -99,7 +99,7 @@ HWY_INLINE void EncodeBase64Block(const uint8_t* HWY_RESTRICT input,
                     reinterpret_cast<uint8_t*>(output));
 }
 
-#endif  // HWY_ARCH_ARM_A64 && HWY_TARGET_IS_NEON
+#endif  // NEON64 || HWY_TARGET <= HWY_AVX3_DL
 
 template <class D>
 HWY_INLINE VFromD<D> DecodeBase64Vector(D d, VFromD<D> encoded) {
@@ -187,9 +187,13 @@ HWY_INLINE size_t Base64Encode(const uint8_t* HWY_RESTRICT input,
                                char* HWY_RESTRICT output) {
   size_t in = 0;
   size_t out = 0;
-#if HWY_ARCH_ARM_A64 && HWY_TARGET_IS_NEON
-  for (; in + 48 <= input_size; in += 48, out += 64) {
-    detail::EncodeBase64Block(input + in, output + out);
+#if (HWY_ARCH_ARM_A64 && HWY_TARGET_IS_NEON) || HWY_TARGET <= HWY_AVX3_DL
+  const ScalableTag<uint8_t> d;
+  const size_t input_block = 3 * Lanes(d);
+  const size_t output_block = 4 * Lanes(d);
+  for (; input_size - in >= input_block;
+       in += input_block, out += output_block) {
+    detail::EncodeBase64Block(d, input + in, output + out);
   }
 #endif
   for (; in + 3 <= input_size; in += 3, out += 4) {
--- a/hwy/contrib/base64/base64-inl.h
+++ b/hwy/contrib/base64/base64-inl.h
@@ -1,0 +1,271 @@
+// Copyright 2026 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Include guard (still compiled once per target)
+#if defined(HIGHWAY_HWY_CONTRIB_BASE64_BASE64_INL_H_) == \
+    defined(HWY_TARGET_TOGGLE)
+#ifdef HIGHWAY_HWY_CONTRIB_BASE64_BASE64_INL_H_
+#undef HIGHWAY_HWY_CONTRIB_BASE64_BASE64_INL_H_
+#else
+#define HIGHWAY_HWY_CONTRIB_BASE64_BASE64_INL_H_
+#endif
+
+#include <stddef.h>
+#include <stdint.h>
+
+#include "hwy/highway.h"
+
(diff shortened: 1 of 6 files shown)


hwy/contrib/base64/base64-inl.h
```diff
@@ -0,0 +1,271 @@
...
+    output[3] = '=';
+    return;
+  }
+
+  const uint8_t b2 = input[2];
+  output[2] = kAlphabet[((b1 & 15) << 2) | (b2 >> 6)];
+  output[3] = kAlphabet[b2 & 63];
+}
+
+#if HWY_ARCH_ARM_A64 && HWY_TARGET_IS_NEON
```
jan-wassenberg: This can/probably should also be enabled #if HWY_TARGET <= HWY_AVX3_DL.

hwy/contrib/base64/base64-inl.h
```diff
@@ -0,0 +1,271 @@
...
+                                  char* HWY_RESTRICT output) {
+  const Full128<uint8_t> d;
+  Vec<decltype(d)> b0;
+  Vec<decltype(d)> b1;
+  Vec<decltype(d)> b2;
+  LoadInterleaved3(d, input, b0, b1, b2);
+
+  const auto mask63 = Set(d, uint8_t{63});
+  const auto s0 = ShiftRight<2>(b0);
+  const auto s1 = And(Or(ShiftLeft<4>(b0), ShiftRight<4>(b1)), mask63);
```
jan-wassenberg: Mainly for AVX-512: you can fuse these into AndXor - I think the arguments to Or() are mutually exclusive, so should be the same.

hwy/contrib/base64/base64-inl.h
```diff
@@ -0,0 +1,271 @@
...
+      32,   33,   34,   35,   36,   37,   38,   39,   40,   41,   42,   43,  44,
+      45,   46,   47,   48,   49,   50,   51,   0x80, 0x80, 0x80, 0x80, 0x80};
+
+  const uint8x16x4_t low_table = {
+      {Load(d, kLow + 0).raw, Load(d, kLow + 16).raw, Load(d, kLow + 32).raw,
+       Load(d, kLow + 48).raw}};
+  const uint8x16x4_t high_table = {
+      {Load(d, kHigh + 0).raw, Load(d, kHigh + 16).raw, Load(d, kHigh + 32).raw,
+       Load(d, kHigh + 48).raw}};
+  const auto low = Vec128<uint8_t>{vqtbl4q_u8(low_table, encoded.raw)};
```
jan-wassenberg, line 196: Should we use Lookup64 here, at least as a non-Arm fallback? I suppose tbx takes care of merging results, maybe we want an Arm-specific path for that.

hwy/contrib/base64/base64-inl.h
```diff
@@ -0,0 +1,271 @@
...
+  Vec128<uint8_t> encoded2;
+  Vec128<uint8_t> encoded3;
+  LoadInterleaved4(d, reinterpret_cast<const uint8_t*>(input), encoded0,
+                   encoded1, encoded2, encoded3);
+
+  const auto sextets0 = DecodeBase64Vector(d, encoded0);
+  const auto sextets1 = DecodeBase64Vector(d, encoded1);
+  const auto sextets2 = DecodeBase64Vector(d, encoded2);
+  const auto sextets3 = DecodeBase64Vector(d, encoded3);
+  const auto invalid = Or(Or(Or(encoded0, sextets0), Or(encoded1, sextets1)),
```
jan-wassenberg: Can use Or3 to reduce the number of Or() (also only helps on AVX3).

hwy/contrib/base64/base64-inl.h
```diff
@@ -0,0 +1,271 @@
...
+                          Or(Or(encoded2, sextets2), Or(encoded3, sextets3)));
+
+  const auto out0 =
+      Vec128<uint8_t>{vsliq_n_u8(vshrq_n_u8(sextets1.raw, 4), sextets0.raw, 2)};
```
jan-wassenberg, line 255: Portable fallback here as well?

hwy/contrib/base64/base64-inl.h
```diff
@@ -0,0 +1,271 @@
...
+    const auto invalid0 =
+        detail::DecodeBase64Block(input + in + 0, output + out + 0);
+    const auto invalid1 =
+        detail::DecodeBase64Block(input + in + 64, output + out + 48);
+    const auto invalid2 =
+        detail::DecodeBase64Block(input + in + 128, output + out + 96);
+    const auto invalid3 =
+        detail::DecodeBase64Block(input + in + 192, output + out + 144);
+    const auto invalid = Or(Or(invalid0, invalid1), Or(invalid2, invalid3));
+    if (HWY_UNLIKELY(ReduceMax(d, invalid) >= 0x80)) {
```
jan-wassenberg: Maybe faster, likely clearer: `!AllFalse(d, Ge(invalid, 0x80))`.
Can also use Or3 above to reduce.

jan-wassenberg (changes requested):
Congrats on the speed result!

hwy/contrib/base64/base64-inl.h
KimYannn: iam not sure we could get benefits on x64 AVX512 but let me have a try

hwy/contrib/base64/base64-inl.h
```diff
@@ -0,0 +1,333 @@
...
+      {Load(d, kHigh + 0).raw, Load(d, kHigh + 16).raw, Load(d, kHigh + 32).raw,
+       Load(d, kHigh + 48).raw}};
+  const auto low = Vec128<uint8_t>{vqtbl4q_u8(low_table, encoded.raw)};
+  const auto high_index = Sub(encoded, Set(d, uint8_t{64}));
+  return Vec128<uint8_t>{vqtbx4q_u8(low.raw, high_table, high_index.raw)};
+#else
+  const auto index = And(encoded, Set(d, uint8_t{63}));
+  const auto low = Lookup64(d, kLow, index);
+  const auto high = Lookup64(d, kHigh, index);
+  return IfThenElse(Ne(And(encoded, Set(d, uint8_t{64})), Zero(d)), high, low);
```
jan-wassenberg: We could fuse Ne+And into TestBit, in case you'd like to make that change.

jan-wassenberg (approved):
Nice :) Looks like this would also do well on AVX-512 now.

hwy/contrib/base64/base64_test.cc
```diff
@@ -0,0 +1,178 @@
...
+
+HWY_NOINLINE void TestBase64Invalid() {
+  static const char* const kInvalid[] = {"A",        "AAA",
+                                         "====",     "=AAA",
+                                         "A=AA",     "AA=A",
+                                         "AAA==",    "AA==AAAA",
+                                         "AAA=AAAA", "AA A",
+                                         "AA?A",     "AB==",
+                                         "AAB=",     "Zm9vYmFyZm9vYmF?"};
+  uint8_t decoded[64];
```
jan-wassenberg, line 138: GCC static analysis complains about an overrun here:
https://github.com/google/highway/actions/runs/30338917241/job/90209902022?pr=3235

It might help to change loops from `i+N <= num` to `if (num >= N) for (; i <= num-N; i += N)` - the latter does not overflow.

jan-wassenberg (changes requested):

KimYannn:
> Nice :) Looks like this would also do well on AVX-512 now.

work but not with accepted performance, when compared to simduft, only 0.85x

jan-wassenberg:
hm, OK. Do you know how their approach differs? Can that be expressed using Highway, or are we missing any ops?

KimYannn:
> hm, OK. Do you know how their approach differs? Can that be expressed using Highway, or are we missing any ops?

seems we need  lookup128 for avx512

jan-wassenberg:
I'd welcome Lookup128, added the same way you added Lookup64, if that is helpful.
It seems we could also already land this PR as-is? What's the current perf gap?

KimYannn:
> I'd welcome Lookup128, added the same way you added Lookup64, if that is helpful. It seems we could also already land this PR as-is? What's the current perf gap?

 tried implementing Lookup128 in the same way simdutf uses it, and the only performance improvement was observed with 4K blocks. For larger blocks, such as 4M, there was a 5% performance degradation. so i mark this as ready now

jan-wassenberg (approved):

PR #3244, opened 2026-08-02
perf(phast): prune builder collision and scan tests by slice structure

### LINKED ISSUE #3243: phast: TryPlaceBucket runs 894M pairwise comparisons per 1M-key build for 937 hits
## Summary

`PerWorkerBuilder::TryPlaceBucket` tests every one of a bucket's
`b_size*(b_size-1)/2` key pairs for a shared slot, and repeats that test for each
of the `num_candidates` seeds:

https://github.com/google/highway/blob/d043aee/hwy/contrib/hash/phast.cc#L429-L437

Instrumented over a full 1,000,000-key `BuildPhast` on `d043aee`, that loop
executes **894,081,141 comparisons and finds 937 duplicates**. At 200,000 keys,
480,728,084 comparisons for 2,939 duplicates.

## Why nearly all of it is skippable

`Phast::PosFromHashAndSeed` (`phast-inl.h:107-112`) is

```
LemireMod(hash, num_slice_offsets) + (Placement(hash, seed) & slice_mask)
```

The first term does not depend on the seed and the second is masked to
`slice_length`, so every position reachable from one hash lies in
`[LemireMod(hash, num_slice_offsets), + slice_length)`.

Two keys can therefore only share a slot if those ranges overlap. Writing `b_i`,
`b_j` for the range bases and `W` for `slice_length`, if `|b_i - b_j| >= W` then
for any seed `s`

```
Pos(h_i, s) = b_i + psi_i < b_i + W <= b_j <= b_j + psi_j = Pos(h_j, s)
```

so the pair cannot collide at *any* seed. That is decidable once per bucket from
two hash-derived integers rather than retested per seed. Measured over the same
1,000,000-key build, **48,796 of 8,571,328 pairs overlap — 0.569%**. At 200,000
keys, 121,497 of 4,119,273.

The bound is `< W`, not `<= W` or `< W - 1`: ranges exactly `W - 1` apart share
one slot and that slot is reachable. Witness at a config the builder enumerates
(`num_keys` 200, headroom 1%, `num_slots` 203, `slice_length` 64):

```
h_i = 0x000c0000   h_j = 0x73570000   seed = 0
range bases 0 and 63, exactly slice_length - 1 apart
both positions equal 63
```

Substituting `< slice_length - 1` makes the existing `TestMultipleSizes` fail
with `Collision detected` at 2,048 keys or fewer.

## The same structure applies to `TryCuckooSwap`

It scans up to `kScanWindow` = 50,000 committed buckets, testing each scanned
position for membership in `all_pos_` with `Find`, a linear SIMD scan of
`fail_b_size * 256` elements whose result is used only as a boolean:

https://github.com/google/highway/blob/d043aee/hwy/contrib/hash/phast.cc#L298-L312

`all_pos_` is covered by `fail_b_size` ranges spanning
`fail_b_size * slice_length / num_slots` of the table — 0.402% at 1,000,000 keys
with 2% headroom — so most scanned positions can be rejected before `Find`.

## Measured effect

Restricting the pair test to overlapping ranges and bounds-checking the scan,
both exact rejections, on clang 22.1.7 `x86_64-pc-windows-msvc` `-O2`, AVX2,
Windows 11, 9 workers:

| num_keys | comparisons `d043aee` | patched | build time `d043aee` | patched |
|---|---|---|---|---|
| 200,000 | 480,728,084 | 18,265,157 | 324.5 ms | 204.6 ms |
| 1,000,000 | 894,081,141 | 13,643,737 | 835.8 ms | 507.5 ms |

Six paired interleaved runs at 1,000,000 keys, 7 repetitions each, give 1.65x to
1.98x on the minimum. Both builds report the same duplicate counts, 2,939 and
937, and an FNV-1a 64 digest over the full config and seed array matches on 42/42
key sets across 252 builds.

## Limits

Timings are one machine, one compiler, one OS and the AVX2 target only, and the
machine was not quiesced. Comparison and duplicate counts are exact instrumented
counts. `b_size` was 2 to 12 on every build measured; `kMaxHashesPerBucket` is 32
and once-per-bucket collection cost grows as `b_size^2`, a regime not exercised
here. Separately, `hwy/contrib/hash/hash_test.cc` does not build under clang-cl
on this setup (19 `always_inline ... requires target feature 'ssse3'` errors) on
`d043aee` itself, noted only so it is not attributed to a patch.

Draft PR: #3244.

teerthsharma:
## Summary

`Phast::PosFromHashAndSeed` is `LemireMod(hash, num_slice_offsets)` plus
`Placement(hash, seed) & slice_mask`. The first term does not depend on the seed
and the second is masked to `slice_length`, so every position reachable from one
hash lies in a range of width `slice_length` fixed by the hash alone.

`TryPlaceBucket` tests all `b_size*(b_size-1)/2` key pairs of a bucket for a
shared slot, once per candidate seed. Two keys can only collide if their ranges
overlap, so this change collects the overlapping pairs once per bucket and tests
only those. `TryCuckooSwap` calls `Find`, a linear scan of `fail_b_size * 256`
elements, on every scanned position; `all_pos_` lies inside `fail_b_size` ranges,
so this change bounds-checks first.

Both are exact rejections, not heuristics. Seed selection, placement order,
config selection and the returned `PhastData` are unchanged. No public API,
header, storage-format or query-path change.

Fixes #3243.

## Comparison counts

Atomic counters in the pairwise check, whole builds, 9 workers.

| num_keys | `d043aee` | PR head (collect + per-seed) | reduction | duplicates found |
|---|---|---|---|---|
| 200,000 | 480,728,084 | 4,119,273 + 14,145,884 = 18,265,157 | 26.3x | 2,939 on both |
| 1,000,000 | 894,081,141 | 8,571,056 + 5,072,681 = 13,643,737 | 65.5x | 937 on both |

Overlapping pairs: 121,497 of 4,119,273 at 200,000 keys, 48,796 of 8,571,328 at
1,000,000. Equal duplicate counts on both builds is the restriction confirmed on
real inputs rather than only on constructed pairs.

## Build time

`d043aee` vs PR head, clang 22.1.7 `x86_64-pc-windows-msvc`, `-O2`, AVX2,
Windows 11, 9 workers, `payload_bytes` = 4. Both binaries link the same
`hwy.lib`; only `hwy_contrib.lib` differs. Six paired interleaved runs at
1,000,000 keys, 7 repetitions each, alternating arms so both see the same
machine state:

| pair | `d043aee` min / median | PR head min / median | ratio min / median |
|---|---|---|---|
| 1 | 1047.2 / 1510.9 ms | 527.9 / 587.9 ms | 1.98x / 2.57x |
| 2 | 903.9 / 1062.2 ms | 501.1 / 514.5 ms | 1.80x / 2.06x |
| 3 | 880.5 / 986.4 ms | 528.8 / 551.3 ms | 1.66x / 1.79x |
| 4 | 964.0 / 1073.9 ms | 538.7 / 549.7 ms | 1.79x / 1.95x |
| 5 | 835.8 / 884.3 ms | 507.5 / 532.2 ms | 1.65x / 1.66x |
| 6 | 883.7 / 929.3 ms | 508.6 / 526.2 ms | 1.74x / 1.77x |

Worst pair 1.65x on the minimum. At 200,000 keys, 324.5 -> 204.6 ms minimum. At
10,000 keys, 4.9 -> 3.3 ms; greedy placement almost always succeeds there so
`TryCuckooSwap` rarely runs and the gain comes from `TryPlaceBucket` alone.
Isolating the two changes at 1,000,000 keys, medians over three rounds: the scan
bound alone takes 833 -> 490, 805 -> 483, 837 -> 523 ms, and the pair collection
takes those to 462, 453, 444 ms.

## Tests

`phast_test.cc` gains three tests calling `Phast::PosFromHashAndSeed` and
`PhastPlacement` directly rather than reimplementing them. Suite goes 3 -> 6.

| test | asserts | scale |
|---|---|---|
| `TestSliceInvariant` | `base <= Pos < base + slice_length`, `Pos < num_slots` | 1,664,000 (hash, seed) pairs, 4 `num_slots` x 4 `slice_length` |
| `TestDisjointSlicesNeverCollide` | disjoint ranges never collide at any of 256 seeds | 19,929 random pairs plus 3,558 constructed pairs exactly `slice_length` apart |
| `TestSliceBoundIsTight` | the bound is `< slice_length`, not `< slice_length - 1` | one recorded witness |

Each was verified to fail before being relied on. Mutating
`Placement(hash, seed) & pp.slice_mask` to `% (pp.slice_mask + 2)` in
`phast-inl.h` aborts all three:

```
TestSliceInvariant               pos - base < slice_length: position beyond its slice
TestDisjointSlicesNeverCollide   slices exactly slice_length apart must not collide
TestSliceBoundIsTight            witness pair must collide; the bound cannot be tightened
```

Mutating the `TryPlaceBucket` predicate to `< slice_length - 1` makes the
existing `TestMultipleSizes` fail with `Collision detected` at 2,048 keys or
fewer, so the off-by-one is a correctness bug and the pre-existing suite already
catches it. The same mutation on the `TryCuckooSwap` bound leaves all 6 tests
passing — it changes which swaps run and so which seeds are chosen, while still
producing a valid perfect hash — and is caught only by the digest below. The two
bounds fail differently and are covered accordingly.

Random sampling is blind to that boundary: the set of hash-block gaps at exactly
`slice_length - 1` has width `65536 / num_slice_offsets`, which is 0.064 at 1M
keys, so the case is unreachable there and reachable only at the small configs
`MinSliceLength` selects below 1,300 keys. `TestSliceBoundIsTight` uses a
constructed witness for that reason.

Buildable targets on this setup, PR head: `phast_test` 6/6, `cuckoo2x2_test` 3/3,
`shardmul_test` 3/3. Skips are the EMU128 variants, skipped on `d043aee` too.

## Output identity

An FNV-1a 64 digest over `num_slots`, `hash_key`, `bucket_mask`,
`num_slice_offsets`, `slice_mask`, `config_idx`, `attempt_idx` and every packed
seed word `GatherSeeds` reads. 14 key-set sizes (1, 2, 17, 63, 64, 100, 511,
1000, 4096, 9500, 12000, 50000, 140000, 200000) x 3 key-set seeds = 42 cases, 3
repeated builds each, against both libraries: 252 builds. 42/42 digests match
`d043aee`; 0/126 repeated builds per binary differed from their own first build.

## Rejected alternatives

A bitset over `[0, num_slots)` for the `TryCuckooSwap` test, following the
existing `Occupancy` pattern, makes membership exact and drops the `Find`
fallback, but the 256-seed orbit visits 242 distinct positions in a 2048-wide
range, so the range check already rejects 99.60% of probes and the bitset would
filter the residual 0.40% at the cost of a `num_slots`-sized allocation and a
per-call reset. Sorting the range bases before collecting pairs would give
`O(b_size log b_size + pairs)` instead of `O(b_size^2)`, but `b_size` is 2 to 12
on every build measured and `kMaxHashesPerBucket` is 32, so `b_size^2/2 <= 496`
beats a sort at every reachable size.

## Limits

Timings are one machine, one compiler, one OS and the AVX2 target only, and the
machine was not quiesced, hence paired interleaved runs rather than a single
median. Comparison, overlap and duplicate counts are exact instrumented counts
and carry no such caveat. `b_size` was 2 to 12 on every build measured;
collection cost grows as `b_size^2` and `kMaxHashesPerBucket` is 32, a regime not
exercised here. Determinism is conditional on worker count on `d043aee` and on
this change alike, since `reps_per_config` derives from `pool.NumWorkers()`.
Two failures visible around this PR are pre-existing and unrelated to it.
`hwy/contrib/hash/hash_test.cc` does not build under clang-cl on the machine used
above (19 `always_inline ... requires target feature 'ssse3'` errors); verified
by stashing this change and rebuilding on `d043aee`, which produces the same 19
errors. In CI, `Build and test GCC-16 on AArch64` (`multiarch.yml`, job `aarch64_cmake`,
`runs-on: ubuntu-26.04-arm`) is red here, and reproduces identically on the base
commit:

| | `d043aee` master, job 91194654723 | this PR, job 91441853717 |
|---|---|---|
| Build step conclusion | `cancelled` | `cancelled` |
| duration | 33m30s | 36m01s |
| last line | `[ 57%] ... masked_arithmetic_test.cc.o` | `[ 57%] ... masked_arithmetic_test.cc.o` |
| compile errors | 0 | 0 |
| tests executed | 0 | 0 |

Both stop at the same percentage on the same translation unit after ~35 minutes,
so this is the runner being reclaimed at a fixed point in the build rather than
an intermittent fault. In this PR's run nothing else failed — the sibling
`Clang-22 on AArch64` succeeded at 02:25:54, after the GCC-16 build was already
cancelled at 02:18:35, and the other 18 jobs in run 30727519527 all passed — so
neither `fail-fast` nor `cancel-in-progress` explains it. `hwy/contrib/hash/phast.cc`
itself compiled cleanly on that toolchain, at 23%, under `-ftrapv` and
`-DHWY_WARNINGS_ARE_ERRORS=ON`, roughly 34 minutes before the job died.

### DIFF
--- a/hwy/contrib/hash/phast.cc
+++ b/hwy/contrib/hash/phast.cc
@@ -289,6 +289,14 @@ class PerWorkerBuilder {
     const size_t fail_b_size = ComputeBucketPos(rank);
     const size_t num_pos_to_find = fail_b_size * 256;
 
+    // all_pos_ is covered by the fail_b_size slices of this bucket, which
+    // together span at most fail_b_size * slice_length of num_slots - about
+    // 0.4% for 1M keys. Positions outside them cannot be in all_pos_, so the
+    // Find below is skipped for the vast majority of scanned positions.
+    uint32_t slice_begin[kMaxHashesPerBucket];
+    ComputeSliceBegins(rank, fail_b_size, slice_begin);
+    const uint32_t slice_length = config_.placement.slice_mask + 1;
+
     // Scan recent buckets for those whose committed positions match any
     // target position (i.e., they block a candidate seed).
     const ScalableTag<uint32_t> du32;
@@ -307,6 +315,12 @@ class PerWorkerBuilder {
       bool matches = false;
       for (size_t idx = 0; idx < scan_size && !matches; ++idx) {
         const uint32_t pos = ComputeOnePos(scan_begin + idx, scan_seed);
+        // Unsigned wraparound also rejects pos < slice_begin[i].
+        bool in_slice = false;
+        for (size_t i = 0; i < fail_b_size && !in_slice; ++i) {
+          in_slice = (pos - slice_begin[i]) < slice_length;
+        }
+        if (!in_slice) continue;
         matches = Find(du32, pos, all_pos_.data(), num_pos_to_find) !=
                   num_pos_to_find;
       }
@@ -352,6 +366,23 @@ class PerWorkerBuilder {
     return false;
   }
 
+  // Writes, for each key of the bucket at `rank`, the first slot it can occupy.
+  // PosFromHashAndSeed is LemireMod(hash, num_slice_offsets) + (Placement(hash,
+  // seed) & slice_mask): the first term does not depend on the seed and the
+  // second is less than slice_length, so every seed places that key somewhere
+  // in [slice_begin, slice_begin + slice_length). Two keys whose slices are
+  // disjoint therefore cannot share a slot at any seed.
+  void ComputeSliceBegins(uint32_t rank, size_t b_size,
+                          uint32_t* HWY_RESTRICT slice_begin) const {
+    const uint32_t bucket_idx = bucket_idx_largest_first_[rank];
+    const size_t b_begin = total_hashes_before_bucket_[bucket_idx];
+    for (size_t idx = 0; idx < b_size; ++idx) {
+      const uint32_t key_idx = key_idx_for_pos_[b_begin + idx];
+      slice_begin[idx] = LemireMod(hash_for_key_idx_[key_idx],
+                                   config_.placement.num_slice_offsets);
+    }
+  }
+
   // Computes the position for a single hash+seed pair (scalar).
   // Much cheaper than ComputeBucketPos, which computes all 256 seeds.
   uint32_t ComputeOnePos(size_t b_pos, uint32_t seed) const {
@@ -394,6 +425,29 @@ class PerWorkerBuilder {
     const size_t b_size = ComputeBucketPos(rank);
     const size_t num_candidates = WriteSeedCandidates(b_size, seed_candidates);
 
+    // Two keys of this bucket can only share a slot if their slices overlap
+    // (see ComputeSliceBegins). Collecting those pairs once replaces the
+    // b_size*(b_size-1)/2 pairwise test that would otherwise repeat for each of
+    // the num_candidates seeds. Measured over a 1M-key build, 48,796 of
+    // 8,571,328 pairs overlap - 0.57% - so the inner check below is usually
+    // empty and the quadratic term disappears entirely for that bucket.
+    uint16_t overlap[kMaxHashesPerBucket * (kMaxHashesPerBucket - 1) / 2];
+    size_t num_overlap = 0;
+    if (b_size >= 2 && num_candidates != 0) {
+      uint32_t slice_begin[kMaxHashesPerBucket];
+      ComputeSliceBegins(rank, b_size, slice_begin);
+      const uint32_t slice_length = config_.placement.slice_mask + 1;
+      for (size_t i = 0; i < b_size; ++i) {
+        for (size_t j = i + 1; j < b_size; ++j) {
+          const uint32_t bi = slice_begin[i], bj = slice_begin[j];
+          const uint32_t d = bi > bj ? bi - bj : bj - bi;
+          if (d < slice_length) {
+            overlap[num_overlap++] = static_cast<uint16_t>((i << 8) | j);
+          }
+        }
+      }
+    }
+
     uint32_t best_seed = 0;
     uint32_t best_cost = ~0u;
 
@@ -402,12 +456,13 @@ class PerWorkerBuilder {
       if (seed == excluded_seed) continue;
 
       if (b_size >= 2) {
-        // Pairwise collision check.
+        // Pairwise collision check, restricted to pairs with overlapping
+        // slices; the rest cannot collide at this or any other seed.
         bool has_dup = false;
-        for (size_t i = 0; i < b_size && !has_dup; ++i) {
-          for (size_t j = i + 1; j < b_size && !has_dup; ++j) {
-            has_dup = (all_pos_[i * 256 + seed] == all_pos_[j * 256 + seed]);
-          }
+        for (size_t p = 0; p < num_overlap && !has_dup; ++p) {
+          const size_t i = overlap[p] >> 8;
+          const size_t j = overlap[p] & 0xFF;
+          has_dup = (all_pos_[i * 256 + seed] == all_pos_[j * 256 + seed]);
         }
         if (has_dup) continue;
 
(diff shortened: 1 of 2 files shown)


jan-wassenberg (approved):
Interesting observation, thanks! Sounds like an LLM participated - I'm curious which one?

teerthsharma:
> Interesting observation, thanks! Sounds like an LLM participated - I'm curious which one?

Opus 5 at high to handle edge cases multiple different methodologies three arm tests (courtesy of anthropic open source contributors program ) It build three main observations based on my other works ! I too got LLM access of this tier only recently hope it was helpful

PR #3245, opened 2026-08-03
WIP: Gol bit bash

bkmgit:
Game of life using bitboards.

There is a bug with one of the indices for array copies - working on fixing it.  Still also need to add remainder processing.

Other than demonstrating bit shift techniques, it could also be used to demonstrate auto-tuning.  There are three choices of kernels for the update step, and when there are long vectors, doing more than one step before doing an update can save on loads and stores.  The best choice of kernel and number of steps to take may differ from one architecture to another.

Can also leave it out if it does not seem useful or is too similar to #3208


jan-wassenberg:
As discussed, I'd propose to skip this one - it is both similar to game_of_life, and maybe too niche for a beginner's tutorial.

bkmgit:
Ok. Speed of SIMD bit CPU implementation is about the same as in https://github.com/binary-banter/fast-game-of-life/blob/master/src/simd.rs

PR #3247, opened 2026-08-04
Add baker mix example

bkmgit:
Demonstrate use of iota, print debugging and loading of half vectors by permuting a list of ordered numbers.


jan-wassenberg (approved):
Nice idea, thanks for adding!

PR #3248, opened 2026-08-06
Run only boundary compilers on pull requests

### LINKED ISSUE #3025: Minimize testing time while still ensuring quality code
Change GitHub actions to run on newest and oldest compilers for pull requests, but enable testing on all compiler releases before new releases.

NguyenTienDat377:
Fixes #3025

Pull requests now build only the oldest and newest compilers in `build_test.yml`.
The full matrix still runs on manual dispatch and weekly on Mondays, so no
configuration is removed from CI - it is only deferred off the pull request path.

### Mechanism

GitHub Actions cannot apply `if:` to an individual matrix entry, only to a whole
job. Each affected job is therefore split into a boundary half that always runs
and a gated half carrying:

```yaml
if: github.event_name != 'pull_request'
```

The `steps:` blocks are identical between the two halves; only the `include:`
list differs.

| Job | Entries | On pull requests |
|---|---|---|
| `cmake` (22.04) | 5 | always |
| `cmake_rest` (22.04) | 7 | gated |
| `cmake_ubuntu_2404` | 8 | gated |
| `cmake_ubuntu_2604` | 5 | always |
| `cmake_ubuntu_2604_rest` | 2 | gated |
| `bazel` | - | always |

`cmake_ubuntu_2404` is gated whole rather than split, since Clang-16..19 and
GCC-13/14 are all interior to the range already covered by the 22.04 and 26.04
boundary jobs.

### What the boundary set retains

Beyond the oldest and newest compilers, the retained set keeps one build per C++
standard (11, 20, 23), the 32-bit build, and the VQSort-disabled build. Those
axes are independent of compiler version - standard-conditional code in the
headers can break without any compiler changing - so reducing them along with
the compiler axis would lose real coverage rather than duplicate coverage.

### Trigger change

The `push` trigger is dropped in favour of `schedule` + `workflow_dispatch`.
Previously `on: [push, pull_request]` ran every job twice for same-repo
branches. Per discussion in #3025 the full matrix is deliberately *not* run on
pushes to master, as those are too frequent.

Also narrows the workflow token to contents: read and documents the manual
trigger in g3doc/release_testing_process.md.

### Verification

Both branches of the condition were exercised on a fork:

- `workflow_dispatch` - 28 jobs (27 matrix entries + bazel), all green
- `pull_request` - 11 jobs, all green, with the three gated jobs reported as
  skipped

Job count per pull request goes from 28 to 11.

Only `build_test.yml` is changed here. `meson_build_test.yml` and
`multiarch.yml` can follow in separate PRs if this approach looks right.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -12,9 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-permissions: read-all
+permissions:
+  contents: read
 name: Build / test
-on: [push, pull_request]
+on:
+  pull_request:
+  schedule:
+    - cron: '0 3 * * 6'   # Saturdays 03:00 UTC
+  workflow_dispatch:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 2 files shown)


bkmgit:
Perhaps run the weekly check on Saturday or Sunday rather than Monday,  there are fewer merges on these days.

.github/workflows/build_test.yml
```diff
@@ -14,7 +14,11 @@
 
 permissions: read-all
```
jan-wassenberg: We have a new vulnerability scanner that is now flagging this.
Please change to permissions: contents: read (see https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defining-access-for-the-github_token-scopes-1)

jan-wassenberg (changes requested):
This is great, thank you for making it happen.
Please also update g3doc/release_testing_process.md with the steps required to trigger.

.github/workflows/build_test.yml
NguyenTienDat377: Thanks! Done — narrowed to permissions: contents: read, documented the manual trigger in g3doc/release_testing_process.md, and moved the weekly run to Saturday per @bkmgit's suggestion.

jan-wassenberg (approved):
Thanks for updating!

PR #3252, opened 2026-08-06
Fix LASX Tgamma failure and enable runtime dispatch for GCC >= 15.3

xry111:
Fix LASX NegMulAdd implementation:
    
    The tgamma implmentation (added in #3141) expects it to be fused
    (i.e. only the final result is rounded and the intermediate result
    from multiplication should be treated as infinite precesion), but
    the LoongArch implementation was unfused.
    
    Use the LASX instruction with fused semantics instead, similar to
    the LSX implementation added in #3141.

Dispatch LSX/LASX at runtime with GCC >= 15.3:
    
    Commit 10d7ab41fda3 ("Dispatch LSX/LASX at runtime") implemented the
    dispatch of LSX/LASX optimized routines with Clang.  Unfortunately GCC
    had a bug so the dispatch was not implemented for GCC at the time.
    Now the bug is fixed with GCC 15.3 and newer, so enable the runtime
    dispatch with GCC >= 15.3 too.
    
    As the GCC target pragma automatically updates the status of
    __loongarch_sx and __loongarch_asx, guard the definition of them to
    prevent a macro redefinition warning.

### DIFF
--- a/hwy/detect_targets.h
+++ b/hwy/detect_targets.h
@@ -357,12 +357,14 @@
 #ifndef HWY_BROKEN_LOONGARCH  // allow override
 // Using __loongarch_sx and __loongarch_asx macros to
 // check whether LSX/LASX targets are available.
-// GCC does not work yet, see https://gcc.gnu.org/PR121875.
+// GCC < 15.3 does not work, see https://gcc.gnu.org/PR121875.
 #if !defined(__loongarch_sx) && \
-    !(HWY_COMPILER_CLANG && HWY_COMPILER_CLANG >= 1800)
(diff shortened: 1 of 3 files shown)


RaviTriv:
thanks for fixing this!

xry111:
Oops, somehow `git add`ed a #error only intended for debugging... I'll remove it.

jan-wassenberg (approved):
Nice, thank you :)
Fine for now, but FYI `HWY_COMPILER_CLANG && HWY_COMPILER_CLANG >= 1800` can be shortened to the equivalent `HWY_COMPILER_CLANG >= 1800`.

PR #3253, opened 2026-08-06
Add IndexOfMin and IndexOfMax to hwy/contrib/algo

Wint3rNight:
From the op wishlist. minmax-inl.h already has MinValue/MaxValue which return the value, so these return where it is. Ties go to the lowest index, matching std::min_element.

Each lane keeps its best value plus the block it came from, then a final pass picks the smallest value, earliest block, lowest lane. Lanes hold the block number rather than the element index so it fits in the lane type — for 8-bit lanes that only reaches 255 blocks, so the loop works in segments and carries the base across. Wider lanes run as a single segment.

Roughly 10x std::min_element on 4M floats (1.46 ms vs 14.31 ms), and about 2x the cost of MinValue — the extra compare and two selects per vector, plus this isn't unrolled the way MinValue is. Happy to add the 4x unroll if you'd like it in the same PR.

Tests follow the existing minmax_value_test.cc pattern against a scalar reference, plus one that walks a single extreme across a span long enough to cross a segment boundary on 8-bit lanes.

### DIFF
--- a/hwy/contrib/algo/minmax-inl.h
+++ b/hwy/contrib/algo/minmax-inl.h
@@ -94,6 +94,126 @@ T MaxValue(D d, const T* HWY_RESTRICT in, size_t count) {
   return ReduceMax(d, acc0);
 }
 
+// Returns the index of the first occurrence of the minimum value in
+// `in[0, count)`, or `count` if `count == 0`. Ties resolve to the lowest index,
+// matching `std::min_element`.
+template <class D, typename T = TFromD<D>>
+size_t IndexOfMin(D d, const T* HWY_RESTRICT in, size_t count) {
+  if (HWY_UNLIKELY(count == 0)) return count;
+
+  const RebindToUnsigned<D> du;
+  using TU = TFromD<decltype(du)>;
+  const size_t N = Lanes(d);
+
+  // Lanes record which block held their best value, so a segment can span at
+  // most this many blocks before that counter would wrap. Capped so neither the
+  // count nor the multiply below can overflow; for 32-bit and wider lanes this
+  // is a single segment in practice.
+  const uint64_t block_limit = static_cast<uint64_t>(LimitsMax<TU>());
+  const size_t max_blocks = static_cast<size_t>(
+      block_limit < (uint64_t{1} << 20) ? block_limit : (uint64_t{1} << 20));
+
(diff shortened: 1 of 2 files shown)


hwy/contrib/algo/minmax-inl.h
```diff
@@ -94,6 +94,126 @@ T MaxValue(D d, const T* HWY_RESTRICT in, size_t count) {
...
+    }
+
+    // Resolve lanes: smallest value, then earliest block, then lowest lane.
+    const T seg_min = ReduceMin(d, acc);
+    const Mask<D> is_min = Eq(acc, Set(d, seg_min));
+    const VFromD<decltype(du)> cand =
+        IfThenElse(RebindMask(du, is_min), blocks, Set(du, LimitsMax<TU>()));
+    const TU min_block = ReduceMin(du, cand);
+    const Mask<D> winners =
+        And(is_min, RebindMask(d, Eq(blocks, Set(du, min_block))));
```
jan-wassenberg, line 145: Minor: could be RebindMask(d, MaskedEq(RebindMask(du, is_min), blocks, Set(du, min_block))).

hwy/contrib/algo/minmax-inl.h
```diff
@@ -94,6 +94,126 @@ T MaxValue(D d, const T* HWY_RESTRICT in, size_t count) {
...
+      const Mask<D> lt = Lt(v, acc);
+      acc = IfThenElse(lt, v, acc);
+      blocks = IfThenElse(RebindMask(du, lt), Set(du, block), blocks);
+    }
+
+    // Resolve lanes: smallest value, then earliest block, then lowest lane.
+    const T seg_min = ReduceMin(d, acc);
+    const Mask<D> is_min = Eq(acc, Set(d, seg_min));
+    const VFromD<decltype(du)> cand =
+        IfThenElse(RebindMask(du, is_min), blocks, Set(du, LimitsMax<TU>()));
```
jan-wassenberg, line 142: I think this could also be IfThenElseZero, then we use ReduceMax.

jan-wassenberg (approved):
Nice work! We can land as-is, I've left two minor suggestions which might help a tiny bit. More important would be to 2x unroll. Could be done in a follow-up if you like.

Wint3rNight:
Thanks! I'll do a follow-up with both suggestions and the 2x unroll.

PR #3257, opened 2026-08-07
Re-implement BitsFromMask() for 8-bit lanes on ARM Neon 64-bit

orestispol:
This optimizes BitsFromMask for ARM64 (AArch64) NEON by using a double 8-bit movemask (via vqtbl1q_u8 shuffle and mask) combined with a 16-bit horizontal sum (vaddvq_u16). This replaces the 3 sequential pairwise additions in the critical path, improving performance on AArch64.


orestispol:
While the new code has fewer instructions, the vaddvq_u16() is slower due to high latency.
Closing this PR.

Issue #3249, opened 2026-08-06
Store pdf documentation documents in markdown


bkmgit:
[Slides](https://github.com/google/highway/blob/master/g3doc/highway_intro.pdf) and [instruction template](https://github.com/google/highway/blob/master/g3doc/instruction_matrix.pdf) are only stored as pdf documents, for easier maintenance and updates it would be helpful to have them stored as text files in a markup language.

For the slides one can use tools such as:
- https://github.com/Om3rr/markdown-to-googleslides 
- https://github.com/adamzap/landslide
- https://remarkjs.com
- https://pandoc.org/

to convert to other formats.  HTML slides would be nice to have at https://google.github.io/highway/en/master/presentation/index.html

One could possibly also use the webassembly version of highway to allow one to run code in the browser though may need to test how well [emcc](https://emscripten.org/docs/tools_reference/emcc.html) can build highway.

The instruction matrix can be stored as a markdown table and converted to pdf and/or html with highlighting as markdown tables do not display well as plain text without careful formatting.

jan-wassenberg:
Agreed on the instruction template, that would be nice.
Any particular reason to convert slides to markdown? I'd think it is fine to have a Google Slides link? (this would be an improvement vs the current PDF).

bkmgit:
Slide maintenance is easier if they can be updated with pull requests as developers mostly will use Git. One can also make interactive demos, see for example https://d28b.github.io/html-slides/demo/  thus allowing people to see how SIMD can make a difference in WASM and exprience the highway difference before building and compiling it.  See for example:
https://abseil.io/resources/swe-book/html/ch10.html#documentation_philosophy

PR #3269, opened 2026-08-11
Convert instruction matrix from pdf to md

bkmgit:
Partially addresses #3249

The pdf document uses color to highlight operations that require many cycles.  This can be done using css, but maybe easier if restructured text is used throughout, see https://stackoverflow.com/questions/33493589/add-a-class-to-a-table-in-sphinx#33497002 
  
Can check if moving from the [custom md to rst conversion script](https://github.com/google/highway/blob/master/docs/mm-converter.py) to [Myst](https://www.sphinx-doc.org/en/master/usage/markdown.html) can also help with this.

If there is willingness to move away from Sphinx,  can examine other static site generators such as [nift](https://nift.dev) and [sblg](https://kristaps.bsd.lv/sblg/)


bkmgit:
Unsure why getting:
```
MANDATORY_CHECKS: ["bot-conditions","insecure-commands","known-vulnerable-actions","template-injection","unpinned-uses","unredacted-secrets","unsound-condition"]
```

g3doc/instruction_matrix.md
```diff
@@ -0,0 +1,342 @@
+# Instruction matrix
+
+## x86 SSE4, AVX2
+
+| Operations ⬇       Types ⮕                   | i8 | u8 | i16 | u16 | i32 | u32 | i64 | u64 | f32 | f64 |
+|----------------------------------------------|----|----|-----|-----|-----|-----|-----|-----|-----|-----|
+|**ARITHMATIC**                                |    |    |     |     |     |     |     |     |     |     |
```
jan-wassenberg: Arithmetic (+multiple instances below)

jan-wassenberg (changes requested):
Thanks for converting :) One nit:

g3doc/instruction_matrix.md
bkmgit: Done.

jan-wassenberg (approved):

PR #3273, opened 2026-08-11
contrib/math: preserve negative zero in Expm1

EylonKrause:
### Summary

`Expm1(-0.0)` returns `+0.0` instead of `-0.0`, losing the sign of zero. C11 Annex F (F.10.3.3) requires `expm1(±0) = ±0`, and scalar `std::expm1` complies — so this is a signed-zero correctness bug (observable via `std::signbit`, or `1.0 / expm1(-0.0)` giving `+inf` instead of `-inf`).

### Cause

On the `|x| < ln2/2` polynomial path, `ExpPoly` computes `MulAdd(poly, x*x, x)`. For `x = -0.0` this is `0.5 * (+0.0) + (-0.0)`, and IEEE addition of oppositely-signed zeros yields `+0.0`, so the sign bit is dropped. Unlike the odd functions in this file (`Sin`, `Tan`, `Sinh`, …), `Expm1` never reapplies the sign of `x`, so the negative zero is not restored.

### Fix

Reapply `x`'s sign on the polynomial path using the sign-bit mask already defined in the function:

```cpp
const V z = IfThenElse(Lt(Abs(x), kLn2Over2), Or(y, And(x, kNegZero)),
                       Sub(impl.LoadExpShortRange(d, Add(y, kOne), q), kOne));
```

`expm1(x)` shares the sign of `x` for all `x`, so OR-ing in `x`'s sign bit is a no-op for every input except `x == -0.0`, which it corrects to `-0.0`. `+0.0`, small nonzero `x`, the large-`|x|` branch, and the `-1` clamp are unchanged, on every SIMD target.

### Testing

- Added `TestExpm1SignedZero`, a **bit-exact** check that `Expm1(±0)` returns `±0` — the existing ULP comparator treats `+0 == -0` and cannot detect this (which is why it went unnoticed).
- Before the fix the new test fails (`Expm1(-0)` yields `+0`); after, it passes.
- Full `math_test` passes (104 tests) across every enabled SIMD target.

### DIFF
--- a/hwy/contrib/math/math-inl.h
+++ b/hwy/contrib/math/math-inl.h
@@ -2660,7 +2660,10 @@ HWY_INLINE V Expm1(const D d, V x) {
 
   // Reduce, approximate, and then reconstruct.
   const V y = impl.ExpPoly(d, impl.ExpReduce(d, x, q));
-  const V z = IfThenElse(Lt(Abs(x), kLn2Over2), y,
+  // Reapply the sign of x on the polynomial path so a negative-zero input is
+  // preserved: expm1(-0) = -0 (C11 F.10.3.3), which the polynomial drops when
+  // it forms 0.5 * (+0) + (-0) = +0.
+  const V z = IfThenElse(Lt(Abs(x), kLn2Over2), CopySign(y, x),
                          Sub(impl.LoadExpShortRange(d, Add(y, kOne), q), kOne));
   return IfThenElse(Lt(x, kLowerBound), kNegOne, z);
 }
--- a/hwy/contrib/math/math_test.cc
+++ b/hwy/contrib/math/math_test.cc
@@ -191,6 +191,25 @@ HWY_NOINLINE void TestAllPow() {
   ForFloat3264Types(ForPartialVectors<TestPow>());
 }
 
+// expm1(+/-0) must preserve the sign of zero (C11 F.10.3.3). The ULP comparator
+// used by the other math tests treats +0 == -0, so check the bit pattern.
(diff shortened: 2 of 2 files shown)


hwy/contrib/math/math-inl.h
```diff
@@ -2660,7 +2660,10 @@ HWY_INLINE V Expm1(const D d, V x) {
 
   // Reduce, approximate, and then reconstruct.
   const V y = impl.ExpPoly(d, impl.ExpReduce(d, x, q));
-  const V z = IfThenElse(Lt(Abs(x), kLn2Over2), y,
+  // Reapply the sign of x on the polynomial path so a negative-zero input is
+  // preserved: expm1(-0) = -0 (C11 F.10.3.3), which the polynomial drops when
+  // it forms 0.5 * (+0) + (-0) = +0.
+  const V z = IfThenElse(Lt(Abs(x), kLn2Over2), Or(y, And(x, kNegZero)),
```
jan-wassenberg: Thanks again for fixing! Rather than Or+And, please use CopySign(y, x) - this is more efficient on some platforms.

jan-wassenberg: Once again thanks for fixing! Rather than Or+And, please use CopySign(y, x) - this is more efficient on some platforms.

jan-wassenberg (changes requested):

EylonKrause:
Done — switched to `CopySign(y, x)` in 1a4eb392. Thanks for the review!

jan-wassenberg (approved):

PR #3276, opened 2026-08-12
Use native LSX instructions

RaviTriv:
Updating LSX to use native instructions instead of generic fallbacks, matching LASX.


jan-wassenberg (approved):
Nice, thanks!

PR #3277, opened 2026-08-12
Faster NEON CompressStore: use fused BitsFromMask and PopCount

a-eee:
I added `detail::BitsAndCountFromMask`. It returns bits = BitsFromMask and count of ones in the mask. This function calculates count inside the existing algorithm for bits, so it spends only one operation for count instead of a call to PopCount.

It gives a win of about 30% (apple m5) here:
```
size_t loop(const uint32_t* src, uint32_t* dst, const size_t n) {
  ScalableTag<uint32_t> d;
  const size_t W = Lanes(d);
  size_t j = 0;
  for (size_t i = 0; i + W <= n; i += W) {
    auto v = Load(d, src + i);
    auto m = Gt(v, Set(d, 0x80000000u));
    j += CompressStore(v, m, d, dst + j);
  }
  return j;
}
```
Density is 50%.

### DIFF
--- a/hwy/ops/arm_neon-inl.h
+++ b/hwy/ops/arm_neon-inl.h
@@ -10260,13 +10260,52 @@ HWY_INLINE Vec128<T, N> CompressBits(Vec128<T, N> v,
   return detail::Compress(v, mask_bits);
 }
 
+namespace detail {
+struct BitsAndCount {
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Nice idea to fuse count into the table, thanks!

PR #3286, opened 2026-08-15
faster vqsort when <4 keys per vector: add new Sort16Rows cases to BaseCase

a-eee:
I added sorting networks for cases, when <4 keys fits the register. BaseCase::funcs:

1) <= 16: if 1 key, use Sort16Rows<1>
2) <= 32: if 2 keys, use Sort16Rows<2>

Also, kLPK in BaseCaseNumLanes is no longer needed, because cases for <4 added. I removed it.

bench_sort: for n <= 128 a single bench is terrible noisy, so now 30 sorts of 30 different buffers are measured for them in one measurement.

Final speedup(I alternated many benches and took the median, apple m5)
| type | n | speedup |
|-|-|-:|
| U128 | 16 | +326.3% |
| k+v=128 | 16 | +270.9% |
| i64 | 32 | +355.8% |
| U128 |  32 | +58.6% |
| k+v=128 | 32 | +56.5% |
| i64 | 64 | +71.7% |
| U128 |  64 | +53.0% |
| k+v=128 | 64 | +47.1% |
| i64 | 128 | +59.2% |
| U128 |  128 | +43.8%  |
| k+v=128 | 128 |  +39.9% |
| i64 | 10000 | +26.8% |
| U128 |  10000 | +23.6%  |
| k+v=128 | 10000 |  +22.6% |

for n=1e4 i used uniform32.
Other types remain unchanged

### DIFF
--- a/hwy/contrib/sort/bench_sort.cc
+++ b/hwy/contrib/sort/bench_sort.cc
@@ -17,6 +17,7 @@
 #include <stdio.h>
 
 #include <vector>
+#include <array>
 
 // clang-format off
 #undef HWY_TARGET_INCLUDE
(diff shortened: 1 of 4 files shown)


jan-wassenberg (approved):
Wow, this is impressive - great speedup, and simplifies the code :D Thanks for adding!

PR #3294, opened 2026-08-18
contrib/math: preserve negative zero in Sin

EylonKrause:
### Summary

`Sin(-0.0)` returns `+0.0` instead of `-0.0` on all targets. C11 Annex F (F.10.1.6) requires `sin(±0) = ±0`, and `std::sin` complies, so this is a signed-zero correctness bug.

### Cause

`Sin` folds the sign of `x` into the polynomial argument:

```cpp
return impl.Poly(d, Xor(impl.SinReduce(d, abs_x, q),
                        Xor(impl.SinSignFromQuadrant(d, q), sign_x)));
```

For `x = -0.0` the reduced argument is zero and the polynomial produces `+0.0`, so the negative-zero sign is lost.

### Fix

`sin` is odd, so its result must carry the sign of `x` when the magnitude is zero. Reapply `x`'s sign when the result is exactly zero:

```cpp
const V s = impl.Poly(...);
return IfThenElse(Eq(s, Zero(d)), CopySign(s, x), s);
```

For the supported range the result is exactly zero only at `x = ±0`, so the guard fires only there; every other input is unchanged on every target.

### Testing

- Added `TestSignedZero`, a **bit-exact** check that the sign-preserving math functions return `±0` for `±0`: `Sin`, `Tan`, `Asin`, `Atan`, `Asinh`, `Atanh`, `Cbrt`, `Erf`, `Log1p`, `Sinh`, `Tanh`. The existing ULP comparators treat `+0 == -0` and could not catch this. Before the fix the `Sin` case fails on all targets; after, all pass (the others were already correct).
- Full `math_test` passes (104 tests) across every enabled SIMD target.

### DIFF
--- a/hwy/contrib/math/math-inl.h
+++ b/hwy/contrib/math/math-inl.h
@@ -2860,8 +2860,11 @@ HWY_INLINE V Sin(const D d, V x) {
   const VI32 q = impl.ToInt32(d, MulAdd(abs_x, kOneOverPi, kHalf));
 
   // Reduce range, apply sign, and approximate.
-  return impl.Poly(d, Xor(impl.SinReduce(d, abs_x, q),
-                          Xor(impl.SinSignFromQuadrant(d, q), sign_x)));
+  const V s = impl.Poly(d, Xor(impl.SinReduce(d, abs_x, q),
+                               Xor(impl.SinSignFromQuadrant(d, q), sign_x)));
+  // sin is odd, so sin(-0) must be -0; the polynomial yields +0 for a zero
+  // reduced argument, so reapply x's sign when the result is zero.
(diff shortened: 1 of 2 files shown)


jan-wassenberg (dismissed):
Nice, thanks :)

EylonKrause:
Rebased onto master to resolve the merge conflict with #3273 (both PRs added a signed-zero test to `math_test.cc`). The two tests now sit side by side — `TestExpm1SignedZero` (from #3273) and `TestSignedZero` (this PR, covering Sin/Tan/Asin/Atan/Asinh/Atanh/Cbrt/Erf/Log1p/Sinh/Tanh). No other changes; `math_test` passes across all targets.

jan-wassenberg (approved):

PR #3296, opened 2026-08-19
Add float16 trig math functions, promote per half by default

dcsid:
Towards #2682, second PR of the series: adds `float16_t` overloads of `Sin`, `Cos`, `Tan` and `SinCos`. `SinCos` gets a two-output adapter; `Tan` calls the float32 `Tan` so the division happens in float32 rather than on demoted float16 sine/cosine.

As discussed in #3228, the adapter now promotes per half by default via `RepartitionToWide`, which keeps kPow2 and so always has exactly `Lanes(d) / 2` lanes; the MaxLanes-based predicate is gone. A single `Rebind` promotion remains only where per-half is impossible: single-lane tags, and fractional tags, whose `Rebind` is lane-exact and which cannot `Repartition` at kPow2 == -3 (same split as `IotaForSpecial`). Neither condition compares MaxLanes against a vector size, so both hold at every vector length; `HWY_DASSERT`s check the lane invariants.

The trig range is [-39000, +39000], mirroring the float32 kernels: beyond that the Cody-Waite reduction loses exactness without FMA (Cos reaches 8 ULP at x=54304). Over that range the exhaustive test measures 1 ULP for all four.

### DIFF
--- a/hwy/contrib/math/math-inl.h
+++ b/hwy/contrib/math/math-inl.h
@@ -165,7 +165,7 @@ HWY_NOINLINE V CallCbrt(const D d, VecArg<V> x) {
  *      Valid Range: [-39000, +39000]
  * @return cosine of 'x'
  */
-template <class D, class V>
+template <class D, class V, HWY_IF_NOT_SPECIAL_FLOAT_D(D)>
 HWY_INLINE V Cos(D d, V x);
 template <class D, class V>
 HWY_NOINLINE V CallCos(const D d, VecArg<V> x) {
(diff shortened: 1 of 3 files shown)


hwy/contrib/math/math-inl.h
```diff
@@ -2549,7 +2549,9 @@ HWY_INLINE V Cbrt(const D d, V x) {
   return y;
 }
 
-template <class D, class V>
+template <class D, class V,
+          hwy::EnableIf<
```
jan-wassenberg, line 2553: Can't we also use HWY_IF_NOT_SPECIAL_FLOAT_D(D) here and below?

jan-wassenberg (commented):
Nice work :) Just one question/nit:

hwy/contrib/math/math-inl.h
dcsid: I tried that, but the macro carries `= nullptr`, so on the out-of-line definition it redefines the default argument:

    math-inl.h:2552:29: error: template parameter redefines default argument
    math-inl.h:168:29: note: previous default template argument defined here

Unless I'm missing a trick, these four have to match the seven exp/log ones in master. Happy to do the declare-then-define cleanup you mentioned on #3228 as a separate PR.

jan-wassenberg: Ah, got it - the non-f16 Cos is blocking this because it's declared.
Thank you, would appreciate if you make that update later, and we can land this already.

jan-wassenberg (approved):

PR #3298, opened 2026-08-19
faster slide ops on NEON

a-eee:
I added to arm_neon-inl.h:
- SlideUpLanes and SlideDownLanes for 128-bit vectors: via table lookup, this saves 1 instruction for up ([godbolt](https://godbolt.org/z/GM7rTMErn)) and 4 for down ([godbolt](https://godbolt.org/z/xMvfxM1Po))
- Native SlideUpLanesOr(11 -> 5 instructions) and SlideDownLanesOr(16 -> 5 instructions):
   1) instead of the generic way with IfThenElse, FirstN and SlideUpLanes, now using table lookup, which costs the same as SlideUpLanes (but tbl2 instead of tbl1).
   2) For __builtin_constant_p now using the same way as before
   3) For 1 lane registers now using IfThenElse(FirstN, lo, hi), 3 instructions.

### DIFF
--- a/hwy/ops/arm_neon-inl.h
+++ b/hwy/ops/arm_neon-inl.h
@@ -6811,6 +6811,17 @@ HWY_INLINE V Per4LaneBlockShuffle(hwy::SizeTag<0xFA> /*idx_3210_tag*/,
 // ------------------------------ SlideUpLanes
 
 namespace detail {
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Nice work, thanks! Makes sense to load the indices directly, and good idea to use TwoTables.

PR #3299, opened 2026-08-20
Use constexpr for VQSORT_PRINT conditionals

wmamrak:
Fixes `Unresolved External Symbol` (LNK2019) linker error in MSVC when compiled with HWY_HEADER_ONLY set to 1, which arises due to missing definition of ToString function called by PrintValue function in ChoosePivotByRank function. Obviously, it just hides the real issue, as increasing verbosity (VQSORT_PRINT != 0) will reveal it again, but is simple and makes the default case (VQSORT_PRINT 0) work.


jan-wassenberg (approved):
Thanks! Note that (only) contrib/ allows C++17, hence this is fine.

wmamrak:
The readme states `Highway only requires C++17` (i.e. without any asterisks :-))

PR #3310, opened 2026-08-21
NEON: add native 8bit compress

a-eee:
Every func was rewritten using tbl without a store to the stack. Operations on 16lane Vec are performed through 2 8lane operations, because otherwise the tables would be too large. Also a table is used for the 8bit popcount.

godbolt:
| Func | total diff | per iteration diff |
|-|-|-|
[Compress](https://godbolt.org/z/Y44jsd18e) |-2|-5|
[CompressNot](https://godbolt.org/z/ajoTz5cYo) | -2 | -5 |
[CompressBits](https://godbolt.org/z/sa7eeb7Eh) | -2 | -5 |
[CompressStore](https://godbolt.org/z/MorW4eqoE) | 0 | 0|
[CompressBlendedStore](https://godbolt.org/z/dKbjWM1KE) | -3 | -5 |
[CompressBitsStore](https://godbolt.org/z/da7YoYa7h) | -2 | -4|

per iteration = count of the instructions that depend on the data. The compiler cannot hoist them out of the loop.

### DIFF
--- a/hwy/ops/arm_neon-inl.h
+++ b/hwy/ops/arm_neon-inl.h
@@ -9947,7 +9947,11 @@ HWY_API bool AllTrue(D d, MFromD<D> m) {
 }
 
 // ------------------------------ Compress
-
+#ifdef HWY_NATIVE_COMPRESS8
+#undef HWY_NATIVE_COMPRESS8
+#else
+#define HWY_NATIVE_COMPRESS8
+#endif
 template <typename T>
 struct CompressIsPartition {
   enum { value = (sizeof(T) != 1) };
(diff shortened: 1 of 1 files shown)


hwy/ops/arm_neon-inl.h
```diff
@@ -10380,6 +10682,34 @@ HWY_INLINE Vec128<T, N> Compress(Vec128<T, N> v, uint64_t mask_bits) {
   return BitCast(D(), TableLookupBytes(BitCast(di, v), BitCast(di, idx)));
 }
 
+template <typename T, size_t N, HWY_IF_T_SIZE(T, 1), HWY_IF_LANES(N, 16)>
+HWY_INLINE Vec128<T, N> Compress(Vec128<T, N> v, uint64_t lo, uint64_t hi) {
+  using D = DFromV<decltype(v)>;
+
+  alignas(16) static constexpr uint8_t SlideTable[144] = {
```
jan-wassenberg: A naming nit: please use **k**SlideTable for constant data. Maybe also replace 144 with 8*16 + 16?

hwy/ops/arm_neon-inl.h
```diff
@@ -10504,19 +10909,42 @@ HWY_INLINE auto BitsAndCountFromMask(D, MFromD<D> m) {
   const TU s = ReduceSum(du, And(VecFromMask(du, RebindMask(du, m)), Load(du, weights)));
   return BitsAndCount{static_cast<uint64_t>(s & 255), static_cast<size_t>(s >> 8)};
 }
+template <typename D, HWY_IF_T_SIZE_D(D, 1), HWY_IF_LANES_D(D, 8)>
+HWY_INLINE auto BitsAndCountFromMask(D, MFromD<D> m) {
+  RebindToUnsigned<D> du;
+  using TU = TFromD<decltype(du)>;
+  alignas(16) static constexpr TU weights[8] = {1, 2, 4, 8, 16, 32, 64, 128};
```
jan-wassenberg: Also kWeights, please :)

jan-wassenberg (changes requested):
Nice, thank you for optimizing :) Two minor nits:

a-eee:
Done, thanks. also i renamed other weights in the other BitsAndCountFromMask overloads for consistency.

jan-wassenberg (approved):
Thanks for updating!

a-eee:
I want to make some perf changes in contrib/algo, but there are no benchmarks. Is it OK if I add algo_bench.cc?

PR #3311, opened 2026-08-23
Speed up CompactBTree leaf scans

Magnushst:
CompactBTree leaf scans currently call `CountTrue` for every SIMD vector. This repeatedly transfers comparison masks to scalar code and serialises the accumulated count.

Accumulate the comparison masks in four independent SIMD counters instead, then perform one horizontal reduction after the scan. Four counters also avoid introducing a single vector dependency chain. For 8-bit offsets, `SumsOf8` widens the partial sums before `ReduceSum`, preventing narrow-lane overflow.

The lower-bound, upper-bound, and masked-tail semantics are unchanged. This does not change the public API.

In a local AVX2 benchmark containing one million `uint32_t` keys and two million random `lower_bound` queries per trial, a low-load 12-trial run improved median throughput from 14.31 to 23.72 million queries per second (1.66x). A separate 36-trial run measured a 1.70x improvement. Absolute throughput varied between sessions, but the relative improvement remained consistent. The benchmark binary's `popcnt` instruction count decreased from 66 to 2.

Validation performed:

- Compared `CompactBTreeSet` and `CompactBTreeMap` results against scalar standard-library operations.
- Covered 8-, 16-, 32-, and 64-bit leaf storage modes.
- Checked 20,000 random `lower_bound`, `upper_bound`, and `contains` queries per case.
- Passed GCC 14 AVX2 and SSE2 builds.
- Passed Clang 19 AddressSanitizer and UndefinedBehaviorSanitizer builds.
- Passed direct template-instantiation builds with conversion and sign-conversion warnings treated as errors.
- Passed formatting and `git diff --check`.
- Verified equivalence to the previous scan implementation for 20,000 random and boundary-value inputs per offset width.

### DIFF
--- a/hwy/contrib/btree/btree-inl.h
+++ b/hwy/contrib/btree/btree-inl.h
@@ -411,29 +411,80 @@ HWY_INLINE size_t ScanOffsets(const void* HWY_RESTRICT data,
   const hn::ScalableTag<OffsetT> d;
   const size_t N = hn::Lanes(d);
   const auto v_target = hn::Set(d, target_val);
-  size_t count = 0;
-  size_t i = 0;
-  for (; i + N <= kTotal; i += N) {
-    const auto v = hn::Load(d, offsets + i);
-    if constexpr (kBound == BoundMode::kLowerBound) {
-      count += hn::CountTrue(d, hn::Lt(v, v_target));
+  static_assert(kTotal <= 512 / sizeof(OffsetT));
+
+  if constexpr (HWY_NATIVE_MASK) {
+    // CountTrue is inexpensive with native mask registers and avoids a
+    // horizontal vector reduction.
+    size_t count = 0;
+    size_t i = 0;
+    for (; i + N <= kTotal; i += N) {
+      const auto v = hn::Load(d, offsets + i);
+      if constexpr (kBound == BoundMode::kLowerBound) {
+        count += hn::CountTrue(d, hn::Lt(v, v_target));
+      } else {
+        count += hn::CountTrue(d, hn::Le(v, v_target));
+      }
+    }
+
+    if (i < kTotal) {
+      const size_t remaining = kTotal - i;
+      const auto v = hn::LoadU(d, offsets + i);
+      const auto mask = hn::FirstN(d, remaining);
+      if constexpr (kBound == BoundMode::kLowerBound) {
+        count += hn::CountTrue(d, hn::MaskedLt(mask, v, v_target));
+      } else {
+        count += hn::CountTrue(d, hn::MaskedLe(mask, v, v_target));
+      }
+    }
+
+    return count;
+  } else {
+    const auto is_before = [&](const auto v) HWY_ATTR {
+      if constexpr (kBound == BoundMode::kLowerBound) {
+        return hn::Lt(v, v_target);
+      } else {
+        return hn::Le(v, v_target);
+      }
+    };
+
+    // Without native mask registers, accumulate comparison masks in vectors
+    // and reduce once to avoid repeated vector-to-scalar transfers. The loop
+    // is 2x unrolled to shorten the accumulator dependency chain.
+    auto counts0 = hn::Zero(d);
+    auto counts1 = hn::Zero(d);
+    size_t i = 0;
+    if (kTotal >= 2 * N) {
+      for (; i <= kTotal - 2 * N; i += 2 * N) {
+        counts0 = hn::Sub(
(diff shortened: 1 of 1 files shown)


hwy/contrib/btree/compact_btree-inl.h
```diff
@@ -415,29 +415,60 @@ HWY_INLINE size_t ScanCompactOffsets(const void* HWY_RESTRICT data,
...
+      return hn::Lt(v, v_target);
     } else {
-      count += hn::CountTrue(d, hn::Le(v, v_target));
+      return hn::Le(v, v_target);
+    }
+  };
+
+  // Four accumulators avoid both the scalar CountTrue bottleneck and a vector
+  // dependency chain. Each lane is incremented at most 64 times because
+  // Highway vectors are at least eight bytes and a leaf is at most 512 bytes.
```
Nikhil0250: afaiu this comments sounds a bit misleading : 
4 accumulators have no role in avoiding the CountTrue bottleneck they just help us unroll the loop since the computations are independent of each other. you can just write 4x unrolled in the comment and that would suffice
"Highway vectors are at least eight bytes" -> highway vectors are actually atleast 16bytes, since we dont allow HWY_SCALAR targets in btree

hwy/contrib/btree/compact_btree-inl.h
```diff
@@ -415,29 +415,60 @@ HWY_INLINE size_t ScanCompactOffsets(const void* HWY_RESTRICT data,
...
+  auto counts0 = hn::Zero(d);
+  auto counts1 = hn::Zero(d);
+  auto counts2 = hn::Zero(d);
+  auto counts3 = hn::Zero(d);
+  size_t i = 0;
+  if (kTotal >= 4 * N) {
+    for (; i <= kTotal - 4 * N; i += 4 * N) {
```
Nikhil0250: I ran some benchmarks and it seems 2x unrolling is actually a bit faster than 4x unrolling here

Nikhil0250:
Thanks for raising the PR ! Nice approach ! 
I ran some benchmarks and this approach works well specifically for non native mask register architectures, this is because as you pointed out we have transfer the comparison results from vector registers to the scalar registers which is expensive but for AVX 512 (or any other archs having native masks), the current approach is actually quite a bit faster. 
This is because we have dedicated mask registers here and moving stuff from mask registers to integer registers is very fast and we avoid the horizontal sum at the end.
So to summarize for non-native mask archs, your approach does give a nice speedup 1.3-1.6x for range queries but for native mask archs the current approach is better.

We can use both conditional on HWY_NATIVE_MASK. you can add a brief comment explaining the tradeoff there..
I have added some comments on the PR, can you kindly review and make these changes ? (you will have to rebase/sync since we submitted a bunch of cl's and moved compactbtree implementation into btree-inl.h)

hwy/contrib/btree/compact_btree-inl.h
```diff
@@ -415,29 +415,60 @@ HWY_INLINE size_t ScanCompactOffsets(const void* HWY_RESTRICT data,
   const hn::ScalableTag<OffsetT> d;
   const size_t N = hn::Lanes(d);
   const auto v_target = hn::Set(d, target_val);
-  size_t count = 0;
-  size_t i = 0;
-  for (; i + N <= kTotal; i += N) {
-    const auto v = hn::Load(d, offsets + i);
+  static_assert(kTotal <= 512 / sizeof(OffsetT));
+  static_assert(hwy::LimitsMax<OffsetT>() >= 512 / HWY_MIN_BYTES);
```
Nikhil0250: we can skip this assert, since HWY_MIN_BYTES is atleast 16, 512/16 is 32, and even for 8-bits we have the limit as 255 so we are pretty far off

Magnushst: Removed the counter-width assertion. I verified that the final uint8_t lane count is at most 31; the separate 512-byte storage-bound assertion is retained.

hwy/contrib/btree/compact_btree-inl.h
Magnushst: Updated. Vector accumulation avoids the repeated CountTrue transfers, while 2x unrolling only shortens the accumulator dependency chain. The incorrect eight-byte statement is removed.

Nikhil0250 (approved):
LGTM !

Nikhil0250:
I think @jan-wassenberg will have to approve to merge..

jan-wassenberg (approved):
👍

Issue #2622, opened 2025-07-13
10/804 tests fail on armv7 with gcc-14


stefson:
seems my comment from 4 hours ago was meant to be short lived: 

```
The following tests FAILED:
         85 - PerfCountersTest.TestMem (Subprocess aborted)
        206 - HwyConvertTestGroup/HwyConvertTest.TestAllPromoteOddEvenTo/EMU128  # GetParam() = 2305843009213693952 (Subprocess aborted)
Errors while running CTest
```

here is the verbose test log file: [LastTest.log.gz](https://github.com/user-attachments/files/21205530/LastTest.log.gz)

apologies for being overly confident, I never bothered to run tests nativly in the last year or so, and was only testing latest git with a crosscompiler.

jan-wassenberg:
Thanks! The first is trivial and I've sent out a fix. Surprising that platforms differ by 3 orders of magnitude, but OK :)

The second one is probably a compiler bug, sigh. I'm currently on the road but can have a look next week.

stefson:
thank you for the fast fix, test 85 now passes. 

sadly, there are now 9 additional failing tests: 

```
        206 - HwyConvertTestGroup/HwyConvertTest.TestAllPromoteOddEvenTo/EMU128  # GetParam() = 2305843009213693952 (Subprocess aborted)
        506 - HwyMulPairwiseTestGroup/HwyMulPairwiseTest.TestAllSatWidenMulPairwiseAdd/EMU128  # GetParam() = 2305843009213693952 (Subprocess aborted)
        526 - HwyReductionTestGroup/HwyReductionTest.TestAllSumsOf2/EMU128  # GetParam() = 2305843009213693952 (Subprocess aborted)
        528 - HwyReductionTestGroup/HwyReductionTest.TestAllSumsOf4/EMU128  # GetParam() = 2305843009213693952 (Subprocess aborted)
        574 - HwyShiftTestGroup/HwyShiftTest.TestAllVariableRoundingShr/EMU128  # GetParam() = 2305843009213693952 (Subprocess aborted)
        766 - SortTestGroup/SortTest.TestAllSortIota/EMU128  # GetParam() = 2305843009213693952 (Subprocess aborted)
        768 - SortTestGroup/SortTest.TestAllSort/EMU128  # GetParam() = 2305843009213693952 (Subprocess aborted)
        770 - SortTestGroup/SortTest.TestAllSelect/EMU128  # GetParam() = 2305843009213693952 (Subprocess aborted)
[...]

Issue #1683, opened 2023-08-28
armhf: HwyMulTestGroup/HwyMulTest.TestAllSatWidenMulPairwiseAdd/EMU128


malaterre:
Version c7b1348 currently fails on armhf/gcc-13:

Total Test time (real) =  24.78 sec

The following tests FAILED:
	350 - HwyMulTestGroup/HwyMulTest.TestAllSatWidenMulPairwiseAdd/EMU128  # GetParam() = 2305843009213693952 (Subprocess aborted)
	356 - HwyMulTestGroup/HwyMulTest.TestAllSumOfMulQuadAccumulate/EMU128  # GetParam() = 2305843009213693952 (Subprocess aborted)
Errors while running CTest

ref:
* https://buildd.debian.org/status/fetch.php?pkg=highway&arch=armhf&ver=1.0.7%7Egit20230825.c7b1348-1&stamp=1693207147&raw=0

jan-wassenberg:
Thanks for reporting this. GCC-13 is unfortunately quite broken :/ Looks like this is 13.2.
In detect_targets, how about we change this
`HWY_COMPILER_GCC_ACTUAL && HWY_COMPILER_GCC_ACTUAL < 1203`
to 1303? Or perhaps less optimistically 1400?

jan-wassenberg:
Sounds good, thanks for filing a bug :)

jan-wassenberg:
Thanks for filing the bug :)

PR #3317, opened 2026-08-25
Do not extend the EMU128 opt-out to GCC 14/15 on RISC-V

velonica0:
## Problem

On RISC-V without the V extension there is no native SIMD target, so the fallback target is also the static target. HWY_BROKEN_EMU128 therefore selects HWY_SCALAR (HWY_LANES(T) == 1), and code using fixed-width tags fails to compile rather than running slower.

V8's JSON stringifier builds a FixedTag<uint8_t, 16> unconditionally, so Node.js cannot currently be built for riscv64 with GCC 15 — see [conda-forge/nodejs-feedstock#498](https://github.com/conda-forge/nodejs-feedstock/pull/498) for the downstream build failure.

The threshold was raised to GCC < 14 in 952590cc (#1683) and to GCC < 16 in f38ee257 (#2622). Both reports are Armv7, and both trace to https://gcc.gnu.org/PR111231.

## Change

Keep the blanket GCC < 14 opt-out; don't extend it to GCC 14/15 on RISC-V:

#if (HWY_COMPILER_GCC_ACTUAL &&                                   \
     HWY_COMPILER_GCC_ACTUAL < (HWY_ARCH_RISCV ? 1400 : 1600)) || \
    defined(HWY_NO_LIBCXX)

Only GCC 14/15 on RISC-V change; Armv7, x86, clang and HWY_NO_LIBCXX are unaffected.

## Testing

Native RISC-V hardware (SpacemiT, Bianbu 4.0), g++ 15.2.0, -march=rv64imafdc -mabi=lp64d:

before:  static target = HWY_SCALAR  HWY_LANES(uint8_t)=1
after:   static target = HWY_EMU128  HWY_LANES(uint8_t)=16

All 62 test TUs pass with EMU128 (pass=62 fail=0 buildfail=0), including the tests that fail on Armv7 in #2622 — TestAllPromoteOddEvenTo, TestAllSatWidenMulPairwiseAdd, TestAllSumsOf2, TestAllSumsOf4. The bug does not reproduce on RISC-V.

### DIFF
--- a/hwy/detect_targets.h
+++ b/hwy/detect_targets.h
@@ -409,8 +409,17 @@
 // remains with 13.2, see #1683. This is separate from HWY_BROKEN_TARGETS
 // because it affects the fallback target, which must always be enabled. If 1,
 // we instead choose HWY_SCALAR even without HWY_COMPILE_ONLY_SCALAR being set.
+// The extension of this opt-out to GCC 14 and 15 was prompted by Armv7
+// failures (#1683, #2622), both of which trace to
+// https://gcc.gnu.org/PR111231. RISC-V has no native SIMD baseline unless the
+// V extension is enabled, so there EMU128 is the only fallback able to provide
+// fixed-width tags: substituting HWY_SCALAR does not merely lose speed, it
+// makes FixedTag<T, N> with N > 1 fail to compile, which breaks downstream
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Sounds reasonable, thanks for updating :)

PR #3329, opened 2026-08-30
Add ReduceMin/MaxOrNaN and document reduction NaN behavior

RaviTriv:
Document `ReduceMin/Max` NaN behaviour and add `ReduceMin/MaxOrNan` which returns NaN if any lane is NaN.

### DIFF
--- a/hwy/ops/generic_ops-inl.h
+++ b/hwy/ops/generic_ops-inl.h
@@ -1322,6 +1322,27 @@ HWY_API TFromD<D> MaskedReduceMax(D d, M m, VFromD<D> v) {
 
 #endif  // HWY_NATIVE_MASKED_REDUCE_SCALAR
 
+// ------------------------------ ReduceMinOrNaN/ReduceMaxOrNaN
+
+// AArch64 Neon overrides this because vminv/vmaxv already propagate NaN.
+#if (defined(HWY_NATIVE_REDUCE_MINMAX_OR_NAN) == defined(HWY_TARGET_TOGGLE))
+#ifdef HWY_NATIVE_REDUCE_MINMAX_OR_NAN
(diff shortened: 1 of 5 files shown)


jan-wassenberg:
Sounds nice, but I only see the change to the wishlist - are the other files missing? :)

hwy/ops/generic_ops-inl.h
```diff
@@ -1322,6 +1322,27 @@ HWY_API TFromD<D> MaskedReduceMax(D d, M m, VFromD<D> v) {
...
+#if (defined(HWY_NATIVE_REDUCE_MINMAX_OR_NAN) == defined(HWY_TARGET_TOGGLE))
+#ifdef HWY_NATIVE_REDUCE_MINMAX_OR_NAN
+#undef HWY_NATIVE_REDUCE_MINMAX_OR_NAN
+#else
+#define HWY_NATIVE_REDUCE_MINMAX_OR_NAN
+#endif
+
+template <class D, HWY_IF_FLOAT_D(D)>
+HWY_API TFromD<D> ReduceMinOrNaN(D d, VFromD<D> v) {
+  return AllFalse(d, IsNaN(v)) ? ReduceMin(d, v) : GetLane(NaN(d));
```
jan-wassenberg, line 1337: Maybe we should define a ScalarNaN function to reduce the overhead?

jan-wassenberg (approved):
Interesting, Github seems to have had a little hiccup and was only displaying one of the commits in this PR.

hwy/ops/generic_ops-inl.h
RaviTriv: Good idea!

[Follow up](https://github.com/google/highway/pull/3338)

PR #3338, opened 2026-09-01
add ScalarNaN

RaviTriv:
Followup to https://github.com/google/highway/pull/3329. Adding `ScalarNan` that returns quiet NaN


jan-wassenberg (approved):
Thanks for adding :)

PR #3339, opened 2026-09-02
Add portable interleaved range coder to hwy/contrib/coder

tvost2:
Implements the **Range coder** item from `g3doc/op_wishlist.md`: a Highway
port of Richard Geldreich's public-domain
[`sserangecoding`](https://github.com/richgel999/sserangecoding) — a 24-bit
interleaved Pavlov/Subbotin range coder over an 8-bit alphabet with 16
interleaved streams. The SSE4.1 decode kernel is re-expressed with Highway
ops so a single source runs on every target.

### What's here

| file | contents |
|---|---|
| `hwy/contrib/coder/range_coder.h` / `.cc` | scalar `RangeEncoder` / `RangeDecoder`, the probability-model helpers (`CreateCumulativeProbs`, `BuildDecodeTable`), the scalar interleaved encoder (`EncodeInterleaved`, not vectorized in the original either) and a scalar reference decoder (`DecodeInterleavedScalar`) |
| `hwy/contrib/coder/range_coder-inl.h` | **SIMD `DecodeInterleaved`** over `Full128<uint32_t>` — the 16 streams as four groups of four lanes, held in `value0..value3` / `length0..length3` locals and driven from an unrolled loop, matching the reference's 128-bit kernel. Uses `GatherIndex` for the decode table, `TruncateStore` for the four symbol bytes, one 16-byte `LoadU` + `TableLookupBytes` per group for the big-endian 24-bit header, `TableLookupBytesOr0` for the renormalization shuffles, and float divide + truncating `ConvertTo` for the exact 24-bit `value / range` (per the note in the original crediting @jan-wassenberg for the `_mm_cvttps_epi32` suggestion). Runs on RVV and SVE (masked to four lanes) as well as the fixed-size targets; only `HWY_SCALAR`, which cannot form a 4-lane vector, falls back to the scalar decoder, which produces identical output. Widening the group past 128 bits would need the renorm shuffle tables reworked per block — a possible follow-up. |
| `hwy/contrib/coder/range_coder_test.cc` | round-trips a range of sizes (below/at/above one 16-lane block, plus a scalar-tail remainder; `AdjustedReps` on the large sizes) and models (skewed, uniform, one and two live symbols), cross-checking the SIMD and scalar decoders |

Wired into CMake, Bazel (`BUILD`, `hwy_tests.bzl`) and Meson; wishlist entry removed.

### Validation

- `range_coder_test` passes on AVX3/AVX2/SSE4/SSSE3/SSE2/EMU128 locally.
- Strict `-Werror` build with the Linux CI flag set (`-Wconversion -Wsign-conversion -Wcast-align -Wshadow -Wmissing-declarations -Wextra-semi -Wunreachable-code …`): clean. `clang-format` clean.
- **The encoder is byte-for-byte identical to `sserangecoding`.** Building the upstream encoder and comparing: `book1` → the same 436229-byte stream (56.7 %, within 0.27 % of the order-0 entropy), and both the SIMD and scalar decoders round-trip it.

The one deliberate difference from the reference: `DecodeInterleaved` never reads outside `[src, src + comp_size)` (the reference's `pSrc_end` is deliberately loose); it rejects `comp_size < 48` up front and the short scalar tail decodes from a zero-padded copy, so a truncated/corrupt stream cannot over-read.

### DIFF
--- a/hwy/contrib/coder/range_coder-inl.h
+++ b/hwy/contrib/coder/range_coder-inl.h
@@ -17,9 +17,14 @@
 // target, decoding the bitstream produced by hwy::EncodeInterleaved (see
 // range_coder.h). Ported from Richard Geldreich's public-domain
 // "sserangecoding" (https://github.com/richgel999/sserangecoding); the SSE4.1
-// kernel worked on four u32 lanes at a time, and this keeps that shape
-// (CappedTag<uint32_t, 4>), so the 16 interleaved streams are decoded as four
-// groups of four lanes.
+// kernel worked on four u32 lanes at a time, and this keeps that shape: the 16
+// interleaved streams are four groups of four lanes, held in four vector
+// locals and driven from an unrolled loop. The 4-lane vector is FixedTag, so
+// the same kernel runs on RVV and SVE (masked to four lanes) as well as the
+// fixed-size targets; only HWY_SCALAR, which cannot form a 4-lane vector, uses
+// the scalar reference decoder. Widening the group to Lanes(d) would need the
+// renormalization step (below) to be reworked away from its 4-lane pshufb
+// tables; that is a possible follow-up.
 //
 // The float divide for value/range is exact because `value` is always <= 24
 // bits and `range` <= 12 bits; ConvertTo truncates toward zero, as suggested by
@@ -46,12 +51,10 @@ namespace hwy {
 namespace HWY_NAMESPACE {
 namespace range_coder {
 
-// The kernel keeps four u32 lanes in local variables and drives them from a
-// small loop, which needs a fixed-size (non-sizeless) 4-lane vector. That rules
-// out HWY_SCALAR (1 lane) and the scalable targets (RVV, SVE), whose vector
-// types cannot be array elements. Those get the scalar reference decoder, which
-// produces identical output; a scalable kernel is a possible follow-up.
-#if HWY_TARGET == HWY_SCALAR || HWY_HAVE_SCALABLE || HWY_TARGET_IS_SVE
+// HWY_SCALAR has a single lane and cannot form the 4-lane vector the kernel
+// needs, so it decodes with the scalar reference implementation, which produces
+// identical output.
+#if HWY_TARGET == HWY_SCALAR
 
 HWY_INLINE bool DecodeInterleaved(const uint8_t* HWY_RESTRICT src,
                                   size_t comp_size, uint8_t* HWY_RESTRICT dst,
@@ -64,23 +67,13 @@ HWY_INLINE bool DecodeInterleaved(const uint8_t* HWY_RESTRICT src,
 
 namespace detail {
 
-HWY_INLINE uint32_t ReadBE24(const uint8_t*& src) {
-  const uint32_t res = (static_cast<uint32_t>(src[0]) << 16) |
-                       (static_cast<uint32_t>(src[1]) << 8) |
-                       static_cast<uint32_t>(src[2]);
-  src += 3;
-  return res;
-}
-
 // Decodes one symbol per lane (4 total), writing the 4 symbol bytes to `out4`.
 template <class D>
 HWY_INLINE void Decode(D d, VFromD<D>& value, VFromD<D>& length,
                        const uint32_t* HWY_RESTRICT table,
-                       VFromD<Repartition<uint8_t, D> > pack_idx,
                        uint8_t* HWY_RESTRICT out4) {
   const RebindToSigned<D> di;
   const RebindToFloat<D> df;
-  const Repartition<uint8_t, D> d8;
 
   const VFromD<D> r = ShiftRight<kRangeProbBits>(length);
   const VFromD<D> q =
@@ -90,10 +83,8 @@ HWY_INLINE void Decode(D d, VFromD<D>& value, VFromD<D>& length,
 
   const VFromD<D> e = GatherIndex(d, table, BitCast(di, q));
 
-  // Byte 0 of each u32 lane -> low 4 bytes.
-  HWY_ALIGN uint8_t sym_bytes[16];
-  Store(TableLookupBytesOr0(BitCast(d8, e), pack_idx), d8, sym_bytes);
-  memcpy(out4, sym_bytes, 4);
+  // Byte 0 of each u32 lane is the symbol; keep the low byte of each lane.
+  TruncateStore(e, d, out4);
 
   const VFromD<D> low_prob = And(ShiftRight<8>(e), Set(d, kRangeProbScale - 1));
   const VFromD<D> prob_range = ShiftRight<20>(e);  // 8 + kRangeProbBits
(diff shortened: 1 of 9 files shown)


hwy/contrib/coder/range_coder-inl.h
```diff
@@ -0,0 +1,213 @@
...
+namespace hwy {
+namespace HWY_NAMESPACE {
+namespace range_coder {
+
+// The kernel keeps four u32 lanes in local variables and drives them from a
+// small loop, which needs a fixed-size (non-sizeless) 4-lane vector. That rules
+// out HWY_SCALAR (1 lane) and the scalable targets (RVV, SVE), whose vector
+// types cannot be array elements. Those get the scalar reference decoder, which
+// produces identical output; a scalable kernel is a possible follow-up.
+#if HWY_TARGET == HWY_SCALAR || HWY_HAVE_SCALABLE || HWY_TARGET_IS_SVE
```
jan-wassenberg: Let's also enable scalable vectors here. The simplest path is just to manually unroll the loop with value0..value3 vectors. But I wonder: why not have CappedTag<uint32_t, 16>, then loop over the 16 states in steps of Lanes(d)? This is anyway in a loop, hence the CPU's OoO would also hide latency.

hwy/contrib/coder/range_coder-inl.h
```diff
@@ -0,0 +1,213 @@
...
+  const VFromD<D> q =
+      And(BitCast(d, ConvertTo(di, Div(ConvertTo(df, BitCast(di, value)),
+                                       ConvertTo(df, BitCast(di, r))))),
+          Set(d, kRangeProbScale - 1));
+
+  const VFromD<D> e = GatherIndex(d, table, BitCast(di, q));
+
+  // Byte 0 of each u32 lane -> low 4 bytes.
+  HWY_ALIGN uint8_t sym_bytes[16];
+  Store(TableLookupBytesOr0(BitCast(d8, e), pack_idx), d8, sym_bytes);
```
jan-wassenberg: I think this could also use u8 TruncateTo, which would be necessary if we want to allow larger than 128-bit vectors.

hwy/contrib/coder/range_coder-inl.h
```diff
@@ -0,0 +1,213 @@
...
+// Decodes data produced by hwy::EncodeInterleaved. `table` is from
+// hwy::BuildDecodeTable. Returns false if the input is truncated. Output is
+// identical to hwy::DecodeInterleavedScalar.
+HWY_INLINE bool DecodeInterleaved(const uint8_t* HWY_RESTRICT src_start,
+                                  size_t comp_size,
+                                  uint8_t* HWY_RESTRICT dst_start,
+                                  size_t orig_size,
+                                  const uint32_t* HWY_RESTRICT table) {
+  // CappedTag gives exactly 4 lanes on every fixed-size target.
+  const CappedTag<uint32_t, 4> d;
```
jan-wassenberg: Might as well use Full128<uint32_t> if we really want exactly 128 bits.

hwy/contrib/coder/range_coder-inl.h
```diff
@@ -0,0 +1,213 @@
...
+  const RangeShuffleTables& sh = GetRangeShuffleTables();
+
+  const uint8_t* src = src_start;
+  const uint8_t* const src_end = src_start + comp_size;
+
+  VFromD<decltype(d)> value[4];
+  VFromD<decltype(d)> length[4];
+  for (int g = 0; g < 4; ++g) {
+    HWY_ALIGN uint32_t v4[4];
+    for (int l = 0; l < 4; ++l) v4[l] = detail::ReadBE24(src);
```
jan-wassenberg: This could be a 16-byte load + TableLookupBytes.

hwy/contrib/coder/range_coder_test.cc
```diff
@@ -0,0 +1,140 @@
...
+
+HWY_BEFORE_NAMESPACE();
+namespace hwy {
+namespace HWY_NAMESPACE {  // required: unique per target
+namespace {
+
+namespace rc = hwy::HWY_NAMESPACE::range_coder;
+
+// Small deterministic xorshift PRNG (test-only).
+uint64_t NextRandom(uint64_t& state) {
```
jan-wassenberg: Can also use Random64 from test_util.h.

hwy/contrib/coder/range_coder_test.cc
```diff
@@ -0,0 +1,140 @@
...
+  HWY_ASSERT(memcmp(dec_scalar.data(), data.data(), data.size()) == 0);
+
+  if (expect_shrink) HWY_ASSERT(enc.size() < data.size());
+}
+
+// Round-trips a range of sizes: below one 16-lane block, exactly one, several
+// whole blocks, and whole blocks + a scalar-tail remainder.
+void TestRoundTripSizes() {
+  const size_t kSizes[] = {1,  2,   3,   15,   16,   17,   31,    32,   33,
+                           64, 127, 256, 1023, 4096, 4097, 65535, 65536};
```
jan-wassenberg: Here also (and below) AdjustedReps around the >1k sizes, please.

jan-wassenberg (changes requested):
Neat to see this implemented, thanks :)

hwy/contrib/coder/range_coder-inl.h
tvost2: Done. Went with the manual unroll: the four groups are now `value0..value3` / `length0..length3` locals over a `Full128<uint32_t>`, so the kernel compiles and runs on RVV and SVE (masked to 4 lanes) instead of being hard-gated to SSE4. Only `HWY_SCALAR` still falls back to the scalar reference decoder.

I kept the group at 4 lanes rather than `CappedTag<uint32_t, 16>` + `Lanes(d)` because `Normalize()` consumes up to 2 bytes/lane through 128-bit `TableLookupBytes` shuffle tables indexed by an 8-bit `(len<min | len<256)` mask; going wider needs that renorm reworked per 128-bit block. There's a comment marking it as a follow-up - happy to do it here instead if you'd prefer.

hwy/contrib/coder/range_coder-inl.h
tvost2: Done - the four symbol bytes are now written with `TruncateStore` (u32 -> u8) instead of the old `pack_idx` gather.

hwy/contrib/coder/range_coder-inl.h
tvost2: Done - `Full128<uint32_t>`.

hwy/contrib/coder/range_coder-inl.h
tvost2: Done - the 16 initial BE24 words are now one 16-byte `LoadU` + `TableLookupBytes` per group, reading from a zero-padded copy of the 48-byte header (a raw 16-byte load on the last group would run 4 bytes past the header). `DecodeInterleaved` now also rejects `comp_size < 48` up front.

hwy/contrib/coder/range_coder_test.cc
tvost2: Done - the test now uses `RandomState` + `Random64` from `test_util.h`.

hwy/contrib/coder/range_coder_test.cc
tvost2: Done - `AdjustedReps` now wraps the >1k size cases here and below.

tvost2:
Thanks for the review! All six comments addressed and pushed:

- SIMD kernel now runs on RVV/SVE — the four groups are `value0..value3` / `length0..length3` locals over `Full128<uint32_t>`, unrolled; only `HWY_SCALAR` still uses the scalar reference decoder. Kept the group at 128 bits rather than `CappedTag<16>` + `Lanes(d)` because `Normalize()` uses 128-bit `TableLookupBytes` renorm tables; widening needs those reworked per block (noted as a follow-up, happy to do it here if you'd rather).
- symbol bytes via `TruncateStore`; the 48-byte BE24 header via one 16-byte `LoadU` + `TableLookupBytes` per group (from a zero-padded copy), plus a `comp_size < 48` reject.
- `Full128<uint32_t>` for the tag.
- test uses `RandomState` / `Random64` from `test_util.h`, and `AdjustedReps` around the >1k sizes.

Also rebased onto master (the `BUILD` conflict was the `highwayhash` target landing next to `range_coder`). `range_coder_test` still passes on AVX2/SSE4/SSSE3/SSE2/EMU128 locally.

tvost2:
CI (riscv64, Clang-22 + GCC-16) caught that `BitsFromMask(D, M)` isn't available on RVV — pushed `05997418` swapping `Normalize()` to `StoreMaskBits`, which is universal (scalar/RVV/SVE/NEON/x86). `range_coder_test` still 10/10 on AVX2/SSE4/SSSE3/SSE2/EMU128 locally; waiting on the multiarch runners to re-confirm. (The two `{Clang-22,GCC-16} on AArch64` reds are the chronic runner-shutdown jobs, red on master too.)

jan-wassenberg (approved):
Very nice! TruncateTo is even better. And glad to see this now working also on scalable.

For wider vectors (renormalizing), we can do that in a follow-up for smaller diffs, and to make it easier to compare perf.

Issue #3250, opened 2026-08-06
Add CRC example


bkmgit:
CRC codes are used for checking data integrity and are widely used in applications.  CRC32 is often available as a machine instruction, but other variants  are not, see https://en.wikipedia.org/wiki/Cyclic_redundancy_check  Add a SIMD CRC implementation to the examples, such as CRC64-ECMA or CRC32C.

KimYannn:
maybe u could turn to https://github.com/google/crc32c

bkmgit:
@KimYannn Is the suggestion to use that as an example to start from for a simplified but cross platform example that uses highway, or to make a contribution to the repository instead to migrate it to use highway?

bkmgit:
Thanks for adding this @tvost2 and @tejas5038

PR #3340, opened 2026-09-02
Add portable SIMD CRC-64/XZ to hwy/contrib/crc

tvost2:
Addresses #3250 ("Add CRC example").

### Approach

The `O(n)` part of the message is reduced with **carryless-multiply
folding** — `CLMulLower` / `CLMulUpper` on a 128-bit accumulator, folding
16 bytes per iteration. Every Highway target provides carryless multiply
(natively via PCLMULQDQ / PMULL / `vpmsumd`, or through the portable
emulation in `generic_ops-inl.h`), so a single source runs on all of them
and returns the frozen check value.

- The two fold constants (`x^127 mod P` and `x^191 mod P`, bit-reflected)
  are derived at compile time with a `constexpr` reflected-CRC step, so
  the polynomial is stated once and there is no magic-constants table.
- After the fold loop the 128-bit accumulator (`≡ prefix mod P`) and the
  sub-16-byte head/tail are finished with a short bit-at-a-time reduction.
- The fold kernel holds only `Full128` vectors in named locals, so it runs
  on fixed and scalable SIMD alike; only `HWY_SCALAR` (1 lane) falls back to
  the bit-at-a-time path.

`crc::Crc64Xz(data, size, prev = 0)` is a streaming API: pass the previous
return value to continue over more data.

### Tests

`crc_test` verifies:
- the catalogue check value: `Crc64Xz("123456789") == 0x995DC9BBDF1939FA`;
- every length `0..600` plus `1023/1024/1025/4096` against an independent
  bit-at-a-time reference (no tables, no CLMUL);
- chunked calls equal a single call.

Passes on AVX3/AVX2/SSE4/SSSE3/SSE2/EMU128 locally; strict `-Werror` build
(`-Wconversion -Wsign-conversion -Wshadow -Wcast-align …`) and
`clang-format` are clean.

CRC-32C and replacing the bit-at-a-time finish with a CLMUL Barrett
reduction are natural follow-ups.


hwy/contrib/crc/crc-inl.h
```diff
@@ -0,0 +1,138 @@
...
+  // (RefXn63 offsets 128 and 64).
+  const V k = Dup128VecFromValues(d64, RefXn63(128), RefXn63(64));
+
+  V acc = BitCast(d64, LoadU(d8, data));
+  acc = Xor(acc, Dup128VecFromValues(d64, state, 0));
+
+  size_t pos = 16;
+  for (; pos + 16 <= size; pos += 16) {
+    const V next = BitCast(d64, LoadU(d8, data + pos));
+    acc = Xor(Xor(CLMulLower(acc, k), CLMulUpper(acc, k)), next);
```
jan-wassenberg: We can fuse two Xor into one Xor3.

hwy/contrib/crc/crc-inl.h
```diff
@@ -0,0 +1,138 @@
...
+  return crc;
+}
+
+// Bit-at-a-time reflected CRC on a "raw" state (no init / xor-out applied).
+HWY_INLINE uint64_t Bitwise(uint64_t crc, const uint8_t* data, size_t size) {
+  for (size_t i = 0; i < size; ++i) crc = Byte(crc, data[i]);
+  return crc;
+}
+
+#if HWY_TARGET != HWY_SCALAR && !HWY_HAVE_SCALABLE && !HWY_TARGET_IS_SVE
```
jan-wassenberg: I didn't see anything in the code that requires ruling out HAVE_SCALABLE and especially not TARGET_IS_SVE (the HWY_SVE2_128 target is both common and fixed-size).
Let's remove those two conditions. Generally, we only want to land code that works for both fixed and scalable SIMD.

jan-wassenberg (changes requested):

hwy/contrib/crc/crc-inl.h
tvost2: Done in 1b508a9c.

hwy/contrib/crc/crc-inl.h
tvost2: Done in 1b508a9c -- that was leftover caution from the HighwayHash PR, where State had vector members. The CRC kernel only holds Full128 vectors in named locals, so it compiles fine on scalable SIMD; guard is now just HWY_TARGET != HWY_SCALAR (1 lane can't do the 16-byte block). crc_test has no guard of its own, so the CLMUL fold path now runs on SVE2_128 / SVE / RVV too.

jan-wassenberg (dismissed):
Nice 👍

tvost2:
Rebased onto master to clear the `BUILD`/`CMakeLists`/`meson.build` conflict from #3339 and #3334 landing — the `crc-inl.h` / `crc_test.cc` diff is byte-identical to the approved version, only the three build files gained a line next to `range_coder`. CI is green. Could you re-approve when you have a moment? Thanks!

jan-wassenberg (approved):

PR #3341, opened 2026-09-02
Add Iguana ANS32 entropy decoder to hwy/contrib/iguana

tvost2:
First step of the **Iguana** `op_wishlist.md` item: a Highway port of Iguana's
**ANS32 entropy stage** — the 32-way interleaved rANS coder that is the
"decompress at 10+ GB/s" core of that compressor
(github.com/SnellerInc/sneller, `ion/zion/iguana`).

The upstream repo is gone, so this is ported from the pure-Go reference
(`ans32.go`, `ans_statistics.go`, recovered via the Go module proxy). The
bitstream is byte-for-byte compatible with it.

### What's here (`hwy/contrib/iguana/`)

| file | contents |
|---|---|
| `ans.h` / `ans.cc` | the frequency model (`observe`: 4-way histogram + normalization, with the empty-input and single-symbol edge cases), the dense-table (de)serialization (`EncodeFull` / `ansDecodeFullTable`), the scalar ANS32 encoder, and a scalar reference decoder |
| `ans-inl.h` | **SIMD `Ans32Decode`**. Each round decodes 32 symbols (16 "forward" + 16 "reverse" lanes) with `GatherIndex` into the dense table and one `Mul`/`Add`, then renormalizes the lanes whose state dropped below 2^16 by pulling one 16-bit word each from the two ends of the payload — vectorized with `Expand` (+ `Reverse` for the backward half). Works on `CappedTag<uint32_t, 16>`, so 4 / 8 / 16 lanes per group depending on the target. `HWY_SCALAR` and the scalable targets fall back to the scalar reference. |
| `ans_test.cc` | round-trips a range of sizes (0, sub-round, many rounds, tail remainders) and models (skewed / uniform / one / two symbols), and checks the SIMD decoder against the scalar reference bit-for-bit |

Wired into CMake, Bazel and Meson.

### Validation

- `ans_test` passes on AVX3 / AVX2 / SSE4 / SSSE3 / SSE2 / EMU128 locally.
- Strict `-Werror` build (`-Wconversion -Wsign-conversion -Wshadow -Wcast-align
  …`) and `clang-format` are clean.
- A standalone C++ port of the Go reference (built separately) round-trips every
  test case; the encoder here produces the same block, and both decoders
  reproduce the input.

### Not in this PR

Iguana's LZ77 layer (`decoder.go`), and `ANS1` / `ANS_nibble`. The ANS32
decoder's final reduction is still partly scalar (the sub-32-symbol tail);
`Expand`-heavy targets could also use a `TruncateTo` for the symbol pack. All
straightforward follow-ups.

### DIFF
--- a/hwy/contrib/iguana/ans-inl.h
+++ b/hwy/contrib/iguana/ans-inl.h
@@ -22,15 +22,17 @@
 // lanes whose state fell below 2^16 by pulling one 16-bit word each from the
 // forward / reverse halves of the payload - vectorized with Expand.
 //
-// State is held in up to 4 named vectors per half (fwd0..fwd3, rev0..rev3),
-// not an array: RVV/SVE vectors are sizeless types and cannot be array
-// elements. HWY_SCALAR is excluded at compile time (its 1-lane-only ops
-// can't instantiate RenormLane's multi-lane Repartition/LoadU at all, not
-// just uselessly -- confirmed by trying, not assumed). `Ans32DecodePayload`
-// additionally falls back to the scalar reference decoder at *runtime* for
-// a scalable (RVV/SVE) target whose hardware vector length is unusually
-// small -- everywhere else, including RVV/SVE with a typical vector length,
-// this vectorizes.
+// The 16+16 states are held in named vector locals (fwd0..fwd3, rev0..rev3):
+// RVV/SVE vectors are sizeless and cannot be array elements. The number of
+// groups (16 / Lanes(d), so 1, 2 or 4) is a template argument kNumVectors, so
+// each instantiation is straight-line `if constexpr`-guarded code the compiler
+// keeps in registers -- no pointer-to-vector array. `Ans32DecodePayload`
+// dispatches on it: a fixed target instantiates only the one group count its
+// (compile-time) Lanes(d) produces; a scalable target instantiates all three.
+// HWY_SCALAR is excluded at compile time (its 1-lane ops can't instantiate
+// RenormLane's multi-lane Repartition/LoadU -- confirmed by trying), and a
+// scalable target with an unusually small hardware vector length falls back to
+// the scalar reference decoder at runtime.
 
 #if defined(HIGHWAY_HWY_CONTRIB_IGUANA_ANS_INL_H_) == defined(HWY_TARGET_TOGGLE)
 #ifdef HIGHWAY_HWY_CONTRIB_IGUANA_ANS_INL_H_
@@ -119,41 +121,20 @@ HWY_INLINE hn::VFromD<D> RenormLane(D d, hn::VFromD<D> x, const uint8_t*& p) {
       mask, hn::Or(hn::ShiftLeft<hi::kAnsWordLBits>(x), expanded), x);
 }
 
-// Decodes `payload` (the rANS data, without the frequency table) using an
-// already-built dense `table`. Mirrors Ans32DecodePayloadScalar.
-HWY_INLINE bool Ans32DecodePayload(const uint8_t* HWY_RESTRICT payload,
-                                   size_t payload_size,
-                                   const hi::AnsDenseTable& table,
-                                   uint8_t* HWY_RESTRICT dst,
-                                   size_t orig_size) {
-  if (payload_size < 128) return false;
-  const uint32_t* HWY_RESTRICT tab = table.data();
-
-  const hn::CappedTag<uint32_t, 16> d;
-  const size_t n = hn::Lanes(d);  // <= 16; typically 4, 8 or 16
-  const size_t nv = n == 0 ? 0 : 16 / n;
-  // Falls back to the scalar reference decoder for lane counts that don't
-  // divide 16 into at most 4 groups -- only reachable here on a scalable
-  // (RVV/SVE) target with an unusually narrow hardware vector length
-  // (HWY_SCALAR is excluded at compile time above, never reaches this).
-  // Produces identical output, just without vectorization for that rare case.
-  if (n == 0 || nv == 0 || nv > 4 || n * nv != 16) {
-    return hi::Ans32DecodePayloadScalar(payload, payload_size, table, dst,
-                                        orig_size);
-  }
-
-  // Named variables, not an array: RVV/SVE vectors are sizeless types and
-  // cannot be array elements. A fixed array of POINTERS to them is fine
-  // (pointers are ordinary, fixed-size objects) and lets the loops below
-  // stay index-based like the original, unrolled-by-hand version -- only
-  // `nv` (<=4) of each are ever read.
-  using V = hn::VFromD<decltype(d)>;
-  V fwd0 = hn::Zero(d), fwd1 = hn::Zero(d), fwd2 = hn::Zero(d),
-    fwd3 = hn::Zero(d);
-  V rev0 = hn::Zero(d), rev1 = hn::Zero(d), rev2 = hn::Zero(d),
-    rev3 = hn::Zero(d);
-  V* const fwd[4] = {&fwd0, &fwd1, &fwd2, &fwd3};
-  V* const rev[4] = {&rev0, &rev1, &rev2, &rev3};
+// Decodes `payload` with `kNumVectors` (1, 2 or 4) vector groups per half.
+// The group count is a template argument so the unrolled body is straight-line
+// code in named locals rather than a runtime loop over an array of pointers.
+template <size_t kNumVectors, class D>
+HWY_INLINE bool Ans32DecodePayloadT(D d, size_t n,
+                                    const uint8_t* HWY_RESTRICT payload,
+                                    size_t payload_size,
+                                    const uint32_t* HWY_RESTRICT tab,
+                                    uint8_t* HWY_RESTRICT dst,
+                                    size_t orig_size) {
+  using V = hn::VFromD<D>;
+  V fwd0 = hn::Zero(d), rev0 = hn::Zero(d);
+  HWY_MAYBE_UNUSED V fwd1 = hn::Zero(d), fwd2 = hn::Zero(d), fwd3 = hn::Zero(d);
+  HWY_MAYBE_UNUSED V rev1 = hn::Zero(d), rev2 = hn::Zero(d), rev3 = hn::Zero(d);
 
   {
     HWY_ALIGN uint32_t s[32];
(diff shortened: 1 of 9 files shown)


hwy/contrib/iguana/ans-inl.h
```diff
@@ -0,0 +1,218 @@
...
+#include <stddef.h>
+#include <stdint.h>
+#include <string.h>  // memcpy
+
+#include "hwy/contrib/iguana/ans.h"
+#include "hwy/highway.h"
+
+HWY_BEFORE_NAMESPACE();
+namespace hwy {
+namespace HWY_NAMESPACE {
```
jan-wassenberg: I suspect we'd be better off keeping HWY_NAMESPACE as the innermost - generally we want to avoid direct mention of HWY_NAMESPACE:: , so hwy::iguana_ans would be safer.

hwy/contrib/iguana/ans-inl.h
```diff
@@ -0,0 +1,218 @@
...
+namespace iguana_ans {
+
+namespace hi = hwy::iguana;
+
+// The kernel keeps the 32 rANS states in fixed-size 4-lane-or-wider u32 vectors
```
jan-wassenberg: Here also, we'd like to support scalable targets. Can these states simply be a uint32_t[32]?

hwy/contrib/iguana/ans-inl.h
```diff
@@ -0,0 +1,218 @@
...
+  const RebindToSigned<D> di;
+  const VFromD<D> slot = And(x, Set(d, hi::kAnsFreqMask));
+  const VFromD<D> t = GatherIndex(d, table, BitCast(di, slot));
+  const VFromD<D> freq = And(t, Set(d, hi::kAnsFreqMask));
+  const VFromD<D> bias =
+      And(ShiftRight<hi::kAnsWordMBits>(t), Set(d, hi::kAnsFreqMask));
+
+  HWY_ALIGN uint32_t syms[16];
+  StoreU(ShiftRight<24>(t), d, syms);
+  for (size_t i = 0; i < Lanes(d); ++i) out[i] = static_cast<uint8_t>(syms[i]);
```
jan-wassenberg: Probably better to use TruncateTo u8, then Store to out (possibly mentioned already in the commit message).

hwy/contrib/iguana/ans-inl.h
```diff
@@ -0,0 +1,218 @@
...
+
+  return Add(Mul(freq, ShiftRight<hi::kAnsWordMBits>(x)), bias);
```
jan-wassenberg: MulAdd - Arm targets can fuse them.

hwy/contrib/iguana/ans-inl.h
```diff
@@ -0,0 +1,218 @@
...
+                (static_cast<uint32_t>(payload[lane * 4 + 1]) << 8) |
+                (static_cast<uint32_t>(payload[lane * 4 + 2]) << 16) |
+                (static_cast<uint32_t>(payload[lane * 4 + 3]) << 24);
+      const size_t o = rev_off + static_cast<size_t>(lane) * 4;
+      s[lane + 16] = static_cast<uint32_t>(payload[o]) |
+                     (static_cast<uint32_t>(payload[o + 1]) << 8) |
+                     (static_cast<uint32_t>(payload[o + 2]) << 16) |
+                     (static_cast<uint32_t>(payload[o + 3]) << 24);
+    }
+    for (size_t g = 0; g < nv; ++g) {
```
jan-wassenberg: For scalable vectors, we could unroll this loop manually, and have fwd0..fwd3 vector variables, and then load them with strides of n, but capped to 32 in total.

hwy/contrib/iguana/ans-inl.h
```diff
@@ -0,0 +1,218 @@
...
+    }
+  }
+
+  const uint8_t* pf = payload + 64;
+  const uint8_t* pr = payload + payload_size - 64;
+  size_t pos = 0;
+
+  // Vectorized rounds, kept clear of the point where the two halves meet.
+  while (pos + 32 <= orig_size && pf + 64 <= pr - 64) {
+    for (size_t g = 0; g < nv; ++g) {
```
jan-wassenberg: Can we fuse these loops? For scalable, ideally just unroll them manually. It might help to have a lambda (with HWY_ATTR before the opening brace).

hwy/contrib/iguana/ans.h
```diff
@@ -0,0 +1,105 @@
...
+#include <stdint.h>
+
+#include <vector>
+
+namespace hwy {
+namespace iguana {
+
+// ------------------------------ Constants (bitstream format)
+
+constexpr uint32_t kAnsWordLBits = 16;
```
jan-wassenberg: These don't have to be exposed in the interface, right? Want to move to inl.h or a private header?

hwy/contrib/iguana/ans_test.cc
```diff
@@ -0,0 +1,117 @@
...
+// clang-format on
+
+HWY_BEFORE_NAMESPACE();
+namespace hwy {
+namespace HWY_NAMESPACE {
+namespace {
+
+namespace ans = hwy::HWY_NAMESPACE::iguana_ans;
+
+uint64_t NextRandom(uint64_t& state) {
```
jan-wassenberg: We have a Random32/64 in test_util.h.

hwy/contrib/iguana/ans_test.cc
```diff
@@ -0,0 +1,117 @@
...
+  HWY_ASSERT(hwy::iguana::Ans32DecodeScalar(enc.data(), enc.size(), dec2.data(),
+                                            data.size()));
+  HWY_ASSERT(data.empty() ||
+             memcmp(dec2.data(), data.data(), data.size()) == 0);
+}
+
+void TestRoundTripSizes() {
+  const size_t kSizes[] = {0,    1,    2,     31,    32,   33,  63,
+                           64,   65,   100,   255,   256,  257, 1024,
+                           4096, 4097, 65535, 65536, 70000};
```
jan-wassenberg: Please wrap the >1024 sizes in AdjustedReps so that debug/emulated test don't time out.

hwy/contrib/iguana/ans_test.cc
```diff
@@ -0,0 +1,117 @@
...
+    const size_t n = kSizes[idx];
+    RoundTrip(MakeData(n, n * 3 + 1, 3));
+    RoundTrip(MakeData(n, n * 5 + 2, 1));
+    RoundTrip(MakeData(n, n * 7 + 3, 0));
+  }
+}
+
+void TestRoundTripModels() {
+  // Large skewed input (compresses; exercises many vector rounds).
+  RoundTrip(MakeData(300000, 0xABCDEF, 3));
```
jan-wassenberg: Also AdjustedReps here for the larger sizes.

jan-wassenberg (changes requested):

jan-wassenberg (commented):
Everything looks good, just one remaining improvement on the nv/fwd0..3 unrolling, WDYT?

hwy/contrib/iguana/ans-inl.h
```diff
@@ -109,12 +129,32 @@ HWY_INLINE bool Ans32DecodePayload(const uint8_t* HWY_RESTRICT payload,
...
+  // cannot be array elements. A fixed array of POINTERS to them is fine
+  // (pointers are ordinary, fixed-size objects) and lets the loops below
+  // stay index-based like the original, unrolled-by-hand version -- only
+  // `nv` (<=4) of each are ever read.
+  using V = hn::VFromD<decltype(d)>;
+  V fwd0 = hn::Zero(d), fwd1 = hn::Zero(d), fwd2 = hn::Zero(d),
+    fwd3 = hn::Zero(d);
+  V rev0 = hn::Zero(d), rev1 = hn::Zero(d), rev2 = hn::Zero(d),
+    rev3 = hn::Zero(d);
+  V* const fwd[4] = {&fwd0, &fwd1, &fwd2, &fwd3};
```
jan-wassenberg: hm, I previously wrote a lengthy comment but Github seems to have eaten it?
Have we checked the codegen here? I'm a bit skeptical.
Rather than a pointer to vector, which does work but can be inefficient, I think it could be useful to compute nv as now, but convert it to a template argument kNumVectors. At the top level Decode, we can compute nv and then dispatch to Decode</*kNumVectors=*/2> etc. Inside it, we can have `if constexpr (kNumVectors >= 2) fwd1 = ..`. That way we unroll as appropriate for the current target. The only cost is compile time + binary size, but I suppose we could cap the possible nv values using HWY_MIN_BYTES and HWY_MAX_BYTES. Does that make sense?

tvost2: Agreed, the logic makes total sense. Taking `&fwd0` forces the vector to have an address, so it can't stay in a register — the `V* fwd[4]` array spills all eight state vectors to the stack and every `*fwd[g]` in the hot loop becomes a load+store. Converting `nv` to a template argument fixes that cleanly, and it's the same shape vqsort already uses.

Plan:
- `Ans32DecodePayload` keeps computing `n` / `nv` at runtime, then dispatches: `nv == 1 → Impl<1>`, `2 → Impl<2>`, `4 → Impl<4>` (else the existing scalar fallback).
- `Impl<kNumVectors>` declares `fwd0..fwd3` / `rev0..rev3` as plain named locals and guards every use with `if constexpr (kNumVectors >= N)`, so each instantiation is straight-line with no addresses taken.
- `nv` is `16 / Lanes(CappedTag<uint32_t, 16>)`, so the only reachable values are 1/2/4; I'll bound the instantiations with `HWY_MAX_BYTES` / `HWY_MIN_BYTES` so a target that can only be one width doesn't compile the others.

Also rebased onto master — #3339 and #3334 landing put `range_coder.cc` next to `ans.cc` in `meson.build`; resolved, branch is mergeable again. Pushing the `kNumVectors` refactor shortly.

jan-wassenberg:
The plan sounds great :)

hwy/contrib/iguana/ans-inl.h
tvost2: Done in `02c7e3d0`. `nv` is now the template argument `kNumVectors`: `Ans32DecodePayloadT<1|2|4>` declares `fwd0..fwd3` / `rev0..rev3` as named locals and guards each group with `if constexpr (kNumVectors >= N)` — no `V*` array, no addresses taken. `Ans32DecodePayload` dispatches: on a fixed target `Lanes(d)` is constexpr so only the one group count is instantiated (`if constexpr (!HWY_HAVE_SCALABLE)`); a scalable target instantiates all three and picks at runtime. `ans_test` still 10/10 on AVX2/SSE4/SSSE3/SSE2/EMU128.

jan-wassenberg (approved):
Very nice, congrats on also supporting scalable vectors here, definitely a non-trivial use case.

PR #3342, opened 2026-09-02
Add Iguana container + LZ77 layer to hwy/contrib/iguana

tvost2:
Completes the **Iguana** `op_wishlist.md` item: the container + LZ77 layer on
top of the ANS32 entropy stage from #3341, so `hwy/contrib/iguana/` is now a
working portable Iguana codec (github.com/SnellerInc/sneller, ion/zion/iguana).

**Stacked on #3341** — review that one first; the diff here is only the new
`iguana.{h,cc}` / `iguana-inl.h` / `iguana_test.cc`.

Ported from the pure-Go reference (`decoder.go`, `encoder.go`, `stream.go`,
recovered via the Go module proxy). Bitstream-compatible with it.

### What's here

| file | contents |
|---|---|
| `iguana.h` / `iguana.cc` | the container format (backward control varints, `CopyRaw` / `DecodeIguana` / `DecodeANS32` commands), the LZ77 stage (`DecompressIguanaLZ`: token loop + overlap-aware wild copy), and the scalar encoder — a Lizard-style hash-chain match finder (`kChainBits`, `kHashBytes`, `kHistSize`) with the reference's repeat-offset and end-of-buffer handling. `DecompressScalar` ties it together. |
| `iguana-inl.h` | `Decompress` — the same container loop, but each entropy-coded stream goes through the **SIMD** ANS32 decoder from #3341 (`iguana_ans::Ans32Decode`). The container parsing and LZ token loop are inherently serial, so they stay in the target-independent helpers. Output is identical to `DecompressScalar`. |
| `iguana_test.cc` | round-trips a range of sizes and data models (random / skewed / repetitive text / RLE), the >64 KiB far-offset (24-bit) path, and a re-sliced short string (raw-copy fallback); cross-checks the SIMD path against the scalar reference. |

Covers the `EncodingIguana` / `EntropyANS32` pipeline that `Encoder.Compress`
produces. `ANS1` / `ANS_nibble` stream modes are declared-but-rejected (small
scalar coders; a follow-up).

### Validation

- `iguana_test` passes on AVX3 / AVX2 / SSE4 / SSSE3 / SSE2 / EMU128 locally.
- A standalone C++ port of the Go reference (built separately) round-trips
  every test case, including the far-offset path; the encoder here produces the
  same block and both decoders reproduce the input.
- Strict `-Werror` (`-Wconversion -Wsign-conversion -Wshadow -Wcast-align …`)
  and `clang-format` are clean.

### Follow-ups

The LZ wild copy could use wide Highway loads/stores for non-overlapping
matches; `ANS1` / `ANS_nibble`; a `HWY_DYNAMIC_DISPATCH` entry point.


tvost2:
@jan-wassenberg you will want this?

jan-wassenberg:
Thanks for pinging, I had lost track of this one.
Yes, sounds like a useful thing to add.
Please rebase and I'll then have a look :)

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+// ------------------------------ stream reader (LZ77 input)
+
+struct StreamReader {
+  const uint8_t* data = nullptr;
+  size_t size = 0;
+  size_t cursor = 0;
+
+  bool Empty() const { return cursor >= size; }
+  size_t Remaining() const { return size - cursor; }
+  bool Have(size_t n) const { return cursor + n <= size; }
```
jan-wassenberg: HaveBytes would be more clear. Maybe also RemainingBytes.
Also prefer IsEmpty, Empty() could also be understood as a verb.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+      uint32_t x = static_cast<uint32_t>(v >> (i * 7)) & 0x7Fu;
+      if (i == 0) x |= 0x80u;
+      ctrl.push_back(static_cast<uint8_t>(x));
+    }
+  }
+};
+
+// ------------------------------ match finder (encoder)
+
+constexpr size_t kChainSize = size_t{1} << kChainBits;
```
jan-wassenberg, line 113: More comments would be helpful. What is a chain?

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+    uint64_t a, b;
+    memcpy(&a, src + lo + m, 8);
+    memcpy(&b, src + hi + m, 8);
+    const uint64_t d = a ^ b;
+    if (d == 0) {
+      m += 8;
+      continue;
+    }
+    unsigned tz = 0;
+    while (((d >> tz) & 1) == 0) ++tz;
```
jan-wassenberg: Can use Num0BitsBelowLS1Bit_Nonzero64.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+  return m;
+}
+
+bool IsLegal(int64_t offs, int64_t length) {
+  return offs <= static_cast<int64_t>(kMaxU16) || length > kMaxShortMatchLen;
+}
+
+// Extends a candidate match backwards; sets len 0 if the pair is not encodable.
+void MatchExtend(const uint8_t* src, size_t src_len, int64_t minto,
+                 int64_t from, int64_t to, int64_t* tp, int64_t* mp,
```
jan-wassenberg: Longer var names, please :) Also consider HWY_RESTRICT - are these potentially overlapping?

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+        pos += kSkipStep;
+      }
+    }
+    literals.insert(literals.end(), src + litpos, src + src_len);
+  }
+};
+
+// ------------------------------ LZ77 wild copy (decoder)
+
+void WildCopy(Bytes& dst, size_t pos, size_t match_len) {
```
jan-wassenberg: What does wild mean in this context?

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+        if (!ok) return false;
+      }
+      if (lit_len > 0) {
+        const uint8_t* p = lit.Sequence(static_cast<size_t>(lit_len), &ok);
+        if (!ok) return false;
+        dst.insert(dst.end(), p, p + lit_len);
+      }
+      if ((token & 0x80) == 0) {
+        last_offs = -static_cast<int64_t>(o16.U16(&ok));
+        if (!ok) return false;
```
jan-wassenberg: For malformed inputs, we have to check and reject last_offs == 0.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+            if (data_cursor + ulens[i] > src_size) return false;
+            streams[i].data = src + data_cursor;
+            streams[i].size = static_cast<size_t>(ulens[i]);
+            data_cursor += ulens[i];
+          } else if (em == 1) {  // EntropyANS32
+            const uint64_t clen = ReadControlVarUint(src, &ctrl, &ok);
+            if (!ok || data_cursor + clen > src_size) return false;
+            ent_bufs.emplace_back();
+            Bytes& b = ent_bufs.back();
+            b.resize(static_cast<size_t>(ulens[i]));
```
jan-wassenberg: Requires a cap to prevent 'zip bombs'.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+    const uint8_t* p = lit.Sequence(rem, &ok);
+    if (!ok) return false;
+    dst.insert(dst.end(), p, p + rem);
+  }
+  return true;
+}
+
+// ------------------------------ container
+
+std::vector<uint8_t> Compress(const uint8_t* data, size_t size) {
```
jan-wassenberg: We require `HWY_CONTRIB_DLLEXPORT` on definitions and declarations of Compress, DecompressScalar, DecompressIguanaLZ, and ReadControlVarUint.

hwy/contrib/iguana/iguana.h
```diff
@@ -0,0 +1,96 @@
...
+#include <vector>
+
+#include "hwy/contrib/iguana/ans.h"
+
+namespace hwy {
+namespace iguana {
+
+// ------------------------------ Format constants
+
+constexpr int kIguanaChunkSize = 32;
```
jan-wassenberg: Also good to move these to a detail.h. And prefer size_t, though this will require updates to the code and casting.

hwy/contrib/iguana/iguana.h
```diff
@@ -0,0 +1,96 @@
...
+
+// Compresses `data` into a complete Iguana block (EncodingIguana / ANS32).
+std::vector<uint8_t> Compress(const uint8_t* data, size_t size);
+
+// Decompresses a block produced by Compress. Returns false on malformed input.
+// The SIMD path in iguana-inl.h produces identical output.
+bool DecompressScalar(const uint8_t* src, size_t src_size,
+                      std::vector<uint8_t>& out);
+
+// ------------------------------ Internals shared with iguana-inl.h
```
jan-wassenberg: Can also move to detail.h.

hwy/contrib/iguana/iguana_test.cc
```diff
@@ -0,0 +1,128 @@
...
+// clang-format on
+
+HWY_BEFORE_NAMESPACE();
+namespace hwy {
+namespace HWY_NAMESPACE {
+namespace {
+
+namespace ig = hwy::HWY_NAMESPACE::iguana_full;
+
+uint64_t NextRandom(uint64_t& state) {
```
jan-wassenberg: Random64.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+  }
+};
+
+// ------------------------------ match finder (encoder)
+
+constexpr size_t kChainSize = size_t{1} << kChainBits;
+
+uint32_t HashSeq(const uint8_t* seq) {
+  uint64_t u;
+  memcpy(&u, seq, 8);
```
jan-wassenberg: We support big endian, hence this and all other memcpy would have to use a (not yet added) Load64LE.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+  void Command(uint8_t v) {
+    if (last_command_offset >= 0) {
+      ctrl[static_cast<size_t>(last_command_offset)] &= kCommandMask;
+    }
+    last_command_offset = static_cast<int64_t>(ctrl.size());
+    ctrl.push_back(static_cast<uint8_t>(v | kLastCommandMarker));
+  }
+  void VarUint(uint64_t v) {
+    int bit_len = 0;
+    for (uint64_t t = v; t != 0; t >>= 1) ++bit_len;
```
jan-wassenberg: I think this can be replaced with v == 0 ? 0 : static_cast<int>(64 - hwy::Num0BitsAboveMS1Bit_Nonzero64(v)).

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+
+  for (;;) {
+    if (ctrl < 0) return false;
+    const uint8_t cmd = src[ctrl];
+    --ctrl;
+
+    switch (cmd & kCommandMask) {
+      case kCmdCopyRaw: {
+        const uint64_t n = ReadControlVarUint(src, &ctrl, &ok);
+        if (!ok || data_cursor + n > src_size) return false;
```
jan-wassenberg: For such bounds checks, better to subtract because addition can wrap around.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+        out.insert(out.end(), src + data_cursor, src + data_cursor + n);
+        data_cursor += n;
+        break;
+      }
+      case kCmdDecodeANS32: {
+        const uint64_t lu = ReadControlVarUint(src, &ctrl, &ok);
```
jan-wassenberg: We cast this to size_t so on 32-bit platforms we'd have to check this isn't above 4 GiB. Even better, we should sanity check and/or clamp all sizes used to resize/allocate.

hwy/contrib/iguana/iguana-inl.h
```diff
@@ -0,0 +1,140 @@
...
+
+// Decompresses a block produced by hi::Compress, appending to `out`. The
+// entropy-coded streams go through the SIMD ANS32 decoder. Returns false on
+// malformed input; output matches hi::DecompressScalar.
+HWY_INLINE bool Decompress(const uint8_t* HWY_RESTRICT src, size_t src_size,
+                           std::vector<uint8_t>& out) {
+  if (src_size == 0) return false;
+  bool ok = true;
+  int64_t ctrl = static_cast<int64_t>(src_size) - 1;
+  const uint64_t uncompressed_len = hi::ReadControlVarUint(src, &ctrl, &ok);
```
jan-wassenberg: Vulnerable to zip bomb, we should check this is a reasonable size (say 1 GiB max), and also use it to cap the allocations.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+    pos += dist;
+    match_len -= dist;
+  }
+}
+
+}  // namespace
+
+uint64_t ReadControlVarUint(const uint8_t* src, int64_t* cursor, bool* ok) {
+  uint64_t r = 0;
+  while (*cursor >= 0) {
```
jan-wassenberg: We should limit the number of bytes read and/or ensure that r << 7 doesn't overflow.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+    }
+    literals.insert(literals.end(), src + litpos, src + src_len);
+  }
+};
+
+// ------------------------------ LZ77 wild copy (decoder)
+
+void WildCopy(Bytes& dst, size_t pos, size_t match_len) {
+  if (pos + match_len <= dst.size()) {
+    dst.insert(dst.end(), dst.begin() + static_cast<ptrdiff_t>(pos),
```
jan-wassenberg: This is UB - reallocating invalidates iterators. We can dst.resize(old_size + match_len) first, then CopyBytes.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+    unsigned tz = 0;
+    while (((d >> tz) & 1) == 0) ++tz;
+    return m + static_cast<int64_t>(tz / 8);
+  }
+  while (n - (hi + m) > 0 && src[lo + m] == src[hi + m]) ++m;
+  return m;
+}
+
+bool IsLegal(int64_t offs, int64_t length) {
+  return offs <= static_cast<int64_t>(kMaxU16) || length > kMaxShortMatchLen;
```
jan-wassenberg: We should also check length <= kMaxU24.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+
+void AppendVarUint(Bytes& s, uint32_t v) {
+  if (v < kVarThresh1) {
+    s.push_back(static_cast<uint8_t>(v));
+  } else if (v < kVarThresh3) {
+    s.push_back(254);
+    s.push_back(static_cast<uint8_t>(v % 254));
+    s.push_back(static_cast<uint8_t>(v / 254));
+  } else {
+    HWY_DASSERT(v < 254u * 254 * 254);  // fits the encoder's stream varint
```
jan-wassenberg, line 56: What if we are compressing long runs of zeros or literals? Probably we'd have to split into multiple runs?

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+        AppendVarUint(var_match_len, match_len - kLongBase);
+      }
+      tokens.push_back(static_cast<uint8_t>(token));
+    }
+    last_encoded_offset = offs;
+  }
+
+  void CompressSrc() {
+    constexpr int64_t kSkipStep = 2;
+    for (auto& e : chains) e = {};
```
jan-wassenberg: The ctor already zero-initialized these, right? Can probably skip.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+        }
+      }
+    }
+  }
+
+  for (size_t i = cw.ctrl.size(); i-- > 0;) dst.push_back(cw.ctrl[i]);
+  return dst;
+}
+
+bool DecompressScalar(const uint8_t* src, size_t src_size, Bytes& out) {
```
jan-wassenberg: This mostly duplicates the logic in iguana-inl.h around line 65, right? Better to have a single implementation. We can template it with a functor that wraps either Ans32Decode or Ans32DecodeScalar.

hwy/contrib/iguana/iguana_test.cc
```diff
@@ -0,0 +1,128 @@
...
+// NOLINTNEXTLINE(google-readability-namespace-comments)
+}  // namespace HWY_NAMESPACE
+}  // namespace hwy
+HWY_AFTER_NAMESPACE();
+
+#if HWY_ONCE
+namespace hwy {
+HWY_BEFORE_TEST(IguanaTest);
+HWY_EXPORT_AND_TEST_P(IguanaTest, TestRoundTripSizes);
+HWY_EXPORT_AND_TEST_P(IguanaTest, TestRoundTripStructure);
```
jan-wassenberg, line 468: Some corner cases to check (and ensure we reject):
1) Empty input
2) First LZ token with repeat-offset bit set or zero o16 offset
3) Truncated control varints and truncated payload streams
4) Header uncompressed_len smaller or larger than actual decoded output

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,626 @@
...
+    }
+  }
+
+  const size_t rem = lit.Remaining();
+  if (rem > 0) {
+    const uint8_t* p = lit.Sequence(rem, &ok);
+    if (!ok) return false;
+    dst.insert(dst.end(), p, p + rem);
+  }
+  return true;
```
jan-wassenberg: Let's also validate o16.IsEmpty() && o24.IsEmpty() && vll.IsEmpty() && vml.IsEmpty().

jan-wassenberg (changes requested):
Numerous comments because there are a number of security concerns when implementing a codec.
After these, let's get the agent to do a full security review.

BTW for DeserializeAnsTable, we should check total == kAnsWordM || total == kAnsWordM - 1 and freqs[i] < kAnsWordM.

tvost2:
## Security review: Iguana container + LZ77 (`hwy/contrib/iguana`)

Reviewed at `5a3d9ffc` (the review fixes on top of `a8addba2`).

**Threat model.** `Decompress` / `DecompressScalar` parse *untrusted* input: every length,
offset and command byte in a block is attacker-controlled. The properties that matter:

1. no out-of-bounds read/write and no UB;
2. no allocation an attacker can steer with a small input ("zip bomb");
3. no integer overflow in the size arithmetic;
4. malformed input is *rejected*, not decoded "best effort".

### What changed in response to the review

| Area | Issue | Fix |
|---|---|---|
| Container | `data_cursor + n > src_size` can wrap around | checked by subtraction (`n > src_size - data_cursor`) at all four sites: raw copy, ANS32 payload, raw stream, entropy stream |
| Amplification | declared uncompressed length, per-stream and per-command lengths were unbounded | all capped at `kMaxUncompressedSize` (1 GiB) *before* any `resize`/`insert`; the LZ stage and `CopyMatch` cap the output too |
| LZ77 | a match with `last_offs == 0` copied from before the start of the output | rejected explicitly (`last_offs == 0` covers both the "first token with the repeat-offset bit" and the "zero 16-bit offset" cases) |
| LZ77 | leftover bytes in the offset/length streams were silently ignored | all four auxiliary streams must be fully consumed, otherwise reject |
| LZ77 | `dst.insert(dst.begin() + pos, ...)` while `dst` reallocates = UB (invalidated iterators) | `resize` first, then copy forward; the forward copy is also what makes overlapping runs correct |
| Control varint | `r << 7` could shift past the width of `uint64_t` | at most 10 bytes, and the 10th (most significant) may contribute a single bit |
| ANS model | only `total > kAnsWordM` was rejected, so a table summing to *less* left dense-table entries unassigned | require `total == kAnsWordM || total == kAnsWordM - 1`, and `freqs[i] < kAnsWordM` |
| Encoder | a run longer than the 4-byte varint can carry tripped a `HWY_DASSERT` (silent truncation in release) | long literal runs are split into literal-only tokens; matches are clamped to `kMaxEncodableMatchLen` |
| Endianness | 8-byte `memcpy` in `HashSeq` and `Lcp` | both use `Load64LE`, so `Num0Bits*` observes the intended byte order on big-endian hosts |

### Findings

**F1 (fixed, high) — UB and out-of-bounds write in the LZ77 copy.** The wild copy inserted
from `dst.begin() + pos` while appending to the same vector, which is undefined behaviour as
soon as the insertion reallocates; with `match_len` taken from the stream this is directly
attacker-influenced. It is now `resize` + forward copy, and the output is capped.

**F2 (fixed, high) — decompression bomb.** A ~100-byte block could declare a multi-gigabyte
`uncompressed_len` and have the decoder `resize()` that much before finding the first bad
byte. Every allocation is now bounded by `kMaxUncompressedSize` (1 GiB) and validated before
it happens.

**F3 (fixed, medium) — attacker-controlled match offset.** A match referring to offset 0
(no previous match, or an explicit zero 16-bit offset) decoded from before the output
buffer. Rejected now.

**F4 (fixed, medium) — dense-table corruption in the ANS model.** A serialized table that
summed to less than `kAnsWordM` was accepted, leaving entries of the 4096-entry dense table
unassigned, so decoding could take a symbol/frequency pair that does not correspond to the
model. The totals and per-symbol bounds are now enforced.

**F5 (fixed, low) — silent truncation in the encoder.** `AppendVarUint` asserted that a run
fits the 4-byte varint; in release builds it truncated, producing a block the reference
decoder would not reproduce. Long literal runs are split into several literal-only tokens,
and match lengths are clamped to what one token can carry.

**F6 (fixed, low) — big-endian mismatch.** `Lcp` XORed two `memcpy`-loaded 64-bit words and
then located the first differing *bit*. On a big-endian host the lowest set bit of that word
is the highest-order byte, so the reported match length was wrong. `Load64LE` fixes it
(`HashSeq` had the same issue).

### Verified sound (no change required)

* `Ans32DecodePayloadScalar`: rejects `src_size < 128`, and **every** renormalization is
  bounds-checked (`cursor_fwd + 2 > cursor_rev`, `cursor_rev < cursor_fwd + 2`), so the
  16-bit reads cannot leave the payload; writes stop at `orig_size`, which the caller
  allocated exactly.
* `StreamReader`: all reads go through `HaveBytes`, now subtraction-based, so a truncated
  stream sets `ok = false` instead of reading past the end.
* The container rejects the unimplemented stream modes (`ANS1`, `ANS_nibble`) and requires
  the last-command marker before reporting success.
* `DecompressBlock` is now the single implementation of the container loop, shared by the
  scalar and SIMD paths, so the two cannot drift apart.

### Residual risk and recommendations

1. **The 1 GiB cap is policy, not a guarantee.** A legitimate-looking block can still ask
   for 1 GiB. A caller processing many blocks should keep its own budget; an explicit
   max-output parameter would let it make that decision instead of the codec.
2. **Fuzzing.** `TestRejectsMalformed` sweeps every truncation and periodic single-byte
   mutations of a valid block, but this codec is a natural OSS-Fuzz target
   (`LLVMFuzzerTestOneInput` → `DecompressScalar`), which would also cover the SIMD decoder
   and the table deserializer much more thoroughly than a fixed corpus can.
3. `DecompressIguanaLZ` and `ReadControlVarUint` are now `HWY_CONTRIB_DLLEXPORT`ed, so they
   are part of the exported surface of the contrib library; changing their signature later
   is an ABI break.
4. Backwards-compatibility note for `kMaxUncompressedSize`: it is a *decoder* limit, so any
   block the encoder can produce beyond 1 GiB would no longer round-trip. The encoder does
   not produce such blocks today; if that changes, the two limits must move together.

Local evidence: `iguana_test` 15/15 (10 round-trip + 5 malformed-input) and `ans_test`
10/10 pass with mingw-w64, exercising AVX3_SPR / AVX3_ZEN4 / AVX3_DL / AVX3 / AVX2 / SSE4 /
SSSE3 / SSE2 / EMU128.

hwy/contrib/iguana/detail.h
```diff
@@ -0,0 +1,269 @@
...
+  bool ok = true;
+  int64_t ctrl = static_cast<int64_t>(src_size) - 1;
+  const uint64_t uncompressed_len = ReadControlVarUint(src, &ctrl, &ok);
+  if (!ok) return false;
+  out.clear();
+  if (uncompressed_len == 0) return true;
+  if (uncompressed_len > kMaxUncompressedSize) return false;
+
+  size_t data_cursor = 0;
+  std::vector<std::vector<uint8_t>> ent_bufs;
```
jan-wassenberg: This can accumulate across multiple kCmdDecodeIguana - we can clear it after that succeeds?

hwy/contrib/iguana/detail.h
```diff
@@ -0,0 +1,269 @@
...
+        out.insert(out.end(), src + data_cursor,
+                   src + data_cursor + static_cast<size_t>(n));
+        data_cursor += static_cast<size_t>(n);
+        break;
+      }
+      case kCmdDecodeANS32: {
+        const uint64_t lu = ReadControlVarUint(src, &ctrl, &ok);
+        const uint64_t lc = ReadControlVarUint(src, &ctrl, &ok);
+        if (!ok || lc > src_size - data_cursor) return false;
+        if (lu > kMaxUncompressedSize - out.size()) return false;
```
jan-wassenberg: I think we want to compare against uncompressed_len - out.size(). Also for n above.

hwy/contrib/iguana/detail.h
```diff
@@ -0,0 +1,269 @@
...
+          }
+        }
+        if (!DecompressIguanaLZ(out, streams)) return false;
+        break;
+      }
+      default:
+        return false;
+    }
+
+    if (cmd & kLastCommandMarker) return true;
```
jan-wassenberg: We could verify `out.size() == uncompressed_len && data_cursor == static_cast<size_t>(ctrl + 1);`.

hwy/contrib/iguana/detail.h
```diff
@@ -0,0 +1,269 @@
...
+
+  for (;;) {
+    if (ctrl < 0) return false;
+    const uint8_t cmd = src[static_cast<size_t>(ctrl)];
+    --ctrl;
+
+    switch (cmd & kCommandMask) {
+      case kCmdCopyRaw: {
+        const uint64_t n = ReadControlVarUint(src, &ctrl, &ok);
+        if (!ok || n > src_size - data_cursor) return false;
```
jan-wassenberg: Instead of src_size - data_cursor checks, we also have to check that the forward and backward haven't overlapped. We could add some kind of bounds-check lambda:
`return ctrl + 1 < static_cast<int64_t>(data_cursor) || len > static_cast<size_t>(ctrl + 1) - data_cursor`.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,596 @@
...
+
+  void BestMatchAt(int64_t litpos, int64_t pos, int64_t* tp, int64_t* mp,
+                   int64_t* len) {
+    *tp = pos;
+    *mp = 0;
+    *len = 0;
+    const int64_t rep = pos - static_cast<int64_t>(last_encoded_offset);
+    if (rep >= 0 && rep < pos) {
+      *mp = rep;
+      *len = Lcp(src, src_len, rep, pos);
```
jan-wassenberg: Requires clamping to kMaxEncodableMatchLen.

meson.build
```diff
@@ -177,6 +179,7 @@ hwy_contrib_headers = files(
 hwy_contrib_sources = files(
     'hwy/contrib/coder/range_coder.cc',
     'hwy/contrib/iguana/ans.cc',
+    'hwy/contrib/iguana/iguana.cc',
```
jan-wassenberg, line 183: Missing the detail.h here and in CMakeLists.txt :)

hwy/contrib/iguana/detail.h
```diff
@@ -0,0 +1,269 @@
...
+    return r;
+  }
+};
+
+// ------------------------------ Control varint (base-128, read backwards)
+
+// Reads a base-128 varint backwards from src[*cursor], moving *cursor before
+// the consumed bytes. The first byte read is the most significant one. Sets
+// *ok=false on underflow, or if the value would not fit in 64 bits.
+HWY_CONTRIB_DLLEXPORT uint64_t ReadControlVarUint(const uint8_t* src,
```
jan-wassenberg: This is pretty lightweight, and also an implementation detail - we should only export the public API, and could inline this here.

hwy/contrib/iguana/detail.h
```diff
@@ -0,0 +1,269 @@
...
+// One of the six token/literal/offset streams handed to the LZ77 stage.
+struct IguanaStream {
+  const uint8_t* data = nullptr;
+  size_t size = 0;
+};
+
+// The LZ77 stage: expands the six streams into `dst` (appended). The token
+// loop is inherently serial, so this stays scalar on the SIMD path too.
+// Returns false on malformed input. Defined in iguana.cc.
+HWY_CONTRIB_DLLEXPORT bool DecompressIguanaLZ(
```
jan-wassenberg: Let's also inline here, we do not want to export anything in this header.

hwy/contrib/iguana/iguana-inl.h
```diff
@@ -0,0 +1,67 @@
...
+namespace ha = hwy::iguana_ans::HWY_NAMESPACE;
+
+// Decompresses a block produced by hi::Compress, appending to `out`. The
+// container loop is shared with the scalar path (detail.h); only the entropy
+// stage differs: here it is the vectorized ANS32 decoder for this target.
+// Returns false on malformed input; output matches hi::DecompressScalar.
+HWY_INLINE bool Decompress(const uint8_t* HWY_RESTRICT src, size_t src_size,
+                           std::vector<uint8_t>& out) {
+  const auto decode = [](const uint8_t* payload, size_t payload_size,
+                         uint8_t* dst, size_t dst_size) {
```
jan-wassenberg: Lambdas calling SIMD code require HWY_ATTR before the opening brace.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,596 @@
...
+      MatchExtend(src, src_len, litmin, h[i], pos, &at, &ap, &al);
+      if (al > *len) {
+        *t = at;
+        *p = ap;
+        *len = al;
+      }
+    }
+  }
+
+  void BestMatchAt(int64_t litpos, int64_t pos, int64_t* tp, int64_t* mp,
```
jan-wassenberg: Longer names here, please - what's mp? Let's also add HWY_RESTRICT if pointers are non-overlapping.

hwy/contrib/iguana/iguana.h
```diff
@@ -0,0 +1,54 @@
...
+namespace hwy {
+namespace iguana {
+
+// Compresses `data` into a complete Iguana block (EncodingIguana / ANS32).
+HWY_CONTRIB_DLLEXPORT std::vector<uint8_t> Compress(const uint8_t* data,
+                                                    size_t size);
+
+// Decompresses a block produced by Compress. Returns false on malformed input.
+// The SIMD path in iguana-inl.h produces identical output.
+HWY_CONTRIB_DLLEXPORT bool DecompressScalar(const uint8_t* src, size_t src_size,
```
jan-wassenberg, line 49: Seems useful to declare here a Decompress() which is a wrapper function that does HWY_DYNAMIC_DISPATCH to the -inl.h function. This would allow callers to call without having to do dispatch themselves (similar to vqsort.h vs vqsort-inl.h).

jan-wassenberg (changes requested):
Great updates and added comments 👍 A few more finds:

tvost2:
Thanks for the careful pass — all eleven points are addressed in 64befa56.

**detail.h**

- **Payload/control overlap.** Added the `have_payload` helper and used it at all
  four sites: `len > src_size - data_cursor` is no longer the only check, the
  command must also fit in front of the control section
  (`data_cursor + len <= ctrl + 1`), so payload bytes can never be control bytes.
- **Bound by the declared length.** `CopyRaw`, `DecodeANS32` and the LZ stage now
  compare against `uncompressed_len - out.size()`. `kMaxUncompressedSize` still
  caps the declared total and each stream length.
- **End-state verification.** On the last-command marker we now require
  `out.size() == uncompressed_len` *and* `data_cursor == static_cast<size_t>(ctrl + 1)`.
- **`ent_bufs`.** Cleared after each block is decoded, so it no longer accumulates
  across `kCmdDecodeIguana` commands.
- **Export surface.** `ReadControlVarUint`, `CopyMatch` and `DecompressIguanaLZ`
  are now `HWY_INLINE` and nothing in this header is exported; only `iguana.h`
  declares public API. That also removes the ABI concern I had raised.

**iguana.cc**

- `BestMatchAt` / `BestChainMatch`: parameters renamed
  (`out_match_pos` / `out_chain_pos` / `out_len`) and marked `HWY_RESTRICT`.
- The repeat-offset path indeed needed clamping to `kMaxEncodableMatchLen`. While
  fixing it I found that this path also bypasses `MatchExtend` and therefore
  skipped `IsLegal`, so it could emit an offset that does not fit the 24-bit
  stream. It now clamps the length *and* goes through `IsLegal`.

**iguana-inl.h / iguana.h**

- `HWY_ATTR` added to the lambda (it calls `ha::Ans32Decode`).
- `hwy::iguana::Decompress` added: declared in `iguana.h`, defined in
  `iguana-inl.h` with `HWY_DYNAMIC_DISPATCH`, so callers no longer have to do the
  dispatch themselves (the `HWY_EXPORT` sits in the `#if HWY_ONCE` block outside
  the include guard, which is where it is compiled inside the last target pass).
  `iguana.h` now includes `hwy/base.h` for `HWY_INLINE` / `HWY_RESTRICT` instead of
  relying on it being pulled in transitively.

**Build**: `detail.h` added to `meson.build` and `CMakeLists.txt` (Bazel already
listed it).

**Extra coverage**: the literal-splitting path in the encoder had no test, even
though it is the only thing that protects against the truncation the varint
suffered in release builds. Added `TestLongLiteralRunSplit`: it builds a 16 MB
maximal-length LFSR sequence of degree 24 over a two-symbol alphabet, so every
24-bit window is unique, the matcher finds no match at all and the whole input is
one literal run longer than a token can carry — while the two-symbol alphabet
still makes the entropy stage win, which is what selects the LZ path. A wrong
split simply fails to round-trip, so the test does discriminate. It runs for the
static target only (the code is target-independent) and adds ~1.5 s.

Local evidence: `iguana_test` 45/45 (9 cases × AVX2 / SSE4 / SSSE3 / SSE2 /
EMU128) and `ans_test` 10/10 pass with `-Wconversion -Wsign-conversion -Werror`.

hwy/contrib/iguana/detail.h
```diff
@@ -0,0 +1,425 @@
...
+  const size_t old_size = dst.size();
+  dst.resize(old_size + match_len);
+  uint8_t* const HWY_RESTRICT out = dst.data();
+  for (size_t i = 0; i < match_len; ++i) {
+    out[old_size + i] = out[match_pos + i];
+  }
+  return true;
+}
+
+// Reads a base-128 varint backwards from src[*cursor], moving *cursor before
```
jan-wassenberg: Orphaned comment, can remove (it's also before ReadControlVarUint).

hwy/contrib/iguana/detail.h
```diff
@@ -0,0 +1,425 @@
...
+        break;
+      }
+      case kCmdDecodeIguana: {
+        const uint64_t hdr = ReadControlVarUint(src, &ctrl, &ok);
+        if (!ok) return false;
+        IguanaStream streams[kStreamCount];
+        uint64_t ulens[kStreamCount];
+        for (size_t i = 0; i < kStreamCount; ++i) {
+          ulens[i] = ReadControlVarUint(src, &ctrl, &ok);
+          if (!ok || ulens[i] > kMaxUncompressedSize) return false;
```
jan-wassenberg: I think here too we should sum up the ulens[] and compare that to uncompressed_len, not kMaxUncompressedSize, before allocating.

hwy/contrib/iguana/detail.h
```diff
@@ -0,0 +1,425 @@
...
+      }
+      if (lit_len > 0) {
+        const size_t n = static_cast<size_t>(lit_len);
+        if (n > uncompressed_len - dst.size()) return false;
+        const uint8_t* const p = literal_stream.Sequence(n, &ok);
+        if (!ok) return false;
+        dst.insert(dst.end(), p, p + n);
+      }
+      if ((token & 0x80) == 0) {
+        last_offs = -static_cast<int64_t>(off16_stream.U16(&ok));
```
jan-wassenberg, line 248: We can get into trouble with token == 0x80: lit_len == 0 && match_len == 0 - zip bomb again. We should reject those - return false.

hwy/contrib/iguana/detail.h
```diff
@@ -0,0 +1,425 @@
...
+    return data_cursor + len <= static_cast<uint64_t>(ctrl) + 1;
+  };
+  std::vector<std::vector<uint8_t>> ent_bufs;
+
+  for (;;) {
+    if (ctrl < 0) return false;
+    const uint8_t cmd = src[static_cast<size_t>(ctrl)];
+    --ctrl;
+
+    switch (cmd & kCommandMask) {
```
jan-wassenberg, line 346: Let's make sure each case produces >= 1 byte of output to prevent infinite loops. For example, we can check out.size() > prev_out_size after the switch.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,473 @@
...
+    literals.insert(literals.end(), src + litpos, src + src_len);
+  }
+};
+
+}  // namespace
+
+// ------------------------------ container
+
+HWY_CONTRIB_DLLEXPORT std::vector<uint8_t> Compress(const uint8_t* data,
+                                                    size_t size) {
```
jan-wassenberg, line 403: Let's ensure size < 2 GiB here because Insert() casts to i32 for the chain. Better yet kMaxUncompressedSize - otherwise the decoder will fail. We should have a comment or static assert, that kMaxUncompressedSize < 2 GiB.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,473 @@
...
+    Bytes ustreams[kStreamCount] = {enc.tokens,        enc.offset16,
+                                    enc.offset24,      enc.var_lit_len,
+                                    enc.var_match_len, enc.literals};
+    Bytes cstreams[kStreamCount];
+    uint64_t hdr = 0;
+    int64_t total = 0;
+    for (const auto& u : ustreams) total += static_cast<int64_t>(u.size());
+
+    for (size_t i = 0; i < kStreamCount; ++i) {
+      Bytes cs = Ans32Encode(ustreams[i].data(), ustreams[i].size());
```
jan-wassenberg, line 434: Let's assert that ustreams[i].size() < 4 GiB, otherwise we could have wraparound and div by zero in ANS.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,473 @@
...
+    if (repeat_ofs >= 0 && repeat_ofs < pos) {
+      // Unlike the chain path this does not go through MatchExtend, so the
+      // format limits have to be applied here as well: a single token carries
+      // at most kMaxEncodableMatchLen, and the offset must still fit the
+      // 24-bit stream.
+      int64_t repeat_len = Lcp(src, src_len, repeat_ofs, pos);
+      if (repeat_len > kMaxEncodableMatchLen) {
+        repeat_len = kMaxEncodableMatchLen;
+      }
+      if (IsLegal(repeat_ofs, repeat_len)) {
```
jan-wassenberg: I think this call is incorrect. pos - last_encoded_offset is a start position i.e. index in src, right? This rejects all short matches after the first 64 KiB. The other call site has pos-pos as the offset, not pos-last_encoded_offset. I think we should pass last_encoded_offset here, and possibly rename repeat_ofs to repeat_pos.

hwy/contrib/iguana/detail.h
```diff
@@ -0,0 +1,425 @@
...
+  size_t data_cursor = 0;
+  // The payload is read forwards from the start and the control bytes backwards
+  // from the end: they must not overlap, so a command may only use bytes that
+  // are still in front of the control section. `ctrl` is the last unconsumed
+  // control byte, hence the +1.
+  const auto have_payload = [&](uint64_t len) {
+    if (len > src_size - data_cursor) return false;
+    return data_cursor + len <= static_cast<uint64_t>(ctrl) + 1;
+  };
+  std::vector<std::vector<uint8_t>> ent_bufs;
```
jan-wassenberg: To prevent reallocation, let's instead have std::vector<uint8_t> ent_bufs[kStreamCount].

jan-wassenberg (changes requested):
Nice updates! I found a few remaining corner cases with the help of an internal model. Codecs are tricky :)

tvost2:
Round 3 done in dd309226 — thanks, the `0x80` token and the `IsLegal` offset were
real bugs, and pinning the int32/size limits down was worth doing.

**detail.h**

- **Commands that produce nothing are now rejected.** The loop records
  `prev_out_size` before the switch and rejects equality after it. That kills
  `token == 0x80` (and any zero-length command) without special-casing a token
  value, and it is also what makes the stream bound below sound.
- **`ulens[i]` bounded by the declared total.** One deviation, and I would like
  your call on it: I bound each stream individually (`ulens[i] > uncompressed_len`)
  rather than their sum. The sum is not bounded by `uncompressed_len` — the token,
  offset and length streams are metadata rather than output, and for compressible
  data they add up to a few times the output size (a 200 KB input yields roughly
  210 KB of streams). Individually each stream is now at most `uncompressed_len`,
  because every command must emit at least one byte, and that is what makes the
  allocation safe. I can add a sum bound as well, but it would have to be a
  constant factor rather than `uncompressed_len`.
- **`ent_bufs` is now `std::vector<uint8_t> ent_bufs[kStreamCount]`.** No
  reallocation, so appending can no longer invalidate the stream pointers.
- **Orphaned comment removed** — that was the `ReadControlVarUint` doc block left
  behind when its definition moved.

**iguana.cc**

- **`IsLegal` was being passed a position, not an offset** — you are right:
  `pos - last_encoded_offset` is an index into `src`, so every match after the
  first 64 KiB was rejected, which made the repeat offset only useful for small
  inputs. It now passes `last_encoded_offset` (the distance) and the variable is
  called `repeat_pos`.
- **`Compress` refuses `size > kMaxUncompressedSize`** and returns an empty vector.
  Any block it produced beyond that would be rejected by our own decoder, so this
  closes the asymmetry I had only documented in the previous round. Added
  `static_assert(kMaxUncompressedSize < (uint64_t{1} << 31))` for the match
  finder's `int32_t` positions.
- **ANS input size asserted** to stay below 4 GiB (`HWY_DASSERT`) before
  `Ans32Encode`, for the wraparound and divide-by-zero you mention.

Local: `iguana_test` 45/45 (9 cases × AVX2 / SSE4 / SSSE3 / SSE2 / EMU128) and
`ans_test` 10/10 with `-Wconversion -Wsign-conversion -Werror`.

hwy/contrib/iguana/iguana.cc
```diff
@@ -0,0 +1,485 @@
...
+                                    enc.offset24,      enc.var_lit_len,
+                                    enc.var_match_len, enc.literals};
+    Bytes cstreams[kStreamCount];
+    uint64_t hdr = 0;
+    int64_t total = 0;
+    for (const auto& u : ustreams) total += static_cast<int64_t>(u.size());
+
+    for (size_t i = 0; i < kStreamCount; ++i) {
+      // The ANS coder would wrap around (and divide by zero) at 4 GiB.
+      HWY_DASSERT(ustreams[i].size() < (size_t{1} << 32));
```
jan-wassenberg: Nit: size_t can be 32-bit, which would be a compiler warning/error here. Let's use `static_cast<uint64_t>(ustreams[i].size()) < (uint64_t{1} << 32)`.

hwy/contrib/iguana/detail.h
```diff
@@ -0,0 +1,431 @@
...
+
+  bool ok = true;
+  // Offset of the previous match, negated: the streams encode the distance
+  // back from the current output position. Zero means "no match seen yet", and
+  // a match that refers to it is malformed.
+  int64_t last_offs = 0;
+  while (!token_stream.IsEmpty()) {
+    int64_t match_len = 0;
+    const uint8_t token = token_stream.U8(&ok);
+    if (!ok) return false;
```
jan-wassenberg: We must also reject 0x80 (nop): `if (!ok || token == 0x80) return false;`, otherwise we can have a stream with one valid command/byte output, and lots of nops.

jan-wassenberg (changes requested):
Impressed how quickly you are able to iterate - is this Claude code and replying/acting directly on the PR comments?
Two nits remaining, one security critical and one build error/warning on 32-bit.

tvost2:
> Impressed how quickly you are able to iterate - is this Claude code and replying/acting directly on the PR comments? Two nits remaining, one security critical and one build error/warning on 32-bit.

Hahaha, you got me. But I'm not using Claude Code now, I'm trying DeepSeek. I have limited understanding of low-level programming, and I'm awake most of the day. I'm learning a lot from you and I'll fix it. It's not automation though — I just keep an eye on your replies and fix things as soon as you point them out.

tvost2:
Both fixed in c329cdcb, plus a test that covers the NOP.

**`0x80` NOP rejected** (`detail.h`). You are right that the per-command progress
check is not sufficient: a `DecodeIguana` command that emits one byte and then
pads the token stream makes progress from the container's point of view, so the
padding has to be rejected at the token itself. The check is now
`if (!ok || token == 0x80) return false;` right after reading the token.

**ANS size assertion** (`iguana.cc`). Now
`static_cast<uint64_t>(ustreams[i].size()) < (uint64_t{1} << 32)`, so it also
compiles where `size_t` is 32-bit.

**Test.** `TestSecurityMalformedLZ` now builds the same block twice: once with a
literal-only token, and once with that token followed by two NOPs. Both declare
100 bytes of output and the control is accepted, so the rejection is attributable
to the NOPs rather than to the declared length — the first version of this test
was useless because it tripped over the stream-length bound instead. The failure
path is asserted for both the scalar and the SIMD decoder.

Local: `iguana_test` 45/45 and `ans_test` 10/10 with
`-Wconversion -Wsign-conversion -Werror`.

jan-wassenberg (approved):
:) It is also interesting and informative for me, and appreciate you getting this over the line. Congrats on a nice result! We'll land it soon..

jan-wassenberg:
FYI I made some minor patches: adding HWY_INLINE_VAR, fixing unreachable return in test, updating #include.

Issue #3281, opened 2026-08-14
RVV: ReorderWidenMulAccumulate slower than generic WidenMulAccumulate


nils-david-s:
The riscv specific implementation of ReorderWidenMulAccumulate in hwy/ops/rvv-inl.h is slower than WidenMulAccumulate in hwy/ops/generic_ops-inl.h.
On Banana Pi BPI-F3 I see about a 10% to 30% throughput increase for microbenchmarks if WidenMulAccumulate is used instead of ReorderWidenMulAccumulate. 
An algorithm I am working with confirms the speedup.
The asm shows the slow down origin is in the combining before the riscv native vwmacc instruction and splitting afterwards to satisfy the split sum highway API.
Here is the asm for the two functions in question:
0000000000000000 <_ZN5N_RVV25ReorderWidenMulAccumulateEu16__rvv_uint16m1_tS0_u16__rvv_uint32m1_tRS1_>:
   0:	0d1077d7          	vsetvli	a5,zero,e32,m2,ta,ma
   4:	5e003757          	vmv.v.i	v14,0
   8:	0c1077d7          	vsetvli	a5,zero,e8,m2,ta,ma
   c:	5e003857          	vmv.v.i	v16,0
  10:	0c607757          	vsetvli	a4,zero,e8,mf4,ta,ma
  14:	cc007057          	vsetivli	zero,0,e8,m1,ta,ma
  18:	9ea03857          	vmv1r.v	v16,v10
  1c:	02856707          	vl1re32.v	v14,(a0)
  20:	0d1077d7          	vsetvli	a5,zero,e32,m2,ta,ma
  24:	9f00b657          	vmv2r.v	v12,v16
  28:	3ae74657          	vslideup.vx	v12,v14,a4
  2c:	0c807057          	vsetvli	zero,zero,e16,m1,ta,ma
  30:	f2942657          	vwmaccu.vv	v12,v8,v9
  34:	0d107057          	vsetvli	zero,zero,e32,m2,ta,ma
  38:	3ec74557          	vslidedown.vx	v10,v12,a4
  3c:	cd007057          	vsetivli	zero,0,e32,m1,ta,ma
  40:	9ec03457          	vmv1r.v	v8,v12
  44:	02850527          	vs1r.v	v10,(a0)
  48:	8082                	ret

000000000000004a <_ZN5N_RVV18WidenMulAccumulateEu16__rvv_uint16m1_tS0_u16__rvv_uint32m1_tRS1_>:
[...]

PR #3346, opened 2026-09-02
Replaced ReorderWidenMulAccumulate with WidenMulAccumulate #3281

nils-david-s:
As discussed in #3281, I replaced the RVV implementation of ReorderWidenMulAccumulate with WidenMulAccumulate. The only exception is for LMULs smaller than the minimum LMUL specified by the RISC-V specification. For these LMULs, the old implementation is still used because RearrangeToOddPlusEven does not take a D argument and therefore cannot determine the correct Lanes(d) value to emit the right vsetvli instruction for its internals.


jan-wassenberg (approved):
Looks good, thanks :)

jan-wassenberg (changes requested):
I see we are missing the CLA - are you able to sign that?

nils-david-s:
Signed it an hour ago, hope it is through by now.

jan-wassenberg (approved):

PR #3350, opened 2026-09-03
Add float decomposition ops (ilogb/logb/modf/nextafter) to hwy/contrib/math

tvost2:
Part of the `fmod, ilogb, logb, modf, nextafter, scalbn` line in
`g3doc/op_wishlist.md`.

### What's here

A new `hwy/contrib/math/float_manip-inl.h` with the `<cmath>` functions that
take a float apart or rebuild it from its exponent and significand rather than
evaluate a transcendental. All are exact (0 ULP) and work on `float32` /
`float64`, including subnormals, `±0`, `±inf` and `NaN`.

| function | semantics |
|---|---|
| `Ilogb(d, x)` | unbiased base-2 exponent of `\|x\|` as a signed-integer lane; `ilogb(0)`/`ilogb(NaN)` → lane min, `ilogb(inf)` → lane max (glibc's `FP_ILOGB0`/`FP_ILOGBNAN` convention) |
| `Logb(d, x)` | `floor(log2(\|x\|))` as a float; `logb(±0)` → `-inf`, `logb(±inf)` → `+inf`, `logb(NaN)` → `NaN` |
| `Modf(d, x, int_part)` | integer part (truncated toward zero) written to `int_part`, fractional part returned; both keep the sign of `x` |
| `NextAfter(d, from, to)` | the representable value adjacent to `from` toward `to` (== `nexttoward` for these types) |

`scalbn` / `ldexp` are already covered by `MulByPow2` in the core ops.
`fmod` is left for a follow-up — the exponent-gap reduction wants a
word-at-a-time implementation rather than the bit-at-a-time loop.

### Tests

`float_manip_test` checks each function on a grid of systematic magnitudes
(`±0`, `±min_normal`, `±denorm_min`, powers of two, `±max`, `±inf`, `NaN`)
plus 64 random bit patterns per type, and — for `NextAfter` — every pair,
bit-for-bit against `std::ilogb` / `std::logb` / `std::modf` / `std::nextafter`.
Passes on AVX2/SSE4/SSSE3/SSE2/EMU128 locally; `clang-format` and
`-Wconversion`/`-Wsign-conversion` clean.

Wired into CMake, Bazel (`BUILD`, `hwy_tests.bzl`) and Meson.


hwy/contrib/math/float_manip-inl.h
```diff
@@ -0,0 +1,300 @@
...
+}  // namespace detail
+
+// ------------------------------ Ldexp / Scalbn
+
+// std::ldexp(x, exp) == std::scalbn(x, exp) for binary floating-point:
+// returns x[i] * 2^exp[i], with `exp` a signed-integer vector. Overflow
+// saturates to +/-inf and underflow to +/-0, and the exponent is clamped to a
+// range wide enough that any finite result is exact (same as MulByPow2).
+template <class D, HWY_IF_FLOAT3264_D(D)>
+HWY_INLINE VFromD<D> Ldexp(D /*d*/, VFromD<D> x,
```
jan-wassenberg: We prefer to have longer but more readable names than the sometimes cryptic libm norm. Is there any reason to expose Ldexp/Scalbn at all, given we already have MulByPow2?

hwy/contrib/math/float_manip-inl.h
```diff
@@ -0,0 +1,300 @@
...
+template <class D, HWY_IF_FLOAT3264_D(D)>
+HWY_INLINE VFromD<RebindToSigned<D>> Ilogb(D d, VFromD<D> x) {
+  const RebindToSigned<D> di;
+  using TI = TFromD<decltype(di)>;
+
+  VFromD<decltype(di)> e = detail::RawExponent(d, x);
+  e = IfThenElse(RebindMask(di, Eq(Abs(x), Zero(d))), Set(di, LimitsMin<TI>()),
+                 e);
+  e = IfThenElse(RebindMask(di, IsInf(x)), Set(di, LimitsMax<TI>()), e);
+  e = IfThenElse(RebindMask(di, IsNaN(x)), Set(di, LimitsMin<TI>()), e);
```
jan-wassenberg: Can this be merged into the is-zero case above, given they both return LimitsMin?

hwy/contrib/math/float_manip-inl.h
```diff
@@ -0,0 +1,300 @@
...
+      RebindMask(du, up), Add(magn, Set(du, TU{1})), Sub(magn, Set(du, TU{1})));
+  VFromD<D> result = BitCast(d, Or(stepped, sbit));
+
+  // from == 0: the neighbour is the smallest subnormal with the sign of `to`.
+  const VFromD<D> tiny =
+      BitCast(d, Or(Set(du, TU{1}), And(BitCast(du, to), kSign)));
+  result = IfThenElse(Eq(from, Zero(d)), tiny, result);
+
+  result = IfThenElse(Eq(from, to), to, result);
+  result = IfThenElse(Or(IsNaN(from), IsNaN(to)), NaN(d), result);
```
jan-wassenberg: Can fuse into IsEitherNaN(from, to).

hwy/contrib/math/float_manip-inl.h
```diff
@@ -0,0 +1,300 @@
...
+// std::fmod(x, y): the exact floating-point remainder of x / y with the sign of
+// x, i.e. x - n*y where n is x/y truncated toward zero. Computed from the
+// integer significands so it is exact for every input (no cancellation from a
+// rounded quotient). fmod(x, 0), fmod(+/-inf, y) and fmod(x, NaN) are NaN;
+// fmod(x, +/-inf) is x for finite x.
+//
+// The core reduction runs once per bit of the exponent gap ilogb(x)-ilogb(y),
+// masked per lane; inputs whose magnitudes are within a few powers of two
+// finish in a handful of iterations, but a very large ratio (e.g. DBL_MAX
+// mod DBL_TRUE_MIN) is correspondingly slow.
```
jan-wassenberg: hm, I'd rather not provide a function that can be super slow. Let's either implement word at a time, or range reduction if possible.

hwy/contrib/math/float_manip_test.cc
```diff
@@ -0,0 +1,240 @@
...
+// clang-format on
+
+HWY_BEFORE_NAMESPACE();
+namespace hwy {
+namespace HWY_NAMESPACE {
+namespace {
+
+template <typename T>
+bool BitEqual(T a, T b) {
+  if (std::isnan(a) && std::isnan(b)) return true;
```
jan-wassenberg: Prefer ScalarIsNaN to minimize stdlib usage.

jan-wassenberg (changes requested):
Nice work :) One remaining concern about FMod; if too complex, we can leave it for a follow-up.

tvost2:
Thanks. Addressed the four smaller comments, and dropped Fmod to a follow-up per your suggestion -- the word-at-a-time reduction deserves its own change. PR is now Ilogb / Logb / Modf / NextAfter.

hwy/contrib/math/float_manip-inl.h
tvost2: Removed Ldexp and Scalbn entirely; left a one-line note pointing at MulByPow2.

hwy/contrib/math/float_manip-inl.h
tvost2: Merged -- `Or(Eq(Abs(x), Zero(d)), IsNaN(x))` in a single IfThenElse.

hwy/contrib/math/float_manip-inl.h
tvost2: Done.

hwy/contrib/math/float_manip-inl.h
tvost2: Agreed -- removed Fmod from this PR. I'll redo it word-at-a-time (per-lane long division clearing ~ExponentBits per step instead of one bit) as a follow-up.

hwy/contrib/math/float_manip_test.cc
tvost2: Done -- ScalarIsNaN in BitEqual, and ScalarIsNaN / ScalarIsInf for the ilogb special-case check. Also dropped a leftover debug fprintf and <stdio.h>.

jan-wassenberg (approved):
Nice, thanks for updating.

PR #3360, opened 2026-09-07
example: add CRC-64/XZ example

### LINKED ISSUE #3250: Add CRC example

tejas5038:
Fixes #3250

Adds a small CRC-64/XZ example using Highway’s existing portable SIMD CRC implementation. The example validates the standard CRC-64/XZ check value for `123456789` and demonstrates a direct usage pattern for the contrib CRC API.


hwy/examples/crc64.cc
```diff
@@ -0,0 +1,78 @@
...
+  constexpr char kMessage[] = "123456789";
+  constexpr uint64_t kExpected = 0x995DC9BBDF1939FAull;
+
+  const uint64_t actual = ComputeCrc64(
+      reinterpret_cast<const uint8_t*>(kMessage), sizeof(kMessage) - 1);
+
+  printf("CRC-64/XZ(\"%s\") = 0x%016llX\n", kMessage,
+         static_cast<unsigned long long>(actual));
+
+  if (actual != kExpected) {
```
jan-wassenberg: Recommend using `HWY_ASSERT_EQ(kExpected, actual) from test_util-inl.h; we can also then remove the print of actual above, which is non-portable.

jan-wassenberg (changes requested):
Thanks for adding. Just one nit:

jan-wassenberg (approved):
👍

PR #3362, opened 2026-09-08
algo_bench.cc improvements. Add BenchCopyIf

a-eee:
Now operator() runs the measured function only once, and Bench handles data generation and iteration over inner_reps. Also added a way to measure functions with multiple buffers by specifying kNumBuffers. For functions like CopyIf that are not benchmarked at 8bit I add kBench8 = false. Added struct BenchSummary instead of L6_cnt and L6_acc, and now also prints GeoMean. With the kPrintOnlySummary flag prints only L6 and GeoMean

### DIFF
--- a/hwy/contrib/algo/algo_bench.cc
+++ b/hwy/contrib/algo/algo_bench.cc
@@ -19,6 +19,7 @@
 #include <vector>
 #include <cmath>
 #include <algorithm>
+#include <array>
 
 // clang-format off
 #undef HWY_TARGET_INCLUDE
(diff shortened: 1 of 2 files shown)


hwy/contrib/algo/algo_bench.cc
```diff
@@ -117,63 +123,97 @@ struct Ctx {
...
 }
 
 template <class Func, class Gen>
-HWY_NOINLINE void BenchAllTypes(Func func, Gen gen, size_t size, double& L6_acc, size_t& L6_cnt) {
-  Bench<uint8_t>(func, gen, size, L6_acc, L6_cnt);
-  Bench<int16_t>(func, gen, size, L6_acc, L6_cnt);
-  Bench<int32_t>(func, gen, size, L6_acc, L6_cnt);
-  Bench<int64_t>(func, gen, size, L6_acc, L6_cnt);
+HWY_NOINLINE void BenchAllTypes(Func func, Gen gen, size_t size, BenchSummary& summary) {
+  if constexpr (func.kBench8) {
```
jan-wassenberg, line 211: (No action required, YAGNI for now: but we could consider a bitfield of enabled sizes. HWY_HAVE_INTEGER64 is currently always set)

jan-wassenberg (approved):

PR #3365, opened 2026-09-08
Add 52-bit multiply-add operations

tvost2:
Implements MulAdd52Lo and MulAdd52Hi for uint64 lanes. Adds the generic fallback, AVX-512 IFMA intrinsics, IFMA target detection and CPUID gating, partial-vector tests, and quick reference documentation. Focused MulAdd52 tests pass for AVX2, SSE4, SSSE3, SSE2, and EMU128.


hwy/detect_targets.h
```diff
@@ -21,6 +21,17 @@
 
 #include "hwy/detect_compiler_arch.h"
 
+// Whether the current compilation target has native 52-bit multiply-add
+// instructions. MulAdd52Lo/Hi are still available through the generic
+// fallback when this is zero.
+#ifndef HWY_HAVE_MULADD52
+#if HWY_ARCH_X86 && (defined(__AVX512IFMA__) || defined(__AVX10_2__))
```
Andarwinux: avx10.2 implies ifma.

g3doc/quick_reference.md
```diff
@@ -812,6 +812,19 @@ All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
     <code>V **MulHigh**(V a, V b)</code>: returns the upper half of `a[i] *
     b[i]` in each lane.
 
+*   `V`: `u64` \
+    <code>V **MulAdd52Lo**(V c, V a, V b)</code>: adds the low 52 bits of
+    `a[i] * b[i]` to the low 52 bits of `c[i]`. The upper 12 bits of `c[i]`
+    are preserved; inputs are interpreted modulo `2^52`.
```
jan-wassenberg: Please match the argument order of our MulAdd(f,m,a): f*m + a. Let's also define the op in a way that other platforms could also implement. SVE has a 64-bit MulHi; we could require that inputs are < 2^52, then this would work on both. We can say results are implementation-defined if the input >= 2^52.

(Also, I think ifma zero-extends the 52-bit result before adding to the u64 addend)

g3doc/quick_reference.md
```diff
@@ -812,6 +812,19 @@ All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
...
+
+*   `V`: `u64` \
+    <code>V **MulAdd52Hi**(V c, V a, V b)</code>: adds bits 52 through 103 of
+    `a[i] * b[i]` to the low 52 bits of `c[i]`. The upper 12 bits of `c[i]`
+    are preserved; inputs are interpreted modulo `2^52`.
+
+    `HWY_HAVE_MULADD52` is 1 when the current compilation target has native
```
jan-wassenberg: HWY_HAVE is used for features that are unavailable and must not be used when 0. Let's name this `HWY_NATIVE_*`, similar to `HWY_NATIVE_FMA`.

hwy/ops/generic_ops-inl.h
```diff
@@ -36,6 +36,53 @@ HWY_BEFORE_NAMESPACE();
...
+  const auto mask52 = Set(d, 0x000FFFFFFFFFFFFFULL);
+  const auto mask26 = Set(d, 0x0000000003FFFFFFULL);
+  a = And(a, mask52);
+  b = And(b, mask52);
+  const auto a0 = And(a, mask26);
+  const auto b0 = And(b, mask26);
+  const auto a1 = ShiftRight<26>(a);
+  const auto b1 = ShiftRight<26>(b);
+  const auto p0 = Mul(a0, b0);
+  const auto p1 = Add(Mul(a1, b0), Mul(a0, b1));
```
jan-wassenberg: SVE has integer MulAdd which we could use here.

hwy/detect_targets.h
jan-wassenberg: Let's define this in set_macros-inl.h. Later we can expand the definition to include target=SVE as a sufficient condition, but target is unknown in this detect_targets.h (target independent).

jan-wassenberg (changes requested):
Thanks for adding!

g3doc/quick_reference.md
```diff
@@ -812,6 +812,20 @@ All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
     <code>V **MulHigh**(V a, V b)</code>: returns the upper half of `a[i] *
     b[i]` in each lane.
 
+*   `V`: `u64` \
+    <code>V **MulAdd52Lo**(V a, V b, V c)</code>: computes `a[i] * b[i]` in
```
jan-wassenberg, line 816: Let's think from the perspective of targets that have native 64-bit mul. "inputs are interpreted modulo 2^52" would add extra work. Wouldn't it be better to require that inputs be < 2^52 (callers can often ensure this), else results are implementation-defined?
Also, I think "The upper 12 bits of c[i] are preserved" is not helpful and could be removed. The sum could actually carry into the 12th highest bit, hence this is at best confusing.

g3doc/quick_reference.md
```diff
@@ -812,6 +812,20 @@ All other ops in this section are only available if `HWY_TARGET != HWY_SCALAR`:
...
+    52-bit arithmetic and adds the low 52 bits to `c[i]`. The upper 12 bits
+    of `c[i]` are preserved; inputs are interpreted modulo `2^52`.
+
+*   `V`: `u64` \
+    <code>V **MulAdd52Hi**(V a, V b, V c)</code>: computes bits 52 through
+    103 of `a[i] * b[i]` and adds them to the low 52 bits of `c[i]`. The
```
jan-wassenberg: Here also we could delete "low 52 bits of". We're just adding to c[i], right?
We can also delete the last sentence (about upper 12 bits and modulo).

hwy/ops/generic_ops-inl.h
```diff
@@ -36,6 +36,53 @@ HWY_BEFORE_NAMESPACE();
...
```
jan-wassenberg: How about fusing into MulAdd? `MulAdd(a1, b0, Mul(a0, b1));`

hwy/ops/generic_ops-inl.h
```diff
@@ -36,6 +36,53 @@ HWY_BEFORE_NAMESPACE();
...
+  const auto low = Add(p0, ShiftLeft<26>(And(p1, mask26)));
+  return Or(And(c, Not(mask52)), And(Add(low, c), mask52));
```
jan-wassenberg: How about we just And(low, mask52) and add that to c?

hwy/tests/mul_test.cc
```diff
@@ -112,6 +112,44 @@ struct TestMulOverflow {
   }
 };
 
+struct TestMulAdd52 {
+  template <typename T, class D>
+  HWY_NOINLINE void operator()(T /*unused*/, D d) {
+    static_assert(std::is_same<T, uint64_t>::value, "requires uint64_t");
```
jan-wassenberg: Prefer IsSame<T, uint64_t>().

jan-wassenberg (changes requested):
Some more comments on the semantics :)

hwy/ops/generic_ops-inl.h
```diff
@@ -36,6 +36,53 @@ HWY_BEFORE_NAMESPACE();
...
+#undef HWY_NATIVE_MULADD52_IMPL
+#else
+#define HWY_NATIVE_MULADD52_IMPL
+#endif
+
+template <class V, HWY_IF_U64_D(DFromV<V>)>
+HWY_API V MulAdd52Lo(V a, V b, V c) {
+  const auto d = DFromV<V>();
+  const auto mask52 = Set(d, 0x000FFFFFFFFFFFFFULL);
+  const auto mask26 = Set(d, 0x0000000003FFFFFFULL);
```
jan-wassenberg: Now that we have simplified the API contract, Lo() can be implemented as just Mul, And(mask52), Add.

hwy/detect_targets.h
```diff
@@ -658,7 +658,7 @@
     ((defined(__AVX512VNNI__) && defined(__VAES__) &&             \
       defined(__VPCLMULQDQ__) && defined(__AVX512VBMI__) &&       \
       defined(__AVX512VBMI2__) && defined(__AVX512VPOPCNTDQ__) && \
-      defined(__AVX512BITALG__)) ||                               \
+    defined(__AVX512BITALG__) && defined(__AVX512IFMA__)) ||    \
```
Andarwinux, line 661: also update AVX3_DL in https://github.com/google/highway#targets?

hwy/ops/generic_ops-inl.h
```diff
@@ -36,6 +36,53 @@ HWY_BEFORE_NAMESPACE();
...
+  const auto p0 = Mul(a0, b0);
+  const auto p1 = MulAdd(a1, b0, Mul(a0, b1));
+  const auto low = Add(p0, ShiftLeft<26>(And(p1, mask26)));
+  return Add(And(low, mask52), c);
+}
+
+template <class V, HWY_IF_U64_D(DFromV<V>)>
+HWY_API V MulAdd52Hi(V a, V b, V c) {
+  const auto d = DFromV<V>();
+  const auto mask52 = Set(d, 0x000FFFFFFFFFFFFFULL);
```
jan-wassenberg: I think we can implement this as just
```
  const V lo = ShiftRight<52>(Mul(a, b));
  const V hi = ShiftLeft<12>(MulHigh(a, b));
  return Add(Or(lo, hi), c);
```

That would be good on platforms with native 64-bit Mul and especially MulHigh (SVE, RVV, PPC10, LSX). I can add a HWY_NATIVE_MUL64 to detect this.

Otherwise:
It looks like your code wants to use 32-bit ops (26 bit limbs), which would make sense on other targets, but then we should cast to 32-bit before multiplying, right?
And as above, no need to mask `a` and `b` beforehand.

jan-wassenberg (changes requested):
Sorry to have multiple rounds here :) Now that we have simplified the API contract, the emulated implementation can be much simpler as well.

hwy/detect_targets.h
jan-wassenberg: +1, good catch, thanks :)

README.md
```diff
@@ -229,8 +229,8 @@ Highway supports 27 targets, listed in alphabetical order of platform:
     -   `SSE4` (~Nehalem, also includes AES + CLMUL).
     -   `AVX2` (~Haswell, also includes BMI2 + F16 + FMA)
     -   `AVX3` (~Skylake, AVX-512F/BW/CD/DQ/VL)
-    -   `AVX3_DL` (~Icelake, includes `BitAlg` + `CLMUL` + `GFNI` + `VAES` +
-        `VBMI` + `VBMI2` + `VNNI` + `VPOPCNT`),
+    -   `AVX3_DL` (~Icelake, includes `BitAlg` + `CLMUL` + `GFNI` + `IFMA` +
```
Andarwinux: This is ambiguous: Is this AVX512IFMA or AVXIFMA?

README.md
```diff
@@ -229,8 +229,8 @@ Highway supports 27 targets, listed in alphabetical order of platform:
...
+        `VAES` + `VBMI` + `VBMI2` + `VNNI` + `VPOPCNT`),
```
Andarwinux: @jan-wassenberg Is the VNNI here also ambiguous?

README.md
tvost2: Right - it is the EVEX-encoded AVX512IFMA flavor that Ice Lake adds, not the VEX-encoded AVX-IFMA (which came later, with AVX10.2). I have spelled it out as `AVX512IFMA` in the list.

README.md
tvost2: Good point - AVX-VNNI is also a separate, VEX-encoded feature, so I spelled this one out too as `AVX512VNNI` (the flavor Ice Lake adds).

The other short names (`BitAlg`, `GFNI`, `VAES`, `VBMI`, `VBMI2`, `VPOPCNT`) are unique to AVX-512/AVX10, but `CLMUL` here means `VPCLMULQDQ` since Icelake adds the EVEX variant. Happy to spell all of them out if you would prefer that.

jan-wassenberg (approved):
Nice work and thanks for updating the readme :)

jan-wassenberg:
avx512vlifmaintrin.h doesn't exist here, apparently it's avxifmaintrin.h. I'm patching this in.

PR #3369, opened 2026-09-10
feat(detect_targets): enable HWY_WASM by default on WebAssembly (fixes #2831)

### LINKED ISSUE #2831: `hwy::DispatchedTarget` always returns HWY_EMU128 on WASM/Emscripten

jdymitarai:
### Problem
As reported in #2831, `hwy::DispatchedTarget` previously returned `HWY_EMU128` on WebAssembly unless the `-msimd128` compiler flag was passed explicitly, because `HWY_BASELINE_WASM` checked `defined(__wasm_simd128__)`.

### Solution
Enable `HWY_WASM` by default on `HWY_ARCH_WASM` without requiring `-msimd128`, as discussed in #2831. WASM SIMD (simd128) is now widespread across modern web browsers and runtimes.

Developers building a non-SIMD fallback WebAssembly binary can cleanly opt out by specifying `HWY_DISABLED_TARGETS=HWY_WASM` (or `HWY_COMPILE_ONLY_SCALAR`), which falls back to `HWY_EMU128` / `HWY_SCALAR`.

Fixes #2831.

### DIFF
--- a/hwy/contrib/sort/sort_test.cc
+++ b/hwy/contrib/sort/sort_test.cc
@@ -364,8 +364,6 @@ template <typename T, class Order>
 void TestSelectWithNaNForType(Order order) {
   const size_t num = AdjustedReps(40 * 1000);
   if (num < 32) return;
-  constexpr bool asc = hwy::IsSame<Order, hwy::SortAscending>();
-
   size_t num_nan;
   const std::vector<T> input = MakeNaNInfInput<T>(num, 123456789, num_nan);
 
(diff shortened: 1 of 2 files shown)


jan-wassenberg:
Unfortunately this is nontrivial to land, we have some internal projects which turn out to break when enabling SIMD by default, which is why #2851 was rolled back.

How much of an issue is it for you that -msimd128 is required to be passed?

PR #3370, opened 2026-09-10
Add AllOf/AnyOf/NoneOf to contrib/algo

Wint3rNight:
Implements three of the "Remaining STL functions for hwy/contrib/algo" from g3doc/op_wishlist.md: AllOf, AnyOf and NoneOf, matching std::all_of, std::any_of and std::none_of.

They go in find-inl.h next to FindIf, which they are closest to. Three notes on the implementation.

Not wrappers over FindIf. AnyOf could be FindIf(...) != count, but FindIf calls FindFirstTrue to compute where the match is and none of these three need that. Built directly on AllFalse/AllTrue, a loop iteration is the loads, the predicate and one mask test.

Unrolled 4x rather than the 2x in Find/FindIf, matching AllUnique, the other early-exit scan in this file. There is no loop-carried dependency here.

The tail is the part worth reviewing. LoadN zero-fills the lanes past the end, and a predicate like x > -5 is true for those zeros, so they have to be excluded. AnyOf masks with FirstN. AllOf uses AndNot(pred, FirstN), which is the set of lanes holding a real element that failed, so it never has to force the padding lanes true.

Tests compare against the STL over 512 random counts, three misalignments, all types and 20 predicate thresholds, plus predicates pinned always-true and always-false so unsigned types still exercise the AllOf-true path. I checked that the tests actually bite by breaking each tail mask: dropping the AnyOf mask gives "i8x32 count 2 val -5: any 1 want 0", and dropping the AllOf one gives "count 1 val 0: all 0 want 1".

Full build and ctest are green locally, 2475 of 2475.

Also strikes IndexOfMin/Max off the wishlist, since #3253 added it. I can drop that line if you would rather keep it separate.

### DIFF
--- a/g3doc/op_wishlist.md
+++ b/g3doc/op_wishlist.md
@@ -37,8 +37,6 @@ fmod, ilogb, logb, modf, nextafter, nexttoward, scalbn
 
 ### Remaining STL functions for hwy/contrib/algo
 
-*   IndexOfMin/Max
-*   AllOf / AnyOf / NoneOf
 *   EqualSpan
 *   ReverseSpan
 *   ShuffleSpan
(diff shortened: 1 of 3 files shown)


jan-wassenberg (approved):
Nice, thanks for adding!

PR #3366, opened 2026-09-10
fix(sort): capture asc in std::sort lambda in TestSelectWithNaNForType

jdymitarai:
### Summary
This PR resolves two MSVC compilation errors affecting Highway:

1. **`sort_test.cc` error C3493**:
   In `TestSelectWithNaNForType`, the custom comparator lambda for `std::sort` references the local `constexpr bool asc`, but used an empty capture `[]`. On MSVC conforming mode, this triggers:
   ```
   error C3493: 'asc' cannot be implicitly captured because no default capture mode has been specified
   ```
   Fix: Explicitly capture `[asc]`.

2. **`sorting_networks-inl.h` and `x86_128-inl.h` error C2039**:
   In `hwy/contrib/sort/sorting_networks-inl.h`, the `#else` (`!VQSORT_ENABLED`) block was wrapped in a nested `namespace detail { ... }` while already inside `namespace detail`. This created a nested `hwy::N_SSSE3::detail::detail` namespace.
   Inside `hwy/ops/x86_128-inl.h` (and `loongarch_lsx-inl.h`, `ppc_vsx-inl.h`), `CompressBits` and `CompressNotBits` inside `namespace detail` called `detail::IndicesFromBits128(...)`. Under MSVC two-phase name lookup, `detail::` resolved to `detail::detail` and failed with:
   ```
   error C2039: 'IndicesFromBits128': is not a member of 'hwy::N_SSSE3::detail::detail'
   ```
   Fix: Remove the redundant nested `namespace detail` in `sorting_networks-inl.h` and remove the redundant `detail::` qualifiers on `IndicesFromBits128` / `IndicesFromNotBits128` inside `namespace detail`.

### Testing
- Verified MSVC build (`Release` config) of `hwy_contrib`, `sort_test`, and `copy_test`.
- Ran `sort_test.exe` (35 test suites) and `copy_test.exe` across AVX2, SSE4, SSSE3, SSE2, EMU128 - all passed with 0 errors.

PR #3368, opened 2026-09-10
fix(sort): check VQSortHaveFloat16/64 in TestFloatInf (fixes #2259)

jdymitarai:
Fixes #2259.

### Problem

In \SortTestGroup/SortTest.TestAllFloatInf/EMU128\, running on 32-bit ARM (such as Debian armhf) causes an abort at \qsort_f64a.cc:36\ (\HWY_ASSERT(0)\).

The root cause is that \TestAllFloatInf\ iterates over float types using \ForFloatTypesDynamic(ForPartialVectors<TestFloatInf>())\. When testing the \EMU128\ fallback target:
1. \N_EMU128\ has \HWY_HAVE_FLOAT64 == 1\, and \hwy::HaveFloat64()\ returns true on platforms with hardware floating point (e.g. VFPv3/v4).
2. \TestFloatInf::operator()\ is invoked with \T = double\.
3. It calls \hwy::VQSort(in.get(), num, SortAscending())\, which uses \HWY_DYNAMIC_DISPATCH\.
4. \HWY_DYNAMIC_DISPATCH\ selects the best available SIMD instruction set on the CPU, which is \HWY_NEON\ on ARMv7.
5. On 32-bit ARM NEON, 64-bit vector float instructions do not exist (\HWY_HAVE_FLOAT64 == 0\). Hence, \N_NEON::SortF64Asc\ hits \HWY_ASSERT(0)\.

Commit 4f846162 introduced \hwy::VQSortHaveFloat16()\ and \hwy::VQSortHaveFloat64()\ to resolve similar dispatch mismatches in \sort_test.cc\, but \sort_unit_test.cc\'s \TestFloatInf\ was not updated to guard its \VQSort\ calls.

### Solution

In \TestFloatInf::operator()\, check \hwy::VQSortHaveFloat64()\ when \T == double\ and \hwy::VQSortHaveFloat16()\ when \T == float16_t\ before calling \VQSort\, skipping the test if the dynamic dispatch target does not support that float type.

PR #3371, opened 2026-09-11
fix(sort): use a default capture in TestSelectWithNaNForType comparator

Wint3rNight:
`Build and test Clang-22` is currently failing on master, so it shows up red on every open pull request.

The cause is in `TestSelectWithNaNForType`. #3366 changed the comparator lambda from `[]` to `[asc]` so MSVC would stop reporting C3493, which asks for a default capture mode. That fixed MSVC, but `asc` is `constexpr`, so reading it inside the lambda is not an odr-use and the explicit capture is redundant. Clang 22 then rejects it:

```
hwy/contrib/sort/sort_test.cc:379:38: error: lambda capture 'asc' is not required to be
captured for this use [-Werror,-Wunused-lambda-capture]
```

`[&]` satisfies both: MSVC gets the default capture mode C3493 asks for, and Clang has no explicit capture left to call unnecessary. Neither form actually captures anything, since the use is not an odr-use, so there is no behaviour change.

This was easy to miss because the `Build / test` workflow runs on `pull_request` and a weekly schedule, not on pushes to master, so a regression only becomes visible once somebody opens the next PR. The Saturday cron would have caught it.

**Verification.** With Clang 22.1.8, the same version CI uses, and the warning set from CMakeLists plus `-Werror`: master reports the error above, this branch reports none. GCC is clean, and `sort_test` passes 35 of 35 including `TestSelectWithNaN`. I have no MSVC available to confirm that side directly, so if C3493 somehow still fires there, `[=]` is the other form that works and I am glad to switch.

Line 379 is the only explicit-capture lambda in the file, so this should be all of it.

### DIFF
--- a/hwy/contrib/sort/sort_test.cc
+++ b/hwy/contrib/sort/sort_test.cc
@@ -376,7 +376,7 @@ void TestSelectWithNaNForType(Order order) {
     if (!ScalarIsNaN(x)) nonnan_in.push_back(x);
   }
   std::sort(nonnan_in.begin(), nonnan_in.end());
-  std::sort(ref.begin(), ref.end(), [asc](float a, float b) {
+  std::sort(ref.begin(), ref.end(), [&](float a, float b) {
     if (ScalarIsNaN(a)) return false;  // NaN sorts to the back
     if (ScalarIsNaN(b)) return true;
     return asc ? (a < b) : (a > b);


jan-wassenberg:
Thanks for reporting. #3368 includes a fix for this, let's go with that one.

Issue #3374, opened 2026-09-11
contrib/algo: questions on Reduce and ShuffleSpan before implementing


Wint3rNight:
I've just finished AllOf/AnyOf/NoneOf (#3370) and I'd like to keep going
through the "Remaining STL functions for hwy/contrib/algo" list. EqualSpan
and ReverseSpan look unambiguous and I'm working on those now. Reduce and
ShuffleSpan both have choices in them that I'd rather have your answer on
than guess at, since either one changes the whole shape of the function.

Reduce

1. Does the functor take vectors, like Transform's func(d, v), or scalars?
   If it takes vectors, combining the lanes at the end needs a rule too, so
   either the same functor has to be applied across lanes (which means
   requiring it to be associative and commutative) or the caller passes a
   second scalar combiner. I'd rather not invent that API myself.

2. Is the initial value a parameter, like std::reduce's init, or is Reduce
   only defined for count != 0?

3. Should the result be bit-identical across targets? For float addition it
   won't be unless the summation order is pinned, and pinning it costs most
   of the speedup. ReduceSum has the same property today, so I suspect the
   answer is no, but it decides what the test can assert.

My instinct: one vector functor, documented as requiring associativity and
commutativity, an explicit init parameter, and no cross-target bit-identity
guarantee.

ShuffleSpan

1. Which RNG? contrib/algo doesn't depend on contrib/random today. Is it
   fine to add that dependency, or should the caller pass in a generator or
   a seed?

2. Same seed, same permutation on every target? That one decides the
   algorithm: a Fisher-Yates that's reproducible across vector lengths is
   basically scalar, while the fast vectorised approaches produce a
   different permutation per target.

3. Does it need to be a uniform permutation, or is an approximate shuffle
   acceptable? Unbiased index generation costs a rejection loop.

My instinct here is much weaker than on Reduce, so I'm happy to go whichever
way you prefer, or to leave ShuffleSpan alone if it isn't worth the
dependency.

jan-wassenberg:
Thanks, indeed good to discuss beforehand :)

Let's pause Reduce for a while - it's not clear to me which semantics applications require. Generally it's good to implement with a use-case in mind which we don't yet have.

For Shuffle: I am happy with RngStream from random-inl.h. I'd suggest supporting both that, and user-provided generators, as in std::shuffle. But we must avoid a dep because it would be circular (via VQSort).

[...]

PR #3376, opened 2026-09-11
Add EqualSpan and ReverseSpan to hwy/contrib/algo

Wint3rNight:
Implements `EqualSpan` and `ReverseSpan` from the "Remaining STL functions for
hwy/contrib/algo" list in op_wishlist.md, following #3370. That leaves
`ShuffleSpan` and `Reduce`, which I have asked about in #3374 rather than
guessing at the API.

### EqualSpan (find-inl.h)

Returns whether `a[0, count)` and `b[0, count)` are equal element-wise, like
std::equal, stopping at the first mismatch.

The tail takes no FirstN mask, unlike AnyOf. LoadN zero-fills the lanes past
`remaining` on both sides, so those lanes hold the same value and compare
equal. A mask there would be dead work.

Argument order is `(d, a, b, count)` rather than the `(from, count, to)` used
by Copy/CopyIf. Copy has a direction, so `count` attaches to the source; here
the two spans are symmetric and share one count. `Dot::Compute(d, pa, pb, num)`
in contrib/dot is the closest precedent and does the same. Happy to change it
if you prefer the Copy shape.

It unrolls 4x like AnyOf rather than 2x like Find, since the position of the
mismatch is never needed.

NaN follows std::equal and `operator==`, so a NaN is not equal to itself and
spans holding a NaN at the same index are reported unequal. That is documented
on the function.

### ReverseSpan (copy-inl.h)

Reverses a span in place, like std::reverse. Each iteration takes one vector
from each end, reverses it and stores it at the opposite end. Three cases:

- the main loop, while `lo + 2 * N <= hi`
- `N <= remaining < 2 * N`, where the two vectors overlap. Both are loaded
  before either is stored, and the overlapping lanes are written twice with
  the second write being the correct one
- fewer than N elements left, handled with LoadN, Reverse and
  SlideDownLanes(N - remaining) so that no memory past the end is read

I kept the overlapping pair as its own block rather than folding it into the
loop with a second exit, because the three cases read more clearly separated.
Glad to restructure if you would rather see one copy of the body.

### Verification

`find_test` 25 of 25 and `copy_test` 20 of 20 on AVX2, SSE4, SSSE3, SSE2 and
EMU128. Compiled clean for AVX3, AVX3_DL, AVX3_ZEN4, AVX3_SPR and AVX10_2, and
separately for HWY_SCALAR. The two -inl.h headers also compile under C++11 with
HWY_COMPILE_ALL_ATTAINABLE. Zero warnings from both GCC 16 and Clang 22 under
the warning set in CMakeLists, including -Wconversion and -Wsign-conversion.

The tests sweep every count up to `4 * Lanes(d)` with three misalignments and
all types. ReverseSpan additionally checks a sentinel one past the end, so an
over-store is caught, and checks that reversing twice restores the input.

I broke each implementation on purpose to confirm the tests would notice:

| Mutation | Caught at |
| --- | --- |
| EqualSpan tail check removed | count 324, differing element at 320 |
| EqualSpan 4th vector dropped from the Or | count 450, differing element at 225 |
| ReverseSpan made a no-op | count 2 |
| ReverseSpan short-tail branch skipped | count 2 |
| ReverseSpan overlap branch skipped | count 33 |

I have not run NEON, SVE, RVV, WASM or the big-endian targets, only compiled
what this x86 box can reach.

One note on the test for ReverseSpan: it builds its reference by hand instead
of calling std::reverse on a std::vector. GCC 16 cannot bound `count` through
that inlining and emits a spurious -Wstringop-overflow for every instantiation,
418 of them in my first version.

### DIFF
--- a/g3doc/op_wishlist.md
+++ b/g3doc/op_wishlist.md
@@ -37,8 +37,6 @@ fmod, ilogb, logb, modf, nextafter, nexttoward, scalbn
 
 ### Remaining STL functions for hwy/contrib/algo
 
-*   EqualSpan
-*   ReverseSpan
 *   ShuffleSpan
 *   Reduce
 
@@ -190,3 +188,5 @@ For SVE (svld1sb_u32)+WASM? Compiler can probably already fuse.
 *   ~~Document Reduce/Min NaN behavior~~
 *   ~~IndexOfMin/Max~~ (algo)
 *   ~~AllOf / AnyOf / NoneOf~~ (algo)
+*   ~~EqualSpan~~ (algo)
+*   ~~ReverseSpan~~ (algo)
--- a/hwy/contrib/algo/copy-inl.h
+++ b/hwy/contrib/algo/copy-inl.h
@@ -137,6 +137,49 @@ T* CopyIf(D d, const T* HWY_RESTRICT from, size_t count, T* HWY_RESTRICT to,
   return to;
 }
 
+// Reverses `inout[0, count)` in place, like std::reverse. `count` may be zero.
+template <class D, typename T = TFromD<D>>
+void ReverseSpan(D d, T* HWY_RESTRICT inout, size_t count) {
+  const size_t N = Lanes(d);
+
+  // [lo, hi) is the region not yet reversed. Each iteration takes one vector
+  // from each end, reverses it, and stores it at the opposite end.
+  size_t lo = 0;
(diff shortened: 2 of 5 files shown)


jan-wassenberg (approved):
Nice, thanks for adding :)

PR #3377, opened 2026-09-11
Added Partition op

johnplatts:
Resolves one of the issues raised in issue #3142.

`Partition(d, v, m)` returns the lanes of `v` for which `m[i]` is true, followed by the lanes of `v` for which `m[i]` is false.

There is a use case for updating StoreLeftRight in hwy/contrib/sort/vqsort-inl.h to possibly use Partition.


jan-wassenberg (commented):
👍

PR #3382, opened 2026-09-13
RVV: Fix ReorderWidenMulAccumulate for fractional LMUL

### LINKED ISSUE #3305: RVV: Bug: RearrangeToOddPlusEven
The function RearrangeToOddPlusEven in hwy/ops/rvv-inl,h has UB that is not covered by the current test suit. 
Below is a modified hwy/tests/widen_mul_test.cc (saved as .txt because cc is not allowed) to surface the behaviour. 

The output of the function does not solely depend on its inputs, but on the tail state of the sum0 register from previous operations. To trigger the bug, I took the result of a ReorderWidenMulAccumulate, saved it to memory, then loaded it into a new register and used the loaded vector as input to the function. 

The output now diverges from the expected result even though an assert about the inputs show there is no difference. 

[widen_mul_test.txt](https://github.com/user-attachments/files/31296125/widen_mul_test.txt)

tinfengyu:
Fixes #3305
Addresses #3281

Summary
Fix ReorderWidenMulAccumulate on RVV for fractional LMUL.

For real e32,mf2, the native vector type is also used to represent virtual e32,mf4. The previous implementation could leave state required by RearrangeToOddPlusEven in inactive tail lanes. After a Store/Load round trip, those tail lanes are no longer guaranteed to retain the required values, which can change the result.

This change keeps all persistent state required by RearrangeToOddPlusEven in active lanes.

For the real e32,mf2 case, the widening products are first computed in natural order:
[P0 | P1 | P2 | P3]
They are then stored using the block-transposed representation:
sum0 = [P0 | P2]
sum1 = [P1 | P3]
The implementation uses native vwmul.vv / vwmulu.vv followed by vcompress to construct the two accumulators.

This also avoids the slide-based rearrangement considered for this path and addresses the RVV performance concern in #3281.

Testing
Real RVV hardware

Tested on an X60 RISC-V system:
widen_mul_test: 12/12 tests passed

Both signed and unsigned paths were checked in the generated assembly:
int16_t  -> int32_t:  vwmul.vv  + vcompress.vm
uint16_t -> uint32_t: vwmulu.vv + vcompress.vm

Multiple VLENs

TestAllReorderWidenMulAccumulate/RVV passes under QEMU with:
VLEN=128    PASS
VLEN=256    PASS
VLEN=512    PASS
VLEN=1024   PASS

A separate BF16 failure in TestAllRearrangeToOddPlusEven/RVV at VLEN=1024 is also reproducible on unmodified upstream/master and is unrelated to this change.

Performance

Microbenchmarks were run on the X60 with the benchmark pinned to a single CPU. The operands were changed on every iteration to prevent loop-invariant hoisting.

Average over 10 runs:

          Type                |   upstream   | this change
uint16_t -> uint32_t | 9.413 ns/iter | 9.413 ns/iter
int16_t -> int32_t     | 9.415 ns/iter | 9.416 ns/iter


jan-wassenberg (approved):
Makes sense, unfortunately IsEven does not work as-is for virtual.
Thanks for improving this :)

PR #3387, opened 2026-09-15
Add missing HWY_TEST_MAIN for HWY_TEST_STANDALONE

scuzqy:
Provide standalone entry points for 4 contrib tests.

Fixed MSVC builds.


jan-wassenberg (approved):
Thanks!

PR #3389, opened 2026-09-15
contrib/algo: fix Count/CountIf overflow on narrow tags

Wint3rNight:
Count and CountIf only widen their accumulators when the tag has at least 4 i8 or 2 i16 lanes. Narrower tags, and every 8/16-bit call on HWY_SCALAR, fall into the branch meant for 32/64-bit lanes and accumulate matches in their own lane type, so the count wraps. For example an all-matching u8 run of 2400 on a 2-lane tag returns 96.

This limits that branch to lanes of 32 bits or more, so narrow 8/16-bit tags use the existing CountTrue loop. The Count tests only ran on ForGE128Vectors so they now use ForPartialVectors, plus a new test with all-matching runs on narrow tags.


RaviTriv:
Thanks for fixing!

jan-wassenberg (approved):
Thanks for fixing and expanding the test!

PR #3354, opened 2026-09-03
faster Unique and AllUnique. Add algo_bench.cc

a-eee:
Previously, Unique used CompressBlendedStore, with the rationale that CompressStore couldn't be used because of in-place. But this is wrong, because always stores <= loads, so the next load can't be affected.  Also I added 4x unroll, which hides latency.

For AllUnique I also added 4x unroll, so now AllFalse is performed once every 4N elements.

On apple m5:
<details>
<summary>All measurements</summary>
Before:

```
=== BenchAll for NEON_BF16:
NEON_BF16:    unique:  random-runs             0%:  u8 1000000    84324.8 ns,  11.858910 GB/s
NEON_BF16:    unique:  random-runs            10%:  u8 1000000   154445.9 ns,   6.474757 GB/s
NEON_BF16:    unique:  random-runs            20%:  u8 1000000   178771.8 ns,   5.593724 GB/s
NEON_BF16:    unique:  random-runs            50%:  u8 1000000   180329.1 ns,   5.545416 GB/s
NEON_BF16:    unique:  random-runs            80%:  u8 1000000   179799.1 ns,   5.561764 GB/s
NEON_BF16:    unique:  random-runs            90%:  u8 1000000   179533.7 ns,   5.569985 GB/s
NEON_BF16:    unique:  random-runs           100%:  u8 1000000   145673.8 ns,   6.864651 GB/s
NEON_BF16:    unique:  random-runs             0%: i16 1000000    90008.3 ns,  22.220168 GB/s
NEON_BF16:    unique:  random-runs            10%: i16 1000000   155386.9 ns,  12.871094 GB/s
NEON_BF16:    unique:  random-runs            20%: i16 1000000   264025.0 ns,   7.575040 GB/s
NEON_BF16:    unique:  random-runs            50%: i16 1000000   355787.8 ns,   5.621328 GB/s
NEON_BF16:    unique:  random-runs            80%: i16 1000000   356649.2 ns,   5.607751 GB/s
NEON_BF16:    unique:  random-runs            90%: i16 1000000   356161.7 ns,   5.615426 GB/s
NEON_BF16:    unique:  random-runs           100%: i16 1000000   254565.9 ns,   7.856511 GB/s
NEON_BF16:    unique:  random-runs             0%: i32 1000000   166195.0 ns,  24.068116 GB/s
NEON_BF16:    unique:  random-runs            10%: i32 1000000   184724.1 ns,  21.653919 GB/s
[...]

PR #3391, opened 2026-09-16
add safe CopyIf and Unique tail processing, and speed up CopyIf

a-eee:
Before CompressBlendedStore was used in the Unique tail, with HWY_MEM_OPS_MIGHT_FAULT this is unsafe. Fixed

In CopyIf was the same problem, but not only the tail, every write might be unsafe. I found two ways to fix this:

use StoreN everywhere
store to the buffer and copy from the buffer to `to`

The second way turned out to be much faster(measurements in pr #3354), so Ive implemented it. The only thing left: a version with a guarantee of the `to` size.


hwy/contrib/algo/copy-inl.h
```diff
@@ -97,33 +104,53 @@ T* CopyIf(D d, const T* HWY_RESTRICT from, size_t count, T* HWY_RESTRICT to,
...
+      const Vec<D> v = LoadU(d, from + idx + inner);
+      const Mask<D> m = func(d, v);
+      cnt += CompressStore(v, m, d, buf + cnt);
     }
+    Copy(d, buf, cnt, to);
+    to += cnt;
+    idx += inner;
+  }
+#else
+  for (; idx + N <= count; idx += N) {
```
jan-wassenberg, line 152: I'd expect the unrolling to also help here, right?

hwy/contrib/algo/copy-inl.h
```diff
@@ -97,33 +104,53 @@ T* CopyIf(D d, const T* HWY_RESTRICT from, size_t count, T* HWY_RESTRICT to,
...
-    const T* HWY_RESTRICT from_idx =
-        reinterpret_cast<const T * HWY_RESTRICT>(addr + (idx * sizeof(T)));
-    const V1 v = LoadU(d1, from_idx);
-    // Avoid storing to `to` unless we know it should be kept - otherwise, we
-    // might overrun the end if it was allocated for the exact count.
-    if (CountTrue(d1, func(d1, v)) == 0) continue;
-    StoreU(v, d1, to);
-    to += 1;
-  }
+  const size_t remain = count - idx;
```
jan-wassenberg, line 162: Glad to see the one-by-one code gone, thanks for updating!

jan-wassenberg (dismissed):
Nice :) We can land as-is. Would you like to try the unrolling for the !unsafe case?

PR #3392, opened 2026-09-16
Add ShuffleSpan to hwy/contrib/algo

Wint3rNight:
Adds ShuffleSpan as discussed in #3374. Swap positions come from a bijection of the seed and the position, reduced with Lemire's multiply-shift a vector at a time, and the swaps then run in a plain loop. The permutation depends only on the seed, so it is the same on every target. A second overload takes a generator such as RngStream or std::mt19937, like std::shuffle, and draws in the same order as a sequential Fisher-Yates.

The hash is a copy of Triple32 from hash-inl.h, since including that header would make contrib/algo depend on VQSort through random-inl.h. The test checks that the copy matches Triple32. Positions past 2^32 use a scalar 64-bit path.

The tests compare against a sequential reference on every target and check uniformity for 4 and 64 elements. For 1M u32 on Zen 3 (AVX2), it takes 0.96 ns per element, against 6.55 for std::shuffle with mt19937.


jan-wassenberg:
Good to see v1 :) I'd like to avoid the duplication of Triple32, and we should replace SplitMix64 with Moremur or perhaps WeakXMX. So it would be necessary to refactor hash-inl.h to not depend on random-inl, which also shouldn't publicly depend on vqsort.h. hash-inl.h hasn't been released yet so we can still change it. I can take care of that, also requires some internal changes.

Also, on further thought, the seed codepath is too limiting - not enough entropy for longer sequences. Let's remove it and keep only the URBG, which we can use directly or use it to seed a vectorized generator.

Also, we should probably have two codepaths: one for L2-sized arrays (say 1 MB) does something similar to now, but beyond that we should probably instead use VQSort K32V32 or K64V64 to sort with random key, because that's a lot better than individual cache misses.

RngStream is not quite fast enough for that. We could add an overload that returns 128 random bits at a time rather than just 64. It could also be interesting to compare perf with CachedXoshiro. What do you think?

jan-wassenberg:
Update: we might also consider a short-array codepath that generates a random permutation and just does Lookup8..64 depending on the data type and CanLookup64 etc. The permutation can be obtained via `MaskedWeakTwoMul<5>` called with 0..31.

And for the sort codepath: for this to be viable, the caller would have to pass in temp memory with twice as many elements, preferably also padded to Lanes, because allocation overhead would be costly.

Wint3rNight:
Agreed on dropping the seed overload. Only that path uses Triple32 and SplitMix64, so both go with it. I'll wait for the hash-inl refactor before seeding a vectorized generator from the URBG.

Measured on Zen3, AVX2. Drawing one number at a time costs 3 to 4x more per element than the vectorized hash, so the vectorized generator is worth it. Returning both AES halves helps a little. CachedXoshiro was not faster than RngStream.

The random key sort didn't beat Fisher-Yates up to 1e8, though AVX-512 VQSort may differ. K32V32 also ties about a million pairs at that size, and ties keep their input order, so it would need K64V64, which was 2x slower again. Rao-Sandelius was faster: scatter into random L2-sized buckets, then Fisher-Yates per bucket. 9.5 ns per element against 15 to 22, still uniform, and the temp is n elements plus a bucket id. I'd use that for the large path.

MaskedWeakTwoMul<5> is ((x ^ key) * 11) mod 32 for every key, so it gives only 32 permutations. For the short path I'd build the index permutation with Fisher-Yates from one or two 128-bit draws, then Lookup.

:vqsort depends on //:algo, so a sort path can't live in contrib/algo. Move the file to contrib/random?

jan-wassenberg:
The hash-inl refactor has landed :)
Good data, thanks for sharing. AVX-512 vqsort is a decent speedup vs AVX2, but I'd believe that in-cache scatter plus merge is faster - the sorting is conceptually simpler but I'm OK to skip it.
If we aren't using the sort, then no more issues with the dependencies, right?

Wint3rNight:
Thanks for the refactor. And yes, no dependency issues now: hash-inl.h only pulls in hwy.

Though we may not need it. Triple32 and SplitMix64 are only used by the seed path. I'd assumed we could key the hash off the URBG instead, but that caps entropy at 64 bits, the same limit you flagged, so it goes with the seed overload.

That leaves drawing from the URBG directly, or seeding a VectorXoshiro from it, which is faster but makes the permutation depend on lane count. Do we still want consistent results across targets?

About the large path, should I open a follow-up PR for it or would you want it as a separate commit in this PR?

PR #3413, opened 2026-09-22
Refactor random generators and add MT19937 with NumPy-compatible jump

KimYannn:
Random generation currently combines RNG backends, buffering, and distributions in `random-inl.h`. This separates the reusable `hwy::Generator` and uniform distributions from the Xoshiro and AES-CTR backends, so additional engines can use the same sampling and fill interfaces.

- Keep standalone composition in `hwy/generator.h` and provide `hwy/generator-inl.h` for native SIMD returns with per-target attributes. Move distributions to `hwy/contrib/distributions` and engines/buffering to dedicated random headers.
- Remove the legacy `random-inl.h` wrapper API and Bazel `:random` target; migrate existing callers and update CMake, Bazel, and Meson registration. This is an include/API breaking change.
- Add scalar MT19937 with `std::mt19937` single-integer seeding, shared u32/u64 consumption, and `Jump`/`Jumped` matching native NumPy 2.5.3 state and cursor behavior for the same starting state. Keep state refill out of the hot draw path to allow inlining into bulk generation.

Validation:

- 123 related CTests passed; MT19937 ASan/UBSan and Rosetta x86_64 tests each passed 13/13. Distribution/target-Generator tests passed 24/24 natively, with ASan/UBSan, and under Rosetta x86_64.
- The complete distribution test compiles for SVE, SVE2, SVE_256, and SVE2_128 with Clang 23 at both `-O0` and `-O2`. The previous version reproduces the CI's 16 vector-return target errors under the same options. SVE runtime testing is left to CI.
- Direct comparison with native NumPy `.jumped()`: 889 scenarios and 38,542,776 matching outputs, covering cursor boundaries, repeated jumps, mixed u32/u64 consumption, and independent copies. Initial states were aligned using `RandomState(seed)`; this does not claim NumPy `SeedSequence` initialization compatibility.
- C++11 consumers compiled and ran using installed headers without linking Highway. MT19937 generation assembly was verified against the optimized pre-jump implementation.
- Build and installation were validated with CMake; Bazel and Meson registration was checked statically.


hwy/contrib/distributions/uniform.h
```diff
@@ -0,0 +1,59 @@
...
+  void Fill(BitGenerator& bits, double* out, size_t count) const {
+    for (size_t i = 0; i < count; ++i) out[i] = (*this)(bits);
+  }
+};
+
+// Uniform float samples on a 23-bit grid in [-1, 1). Consumes one full-range
+// u64 per result, using its low 23 bits as the mantissa.
+struct NormalizedUniform {
+  static float FromBits(uint64_t bits) noexcept {
+    const uint32_t exponent = hwy::BitCastScalar<uint32_t>(1.0f);
```
jan-wassenberg, line 39: If we use 2.0f, then the range is 2-4 and we can subtract 3 to get -1..1, in 2 fewer instructions.
Also, this adapter is scalar-only - seems vectorizing would make sense.

hwy/contrib/distributions/uniform-inl.h
```diff
@@ -0,0 +1,94 @@
...
+}  // namespace distribution_detail
+
+// The same distribution works with scalar u64 and native SIMD backends.
+// GenerateBlocks is an optional optimization; scalar backends need only ().
+struct Uniform {
+  static double FromBits(uint64_t bits) noexcept {
+    return hwy::Uniform::FromBits(bits);
+  }
+
+  static HWY_INLINE Vec<ScalableTag<double>> FromBits(
```
jan-wassenberg, line 59: Seems undesirable to hardcode the type. We can support any vector type, and use DFromV<V> d; then RebindToUnsigned<decltype(d)> to get the tag to pass to FromBits.

hwy/contrib/random/xoshiro-inl.h
```diff
@@ -0,0 +1,255 @@
...
+      if (valid_lanes == Lanes(tag)) {
+        StoreU(bits, tag, out + offset);
+      } else {
+        StoreN(bits, tag, out + offset, valid_lanes);
+      }
+    }
+  };
+
+  StateType state_;
+  const std::uint64_t streams_;
```
jan-wassenberg, line 235: Makes the class noncopyable, not sure this const is worthwhile.

hwy/contrib/random/cached.h
```diff
@@ -0,0 +1,65 @@
...
+  result_type operator()() {
+    if (HWY_UNLIKELY(index_ == kSize)) {
+      bit_generator_.FillBits(cache_.data(), kSize);
+      index_ = 0;
+    }
+    return cache_[index_++];
+  }
+
+  void FillBits(uint64_t* out, size_t count) {
+    for (size_t i = 0; i < count; ++i) out[i] = (*this)();
```
jan-wassenberg, line 51: Could use CopyBytes instead of a loop.

hwy/contrib/random/cached.h
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+ private:
+  static_assert(kSize != 0 && (kSize & (kSize - 1)) == 0,
```
jan-wassenberg, line 55: Seems unnecessary, we're not using wraparound, right?

jan-wassenberg (changes requested):
Thanks, I agree it makes sense to split AES and Xorshiro.
This is a large change, better to discuss the design of such things first.

As-is, this breaks a lot of (internal) dependencies, which is a non-starter. We could keep the existing header and build target, it can include the new headers, plus aliases.

It seems dubious to add Generator to core Highway - that should reside in contrib/, and I'm not seeing a lot of value added. How about we remove it?

Also, MT19937 is scalar and seems like a poor fit for Highway. It also has a large working set and is not a generator I'd encourage using. How about you keep the MT19937 (if you really can't/don't want to replace it with AES) in your project's source?

distributions/ currently seems unnecessary - or do we intend to extend it? If not, probably better to stay in the same directly.

BufferedBitGenerator: should we have lazy init in case it's unused?

